PR #7, opened 2019-12-07
Add support for pod annotations

avaidyanatha:
## Main Changes
- Adding support for applying custom pod annotations through the CRD

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -177,6 +177,10 @@ type VitessShardTabletPool struct {
 	// optimal scheduling of your pods if you choose to set this field.
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
+	// Annotations can optionally be used to attach custom annotations to Pods
+	// created for this component.
+	Annotations map[string]string `json:"annotations,omitempty"`
+
 	// ExtraEnv can optionally be used to override default environment variables
 	// set by the operator, or pass additional environment variables.
 	// These values are applied to both the vttablet and mysqld containers.
@@ -195,11 +199,6 @@ type VitessShardTabletPool struct {
 	// Typically, these are used to mount volumes defined through extraVolumes.
 	// These values are applied to both the vttablet and mysqld containers.
 	ExtraVolumeMounts []corev1.VolumeMount `json:"extraVolumeMounts,omitempty"`
-
-	// Annotations are key/value metadata pairs that can be attached to any Pod to select
-	// objects that satisfy certain conditions. This can be used for monitoring or access
-	// management tools that use tags for discovery.
-	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // VttabletSpec configures the vttablet server within a tablet.
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -195,6 +195,11 @@ type VitessShardTabletPool struct {
 	// Typically, these are used to mount volumes defined through extraVolumes.
 	// These values are applied to both the vttablet and mysqld containers.
 	ExtraVolumeMounts []corev1.VolumeMount `json:"extraVolumeMounts,omitempty"`
+
+	// Annotations are key/value metadata pairs that can be attached to any Pod to select
+	// objects that satisfy certain conditions. This can be used for monitoring or access
+	// management tools that use tags for discovery.
+	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // VttabletSpec configures the vttablet server within a tablet.
--- a/pkg/controller/vitesscell/reconcile_vtgate.go
+++ b/pkg/controller/vitesscell/reconcile_vtgate.go
@@ -32,6 +32,7 @@ import (
 	"planetscale.dev/vitess-operator/pkg/operator/reconciler"
 	"planetscale.dev/vitess-operator/pkg/operator/results"
 	"planetscale.dev/vitess-operator/pkg/operator/secrets"
+	"planetscale.dev/vitess-operator/pkg/operator/update"
 	"planetscale.dev/vitess-operator/pkg/operator/vtgate"
 )
 
@@ -113,12 +114,12 @@ func (r *ReconcileVitessCell) reconcileVtgate(ctx context.Context, vtc *planetsc
 	annotations := map[string]string{
 		"planetscale.com/secret-hash": secrets.ContentHash(gatewaySecrets...),
 	}
+	update.Annotations(&annotations, vtc.Spec.Gateway.Annotations)
 
 	// Reconcile vtgate Deployment.
 	spec := &vtgate.Spec{
 		Cell:              &vtc.Spec,
 		Labels:            labels,
-		PodAnnotations:    annotations,
 		Replicas:          *vtc.Spec.Gateway.Replicas,
 		Resources:         vtc.Spec.Gateway.Resources,
 		Authentication:    &vtc.Spec.Gateway.Authentication,
@@ -128,7 +129,7 @@ func (r *ReconcileVitessCell) reconcileVtgate(ctx context.Context, vtc *planetsc
 		ExtraEnv:          vtc.Spec.Gateway.ExtraEnv,
 		ExtraVolumes:      vtc.Spec.Gateway.ExtraVolumes,
 		ExtraVolumeMounts: vtc.Spec.Gateway.ExtraVolumeMounts,
-		Annotations:       vtc.Spec.Gateway.Annotations,
+		Annotations:       annotations,
 	}
 	key = client.ObjectKey{Namespace: vtc.Namespace, Name: vtgate.DeploymentName(clusterName, vtc.Spec.Name)}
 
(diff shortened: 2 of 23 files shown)


enisoc:
This isn't ready for review yet, right? It doesn't have the hashing we discussed offline. Can you convert it to a Draft PR if it isn't ready yet, so I can tell what needs attention?

avaidyanatha:
> This isn't ready for review yet, right? It doesn't have the hashing we discussed offline. Can you convert it to a Draft PR if it isn't ready yet, so I can tell what needs attention?

Yeah it's not ready yet, I apologize for not making it a draft PR - I don't think I can convert open PRs to draft.

enisoc:
Ah, yeah. Looks like there's no way to convert to draft after the fact. :( Thanks for adding the label, though!

pkg/controller/vitesscell/reconcile_vtgate.go
```diff
@@ -128,6 +128,7 @@ func (r *ReconcileVitessCell) reconcileVtgate(ctx context.Context, vtc *planetsc
 		ExtraEnv:          vtc.Spec.Gateway.ExtraEnv,
 		ExtraVolumes:      vtc.Spec.Gateway.ExtraVolumes,
 		ExtraVolumeMounts: vtc.Spec.Gateway.ExtraVolumeMounts,
+		Annotations:       vtc.Spec.Gateway.Annotations,
```
enisoc: We already have a `PodAnnotations` field. I think we should just combine them at this call site, since the `operator/vtgate` package shouldn't need to think of these as different use cases.

```go
annotations := ...
update.Annotations(&annotations, vtc.Spec.Gateway.Annotations)
```

pkg/operator/etcd/pod.go
```diff
@@ -315,3 +327,20 @@ func (spec *Spec) Args() []string {
 
 	return flags.FormatArgs()
 }
+
+func annotationHash(m map[string]string) string {
```
enisoc: This should go in a new package like `pkg/operator/stringmaps` since we need it in multiple places already. We can then call the function `HashKeys()` to make clear that we're only hashing the keys.

pkg/operator/etcd/pod.go
```diff
@@ -108,6 +112,14 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
 
+	// Update annotations and recreate pod if annotations are updated or removed.
+	annotations := spec.Annotations
+	annotationsHash := map[string]string{
+		"annotationsDeltaHash": annotationHash(annotations),
```
enisoc: Annotation keys that our operator sets should be prefixed with `planetscale.com/` to avoid overlap with annotations that other entities set. Also the convention is to use dashes to separate lowercase words, so we can call this `planetscale.com/annotation-keys-hash`.

pkg/operator/etcd/pod.go
```diff
@@ -108,6 +112,14 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
...
+	}
+	update.Annotations(&annotations, annotationsHash)
```
enisoc: This is going to end up mutating the input `spec.Annotations` since maps act like pointers. Let's do something like this:

```go
// Update desired annotations.
update.Annotations(&obj.Annotations, spec.Annotations)
// Record a hash of desired annotation keys to force the Pod
// to be recreated if a key disappears from the desired list.
update.Annotations(&obj.Annotations, map[string]string{
  "planetscale.com/annotation-keys-hash": stringmaps.HashKeys(spec.Annotations),
})
```

pkg/operator/vtctld/deployment.go
```diff
@@ -102,6 +103,9 @@ func UpdateDeployment(obj *appsv1.Deployment, spec *Spec) {
 	// Tell Deployment to set the same labels on the Pods it creates.
 	update.Labels(&obj.Spec.Template.Labels, spec.Labels)
 
+	// Tell Deployment to set annotations on Pods that it creates.
+	update.Annotations(&obj.Spec.Template.Annotations, spec.Annotations)
```
enisoc: We need to clobber since we aren't hashing here. Since it looks like we don't already set any annotations, this can just be:

```go
obj.Spec.Template.Annotations = spec.Annotations
```

pkg/operator/vtgate/deployment.go
```diff
@@ -122,8 +123,12 @@ func UpdateDeployment(obj *appsv1.Deployment, spec *Spec) {
 	update.Labels(&obj.Labels, spec.Labels)
 	// Tell Deployment to set the same labels on the Pods it creates.
 	update.Labels(&obj.Spec.Template.Labels, spec.Labels)
-
-	update.Annotations(&obj.Spec.Template.Annotations, spec.PodAnnotations)
+	// Clobber existing annotations so that we don't mutate the input map.
```
enisoc: The existing `spec.PodAnnotations` was already made for this purpose, so we can just reuse that. Let's rename it to `spec.Annotations` to be consistent with all our other new spec fields, though. Then this just needs to be one line like for vtctld:

```go
obj.Spec.Template.Annotations = spec.Annotations
```

pkg/operator/vttablet/pod.go
```diff
@@ -71,6 +74,14 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
 
+	// Update annotations and recreate pod if annotations are updated or removed.
```
enisoc: We were already setting annotations in `UpdatePodInPlace`. We need to stop doing it there and move it to here, since now we are relying on annotation changes to cause a rolling update of Pods.

Other than that, the comments from `etcd/pod.go` apply in this file too.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -195,6 +195,11 @@ type VitessShardTabletPool struct {
 	// Typically, these are used to mount volumes defined through extraVolumes.
 	// These values are applied to both the vttablet and mysqld containers.
 	ExtraVolumeMounts []corev1.VolumeMount `json:"extraVolumeMounts,omitempty"`
+
+	// Annotations are key/value metadata pairs that can be attached to any Pod to select
+	// objects that satisfy certain conditions. This can be used for monitoring or access
+	// management tools that use tags for discovery.
+	Annotations map[string]string `json:"annotations,omitempty"`
```
enisoc: This new field isn't actually being used anywhere yet. We need to update `reconcile_tablets.go` to combine the annotation it injects there (`drain.SupportedAnnotation`) with the user-supplied annotations from this new field before sending them to `vttablet.Spec`. Make sure not to mutate the input map while doing that.

pkg/operator/etcd/pod.go
avaidyanatha: Done.

pkg/operator/etcd/pod.go
avaidyanatha: Done.

pkg/operator/vtctld/deployment.go
avaidyanatha: Done.

pkg/operator/vtgate/deployment.go
avaidyanatha: Done.

pkg/operator/vttablet/pod.go
avaidyanatha: Makes sense. Done.

pkg/operator/stringmaps/stringmaps.go
```diff
@@ -0,0 +1,25 @@
+package stringmaps
```
enisoc: Add package doc comment:

```go
// Package stringmaps contains helpers for working with string maps.
```

pkg/operator/stringmaps/stringmaps.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"fmt"
+	"sort"
+)
+
+func HashKeys(m map[string]string) string {
```
enisoc: Add doc comments on all exported functions:

```go
// HashKeys returns a hex-encoded hash of all the keys in the map.
//
// This can be used to compare two unordered sets of keys for equality.
// The keys are assumed to not contain newlines.
// The values in the map are ignored.
```

pkg/operator/vtgate/deployment.go
```diff
@@ -95,6 +95,7 @@ type Spec struct {
 	ExtraEnv          []corev1.EnvVar
 	ExtraVolumes      []corev1.Volume
 	ExtraVolumeMounts []corev1.VolumeMount
+	Annotations       map[string]string
```
enisoc: Remove the old `PodAnnotations` field to make sure we aren't accidentally using it.

pkg/operator/vttablet/pod.go
```diff
@@ -314,3 +325,20 @@ func AliasFromPod(pod *corev1.Pod) topodatapb.TabletAlias {
 		Uid:  uint32(uid),
 	}
 }
+
+func annotationHash(m map[string]string) string {
```
enisoc: This isn't needed anymore.

pkg/operator/vtgate/deployment.go
avaidyanatha: Done.

pkg/operator/stringmaps/stringmaps.go
avaidyanatha: Done.

pkg/operator/stringmaps/stringmaps.go
avaidyanatha: Done.

pkg/operator/vttablet/pod.go
avaidyanatha: Removed!

Issue #1, opened 2019-12-03
Set up CI/CD


enisoc:
We need to choose and set up infra for running tests and pushing images.

enisoc:
Before we can set up image builds, we probably need to write a new Dockerfile that builds the binary inside the container, instead of the current Dockerfile generated by operator-sdk that expects the binary to be built outside.

enisoc:
Actually if we can run `docker` from GitHub Actions, we might be able to run our existing build/push process there.

enisoc:
I was hoping to use GitHub Packages, since they make it easy to push from GitHub Actions without worrying about credentials. However, it turns out that even public images in GitHub Packages cannot be pulled without authenticating.

Our options now are either:

1. Rewrite the Dockerfile so it's self-contained (builds the binary inside the container), and set up automated builds on Docker Hub.
2. Store Docker Hub credentials in a GitHub Secret, and use that to build and push from GitHub Actions.

enisoc:
I decided to try option 1. I set up an automated build in Docker Hub and triggered it. The build is still Pending.

enisoc:
The build succeeded, so I'll call this done.

PR #8, opened 2019-12-10
Add a Dockerfile for building release images.

enisoc:
We can't change the default Dockerfile because operator-sdk expects it
to behave a certain way. Instead, we'll build releases from this new
Dockerfile.release.

This should let us use Docker Hub for #1.

### DIFF
--- a/build/Dockerfile.release
+++ b/build/Dockerfile.release
@@ -1,0 +1,37 @@
+# This is a Dockerfile for building release images.
+#
+# The base Dockerfile is used by operator-sdk, so we can't use that name.
+# That Dockerfile merely copies a binary that has been built outside Docker,
+# which is useful for development work because it means the build shares the Go
+# module cache from your local machine.
+#
+# In contrast, this file builds the binary inside Docker so it's more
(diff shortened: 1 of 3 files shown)


build/Dockerfile.release
```diff
@@ -0,0 +1,37 @@
...
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
+
+ENV OPERATOR=/usr/local/bin/vitess-operator \
+    USER_UID=1001 \
+    USER_NAME=vitess-operator
+
+# install operator binary
+COPY --from=build /go/bin/manager ${OPERATOR}
+
+COPY build/bin /usr/local/bin
```
dctrwatson: I assume the files in this directory aren't going to change often, so this should go above the COPY of operator

enisoc: This whole section is copied from the Dockerfile that operator-sdk generates. I'd rather keep it the same so it's easy to synchronize. This build is already going to be super slow because of the time to download 4G of Go modules.

PR #13, opened 2019-12-17
Make imagePullPolicy configurable.

enisoc:


deploy/crds/planetscale_v2_vitesscluster_crd.yaml
```diff
@@ -486,6 +494,35 @@ spec:
                   - rootPath
                   type: object
               type: object
+            imagePullPolicies:
+              description: ImagePullPolicies specifies the container image pull policies
+                to use for images defined in the Images field.
```
PrismaPhonic: Should images be capitalized here?  The way this gets generated it's misleading because the key won't be PascalCase in the user supplied config.

enisoc: Changed.

PR #14, opened 2019-12-17
ExtraVitessFlags Feature in VitessClusterSpec

PrismaPhonic:
This PR adds `ExtraVitessFlags` field to `VitessClusterSpec` and appropriate duplication and plumbing where necessary to ensure that flags provided in `VitessClusterSpec` get plumbed through to `vtgate`, `vtctld` and `vttablet`.


deploy/crds/planetscale_v2_vitesscell_crd.yaml
```diff
@@ -36,6 +36,16 @@ spec:
               items:
                 type: string
               type: array
+            extraVitessFlags:
+              additionalProperties:
+                type: string
+              description: 'ExtraVitessFlags can optionally be used to override default
+                flags set by the operator, or pass additional flags to vttablet from
```
deepthi: only vttablet?

deploy/crds/planetscale_v2_vitesscluster_crd.yaml
```diff
@@ -781,6 +791,19 @@ spec:
                                           items:
                                             type: object
                                           type: array
+                                        extraVitessFlags:
+                                          additionalProperties:
+                                            type: string
+                                          description: 'ExtraFlags can optionally
+                                            be used to override default flags set
+                                            by the operator, or pass additional flags
+                                            to child components. All entries must
```
deepthi: It will be good to make the documentation uniform across all CRDs. Here we have `child components`, elsewhere we have `all child components` or `Vitess components`.

pkg/controller/vitesscell/reconcile_vtgate.go
```diff
@@ -116,6 +116,12 @@ func (r *ReconcileVitessCell) reconcileVtgate(ctx context.Context, vtc *planetsc
 	}
 	update.Annotations(&annotations, vtc.Spec.Gateway.Annotations)
 
+	// Merge ExtraVitessFlags into ExtraFlags.
```
deepthi: It seems that the same set of `ExtraVitessFlags` are intended to be passed to vtctld, vtgate and vttablet. That won't work because each component accepts a distinct (though not disjoint) set of flags. Passing a flag that a component doesn't accept will result in a failure to execute the binary.

pkg/apis/planetscale/v2/vitesscell_types.go
```diff
@@ -67,6 +67,13 @@ type VitessCellSpec struct {
 	// in which case this field is automatically managed by the VitessCluster
 	// controller that owns this VitessCell.
 	Images VitessCellImages `json:"images,omitempty"`
+
+	// ExtraVitessFlags can optionally be used to override default flags set by the
```
enisoc: Since this is outside the `VitessCellTemplate`, users can't set it, which is good because this is meant for internal use. In those cases, I prefer to keep the doc comment aimed at an internal audience:

```go
// ExtraVitessFlags is inherited from the parent's VitessClusterSpec.
```

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -68,6 +68,13 @@ type VitessKeyspaceSpec struct {
 
 	// BackupEngine specifies the Vitess backup engine to use, either "builtin" or "xtrabackup".
 	BackupEngine VitessBackupEngine `json:"backupEngine,omitempty"`
+
+	// ExtraVitessFlags can optionally be used to override default flags set by the
```
enisoc: As in VitessCell, we can just say:

```go
// ExtraVitessFlags is inherited from the parent's VitessClusterSpec.
```

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -77,6 +77,13 @@ type VitessShardSpec struct {
 
 	// BackupEngine specifies the Vitess backup engine to use, either "builtin" or "xtrabackup".
 	BackupEngine VitessBackupEngine `json:"backupEngine,omitempty"`
+
+	// ExtraVitessFlags can optionally be used to override default flags set by the
```
enisoc:
```go
// ExtraVitessFlags is inherited from the parent's VitessClusterSpec.
```

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -181,6 +188,13 @@ type VitessShardTabletPool struct {
 	// created for this component.
 	Annotations map[string]string `json:"annotations,omitempty"`
 
+	// ExtraFlags can optionally be used to override default flags set by the
+	// operator, or pass additional flags to child components. All entries must be
+	// key-value string pairs of the form "flag": "value". The flag name should
+	// not have any prefix (just "flag", not "-flag"). To set a boolean flag,
+	// set the string value to either "true" or "false".
+	ExtraFlags map[string]string `json:"extraVitessFlags,omitempty"`
```
enisoc: I don't think we need to expose this. We can just invisibly merge the VitessShard-level ExtraVitessFlags into VttabletSpec.ExtraFlags.

deploy/crds/planetscale_v2_vitesscluster_crd.yaml
PrismaPhonic: I can try to make it more uniform, but the field is somewhat different depending on where it is.  The idea is that there's a single parent `ExtraVitessFlags` in `VitessClusterSpec` which gets copied to where those flags need to be available.  Currently they need to be made available to `vtgate`, `vtctld` and `vttablet`. When it exists in a parent accessible by one of those child components, it has a more specific use case.  I suppose we could generally say that regardless the flags are accessible to `all child components` and have a uniform comment.

pkg/operator/vttablet/spec.go
```diff
@@ -49,6 +49,7 @@ type Spec struct {
 	BackupLocation           *planetscalev2.VitessBackupLocation
 	BackupEngine             planetscalev2.VitessBackupEngine
 	Affinity                 *corev1.Affinity
+	ExtraFlags               map[string]string
```
enisoc: If we merge VitessShard.VitessExtraFlags into VttabletSpec.ExtraFlags, we don't need this new field.

pkg/controller/vitesscell/reconcile_vtgate.go
enisoc: I think we should merge in the opposite order, so the more-specific flags (for vtgate) take precedence over less-specific (vitess-wide) flags.

Also, we should create a new, empty map to avoid mutating our inputs.

PrismaPhonic: This feature is intended for passing flags that are shared by all components - such as topo flags (that's the specific use case we are trying to support in the operator with this feature).  I'll improve the comments to be more explicit about how `ExtraVitessFlags` is intended only for flags that can be passed to `vtctld`, `vtgate`, and `vttablet`.

deploy/crds/planetscale_v2_vitesscluster_crd.yaml
enisoc: I suggest we only document the field in VitessCluster. The other fields are all meant to be for internal use only; users can't even set them because they aren't present in the `*Template`. For these kinds of internal fields, I generally write the doc comment assuming the audience is other vitess-operator contributors. In this case, all they need to know is, "this field is just a copy of the one from VitessCluster; go look there to see what this does."

pkg/apis/planetscale/v2/vitesscell_types.go
PrismaPhonic: Noted - thanks for the tip.  Will adjust now.

pkg/controller/vitesscell/reconcile_vtgate.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/14/commits/1a12cb834722dc60a7ce9fbe93dbfa70ba5cc92a.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/14/commits/226927d4cabfc3a79db1fcea1f2079d61b4ec302.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/14/commits/226927d4cabfc3a79db1fcea1f2079d61b4ec302.

pkg/apis/planetscale/v2/vitesscell_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/14/commits/226927d4cabfc3a79db1fcea1f2079d61b4ec302.

deploy/crds/planetscale_v2_vitesscluster_crd.yaml
PrismaPhonic: Adjusted per @enisoc's suggestion https://github.com/planetscale/vitess-operator/pull/14/commits/226927d4cabfc3a79db1fcea1f2079d61b4ec302.

deploy/crds/planetscale_v2_vitesscell_crd.yaml
PrismaPhonic: This has now been adjusted per @enisoc's suggestion below https://github.com/planetscale/vitess-operator/pull/14/commits/226927d4cabfc3a79db1fcea1f2079d61b4ec302.

pkg/controller/vitesscell/reconcile_vtgate.go
PrismaPhonic: Added more clarification in the doc comment on the root `ExtraVitessFlags` https://github.com/planetscale/vitess-operator/pull/14/commits/f966fb07056e669256ffad20ec674fb2d83e881f.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Adjusted in https://github.com/planetscale/vitess-operator/pull/14/commits/6761ea423399f13542e587676abf9bf4ee58db196.

pkg/operator/vttablet/spec.go
PrismaPhonic: Adjusted in https://github.com/planetscale/vitess-operator/pull/14/commits/6761ea423399f13542e587676abf9bf4ee58db19.

PrismaPhonic:
All review comments and suggestions have been addressed. PTAL.

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -261,6 +261,9 @@ func vttabletSpecs(vts *planetscalev2.VitessShard, parentLabels map[string]strin
 			labels[planetscalev2.TabletTypeLabel] = string(pool.Type)
 			labels[planetscalev2.TabletIndexLabel] = strconv.FormatUint(uint64(tabletIndex), 10)
 
+			// Merge ExtraVitessFlags into the tablet spec ExtraFlags field.
+			update.StringMap(&pool.Vttablet.ExtraFlags, vts.Spec.ExtraVitessFlags)
```
enisoc: We should make a new map like the others, to avoid mutating inputs.

pkg/operator/vttablet/pod.go
```diff
@@ -92,6 +92,14 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 		vttabletAllFlags[key] = value
 	}
 
+	// Apply user-supplied extra flags last so they take precedence.
```
enisoc: This is already done right above.

pkg/controller/vitessshard/reconcile_tablets.go
PrismaPhonic: Isn't the goal in this specific case to mutate ExtraFlags on VttabletSpec?

pkg/operator/vttablet/pod.go
PrismaPhonic: That was very silly of me - you're right.  Fixed https://github.com/planetscale/vitess-operator/pull/14/commits/676c06bac87aa48d59672de11d833a2258a7b3a3.

pkg/controller/vitessshard/reconcile_tablets.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/14/commits/74fb0d0aa4f38f826d38612baf778c94c0ed6095.

PR #19, opened 2019-12-20
Feature: Init Container Injection

PrismaPhonic:
This adds the feature of init container injection anywhere we have `ExtraVolumes` and updates InitContainers in relevant pod specs.

### DIFF
--- a/pkg/apis/planetscale/v2/etcdlockserver_types.go
+++ b/pkg/apis/planetscale/v2/etcdlockserver_types.go
@@ -104,6 +104,10 @@ type EtcdLockserverTemplate struct {
 	// should be mounted.
 	ExtraVolumes []corev1.Volume `json:"extraVolumes,omitempty"`
 
+	// InitContainers can optionally be used to supply extra init containers
+	// that will be run to completion one after another before any app containers are started.
+	InitContainers []corev1.Container `json:"initContainers,omitempty"`
+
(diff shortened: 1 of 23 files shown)


pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -104,6 +104,10 @@ type EtcdLockserverTemplate struct {
 	// should be mounted.
 	ExtraVolumes []corev1.Volume `json:"extraVolumes,omitempty"`
 
+	// InitContainers can optionally be used to supply extra init containers
```
enisoc: Nit: Can you order these new fields after all the `Extra*` fields, so those similarly-named ones remain grouped?

PrismaPhonic: Sure, will do.

PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/19/commits/6f12c1b846c16e4b904f8d7d5f05fd5257bd5892.

PrismaPhonic:
@enisoc I tried to execute on what you were asking - but not sure if my approach of pulling `InitContainers` from the tablet spec was the correct approach for `vtbackup_pod.go`.  Would you please take a look?

PR #23, opened 2020-01-10
Tolerate extra VolumeMounts in Pod Containers.

enisoc:
This fixes a bug introduced in #20 that causes vttablet/etcd Pods to always show scheduled changes. It turns out item 2 there was an incorrect assumption in some cases, so we need a more nuanced solution.

When managing Pods directly, as opposed to filling in a Pod Template in something like a Deployment, we need to tolerate extra VolumeMounts being added to our Containers. The API server adds a mount to every Container for the ServiceAccount token, for example.

At the same time, we need to notice when one of the "extra" VolumeMounts is actually one that we put there ourselves previously, since we need to remove it instead of leaving it untouched.

This change extends a pattern that we borrowed from Deployment itself to handle our Annotations to also apply to VolumeMounts in Containers.

The trick is to set an Annotation whose value depends on the unordered set of Containers/VolumeMounts we ourselves care about. If that set ever changes, the Annotation changes, and our Pod will have to be deleted and recreated, effectively removing anything we no longer want.

Instead of storing the full list of Containers and VolumeMounts that we previously wanted (to distinguish them from ones that someone else added), we simply store a hash of these lists, since all we care about is comparing for equality.


pkg/operator/contenthash/slices.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// StringList returns a hex-encoded hash of a list of strings.
+//
+// This can be used to compare two ordered lists of strings for equality.
+// The keys are assumed to not contain newlines.
+func StringList(in []string) string {
+	h := md5.New()
+	for _, k := range in {
+		fmt.Fprintln(h, k)
```
PrismaPhonic: This seems inconsistent with other functions that produce hex-encoded hashes of things. Should we keep it consistent?
```suggestion
		sHash := md5.Sum([]byte(k))
		h.Write(sHash[:])
```

enisoc: Good point. Updated.

PR #24, opened 2020-01-13
VitessShard Reconcile: Filter Cells by Responsibility.

PrismaPhonic:
This PR changes how we get a tablet map for a shard, by filtering the request by a list of cells. That list of cells is retrieved directly from the spec. Changing the way we pull for this info is early work in preparation for cross-region support, and ensures that shards only act on tablets they are directly responsible for.

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_methods.go
+++ b/pkg/apis/planetscale/v2/vitessshard_methods.go
@@ -128,7 +128,7 @@ func (s *VitessShardSpec) BackupsEnabled() bool {
 	return false
 }
 
-// GetCells returns a filtered and de-duplicated list of all cells the current shard is responsible for.
+// GetCells returns a set of all cells the current shard is responsible for.
 func (s *VitessShardSpec) GetCells() sets.String {
 	cells := sets.String{}
 
--- a/pkg/apis/planetscale/v2/vitessshard_methods.go
+++ b/pkg/apis/planetscale/v2/vitessshard_methods.go
@@ -18,6 +18,7 @@ package v2
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // IsExternalMaster indicates whether the tablet is in a pool of type "externalmaster".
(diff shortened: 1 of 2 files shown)


pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -126,3 +127,14 @@ func (s *VitessShardSpec) BackupsEnabled() bool {
 	}
 	return false
 }
+
+// GetCells returns a set of all cells the current shard is responsible for.
```
enisoc: Let's be more explicit, since it's ambiguous whether "shard" means "this VitessShard object" or "the Vitess shard itself". The tablets in other regions are still part of the same Vitess shard, even if they aren't included in this VitessShard object.

```suggestion
// GetCells returns the set of all cells used by any tablet pools
// defined in this VitessShardSpec.
```

pkg/controller/vitessshardreplication/reconcile_drain.go
```diff
@@ -118,7 +118,7 @@ func (r *ReconcileVitessShard) reconcileDrain(ctx context.Context, vts *planetsc
 	}
 
 	// Get all the tablet records for the shard.
-	tablets, err := wr.TopoServer().GetTabletMapForShard(ctx, keyspaceName, vts.Spec.Name)
+	tablets, err := wr.TopoServer().GetTabletMapForShardByCell(ctx, keyspaceName, vts.Spec.Name, vts.Spec.GetCells().UnsortedList())
```
enisoc: Edit the comment above this line:

```go
// Get all the tablet records for the shard, in cells to which we deploy.
// We ignore tablets in cells we don't deploy, since we assume there's
// a separate operator instance handling drains on those tablets.
```

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Done.

pkg/controller/vitessshardreplication/reconcile_drain.go
PrismaPhonic: Done.

PR #26, opened 2020-01-14
Feature: Make Topo Reconciliation Configurable

PrismaPhonic:
This PR makes topo reconciliation configurable. It allows users to either enable or disable registration or pruning of various vitess components from topo records.


pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -107,6 +107,37 @@ type VitessClusterSpec struct {
...
 	ExtraVitessFlags map[string]string `json:"extraVitessFlags,omitempty"`
+
+	// TopoReconciliation can be used to enable or disable registration or pruning of various vitess components to and from topo records.
+	TopoReconciliation TopoReconcileConfig `json:"topoReconciliation,omitempty"`
+}
+
+// TopoReconcileConfig can be used to turn on or off registration or pruning of specific vitess components from topo records.
+// This should only be necessary if you need to override defaults, and shouldn't be required for the vast majority of use cases.
+type TopoReconcileConfig struct {
+	// RegisterCellsAliases can be used to enable or disable registering cells aliases into topo records.
```
enisoc: Add an extra line to each of these doc comments to give the default value in the format we use throughout the rest of the API:

```go
// Default: true
```

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +109,41 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
 		backup.Engine = defaultBackupEngine
 	}
 }
+
+func defaultTopoReconcileConfig(vt *VitessCluster) {
+	// Defaulting registration code.
+	if vt.Spec.TopoReconciliation.RegisterCells == nil {
+		registerCells := true
+		vt.Spec.TopoReconciliation.RegisterCells = &registerCells
```
enisoc: Use the utility function `pointer.BoolPtr(true)`.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +109,41 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
 		backup.Engine = defaultBackupEngine
 	}
 }
+
+func defaultTopoReconcileConfig(vt *VitessCluster) {
```
enisoc: Pass in `*TopoReconcileConfig` since that's the smallest scope at which the function can do its job.

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -107,6 +107,37 @@ type VitessClusterSpec struct {
...
+	RegisterCellsAliases *bool `json:"registerCellsAliases,omitempty"`
```
teejae:
```suggestion
	RegisterCellAliases *bool `json:"registerCellsAliases,omitempty"`
```

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -107,6 +107,37 @@ type VitessClusterSpec struct {
 	// not have any prefix (just "flag", not "-flag"). To set a boolean flag,
 	// set the string value to either "true" or "false".
 	ExtraVitessFlags map[string]string `json:"extraVitessFlags,omitempty"`
+
+	// TopoReconciliation can be used to enable or disable registration or pruning of various vitess components to and from topo records.
+	TopoReconciliation TopoReconcileConfig `json:"topoReconciliation,omitempty"`
```
enisoc: I think we generally should avoid abbreviations in the public API unless they're common even outside the context of Vitess (e.g. `min/max`). Since `topo` is an abbreviation new Vitess users wouldn't already know, let's spell it out:

```suggestion
	TopologyReconciliation TopoReconcileConfig `json:"topologyReconciliation,omitempty"`
```

pkg/apis/planetscale/v2/vitesscluster_types.go
teejae: i'd put this under RegisterCells, since Aliases are presumably less important but related to Cells.

pkg/apis/planetscale/v2/vitesscluster_types.go
teejae:
```suggestion
	TopoReconcile TopoReconcileConfig `json:"topoReconciliation,omitempty"`
```

PrismaPhonic: This goes in the opposite direction of @enisoc's suggested change. I'm going to go with his suggestion because the justification seems solid for expanding rather than contracting. Thanks for the feedback, will resolve this once I implement @enisoc's suggestion.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: The official feature in Vitess is called CellsAlias, and we are talking about the plural (more than one CellsAlias).

PrismaPhonic: Cells and CellsAlias are actually very different concepts. The name, unfortunately, makes it sound more similar than it is.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/a472f7bf82e16589110301ada180af75a5b98930.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/a472f7bf82e16589110301ada180af75a5b98930.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/b026aec4a4017c54b1651342e80dc811ff55c00c.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/b026aec4a4017c54b1651342e80dc811ff55c00c6.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +122,39 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
...
+func defaultTopoReconcileConfig(t *TopoReconcileConfig) {
+	// Fail fast with static defaults if user didn't supply any field.
+	if t == nil {
+		t = &defaultTopologyReconciliationConfig
+		return
+	}
+
+	// Defaulting registration code.
+	if t.RegisterCells == nil {
+		t.RegisterCells = pointer.BoolPtr(true)
```
enisoc: The default value for each field is now stored in two separate places. We should unify them so there's only one source of truth. Ideally, there would also only be one code path that ever sets the default for a given field.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -22,13 +22,27 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+var (
+	defaultTopologyReconciliationConfig = TopoReconcileConfig{
```
enisoc: So far, the pattern has been to keep the concrete default values in `defaults.go`, and the `*_defaults.go` files just have methods to fill in those concrete values. We might actually rethink the single `defaults.go` file just like we broke up `helpers.go` but we should keep things consistent until we decide to refactor.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +122,39 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
 		backup.Engine = defaultBackupEngine
 	}
 }
+
+func defaultTopoReconcileConfig(t *TopoReconcileConfig) {
+	// Fail fast with static defaults if user didn't supply any field.
+	if t == nil {
+		t = &defaultTopologyReconciliationConfig
```
enisoc: There are two problems here:

1. Assigning to `t` doesn't affect the caller's data structure since `t` is a local variable. We need to pass in `t **TopoReconcileConfig` and then check if `*t` is `nil`. We can assume `t` will never be `nil` in that case.
2. We need to make a deep copy of `defaultTopologyReconciliationConfig` since it contains pointers. Otherwise two different objects that both call this defaulter function will end up sharing memory unexpectedly. Helpfully, `make generate` has already created a `DeepCopy()` method on `TopoReconcileConfig` for you. Alternatively, it might be cleaner to just allocate an empty `&TopoReconcileConfig{}` here and then fall through to the individual checks below, since you're going to need those anyway. That will make it clearer that there's only one code path that could set a given field.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,44 +85,53 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
-	// We need to add an alias for all the cells in each region so that vtgate
-	// knows that it can route traffic between them.
-	// Currently, only one region is supported so we allow routing anywhere.
-	//
-	// We also need to create the aliases before we create the cells because we
-	// don't want any vtgates to start after the cells are created but before
-	// the alias exists.
-	desiredCellsAliases := buildCellsAliases(desiredCells)
-	currentCellsAliases, err := ts.GetCellsAliases(ctx, true)
+	if *vt.Spec.TopologyReconciliation.RegisterCellsAliases == true {
```
enisoc: The `== true` is already implicit in every conditional.

```suggestion
	if *vt.Spec.TopologyReconciliation.RegisterCellsAliases {
```

pkg/apis/planetscale/v2/vitesscluster_defaults.go
PrismaPhonic: Ok, I suppose in that case I would opt for clarity over efficiency since it sounds like the efficient path likely doesn't save us much worthwhile anyways. Thanks for the feedback, I'll refactor.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: Ahhh yeah, of course. This part of the code was a WIP - but thanks.  Tidied up now.

deepthi:
Basic question: what is the motivation for this feature? Is there an issue or requirement written up somewhere that I can read?

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -148,36 +160,54 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
 		if err != nil {
 			// Record the error and continue trying other cells.
 			r.recorder.Eventf(vt, corev1.EventTypeWarning, "TopoUpdateFailed", "failed to update lockserver address for cell %v", name)
-			resultBuilder.RequeueAfter(topoRequeueDelay)
+			return err
```
enisoc: This function (`registerCells`) should create its own result builder because it shouldn't always bail out when it encounters an error, as mentioned in the comment above: "Record the error and continue trying other cells". The result builder is a helper to implement this pattern of recording the error but continuing anyway.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,44 +85,56 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
+
+	// Clean up cells that exist but shouldn't.
+	for _, name := range cellNames {
+		if desiredCells[name] == nil && vt.Status.OrphanedCells[name] == nil {
+			// The cell exists in topo, but not in the VT spec.
+			// It's also not being kept around by a blocked turn-down.
+			// See if we can delete it. This will fail if the cell is not empty.
+			if err := ts.DeleteCellInfo(ctx, name); err != nil {
+				r.recorder.Eventf(vt, corev1.EventTypeWarning, "TopoCleanupFailed", "unable to remove cell %s from topology: %v", name, err)
+				return err
```
enisoc: This used to be a "schedule requeue but don't return" so we shouldn't convert it into a return. That means this function (`pruneCells`) also needs to use its own result builder.

PrismaPhonic: Ahh good point, ok.  Will adjust.

PrismaPhonic: Adjusted - PTAL.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: Adjusted - PTAL.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,43 +85,59 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
+		result, err := r.registerCells(ctx, vt, ts, globalTopoImpl, desiredCells)
+		resultBuilder.Merge(result, err)
+	}
+
+	if *vt.Spec.TopologyReconciliation.PruneCells {
+		result, err := r.pruneCells(ctx, vt, ts, desiredCells)
+		if err != nil {
+			return resultBuilder.RequeueAfter(topoRequeueDelay)
+		} else {
+			resultBuilder.Merge(result, err)
```
enisoc: This can be brought out of the `else` block to reduce indentation of the happy path, since the preceding `if` follows the conventional pattern of `if error, return`.

PrismaPhonic: Done.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -61,6 +61,14 @@ func (r *ReconcileVitessCell) reconcileTopology(ctx context.Context, vtc *planet
 	}
 	defer ts.Close()
 
+	if *vtc.Spec.TopologyReconciliation.PruneSrvKeyspaces {
+		return r.pruneSrvKeyspaces(ctx, vtc, keyspaces, ts)
+	}
+
+	return nil, nil
```
enisoc: Even if we don't prune, we still need to build and return the `topoEntries` list. Let's fetch `srvKeyspaceNames` here, compute `topoEntries` and `wanted`, and then pass `srvKeyspaceNames` and `wanted` into the prune subroutine.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -190,6 +223,17 @@ func (r *ReconcileVitessCluster) reconcileKeyspaceTopology(ctx context.Context,
 		desiredKeyspaces[keyspace.Name] = keyspace
 	}
 
+	if *vt.Spec.TopologyReconciliation.PruneKeyspaces == true {
+		result, err := r.pruneKeyspaces(ctx, vt, ts, desiredKeyspaces, resultBuilder)
+		if err != nil {
+			return result, err
```
enisoc: This should be a `resultBuilder.Merge()` rather than a return. The old code did not bail out if something went wrong with keyspace pruning. It doesn't technically matter right now because `reconcileKeyspaceTopology` is not doing anything else yet, but since it uses `resultBuilder`, we should follow the proper conventions so everything still works as more is added later.

pkg/controller/vitesskeyspace/reconcile_topo.go
```diff
@@ -55,6 +55,17 @@ func (r *ReconcileVitessKeyspace) reconcileTopology(ctx context.Context, vtk *pl
...
+		resultBuilder.Merge(result, err)
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessKeyspace) pruneShards(ctx context.Context, vtk *planetscalev2.VitessKeyspace, ts *toposerver.Conn) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
 	// Get list of keyspaces in topo.
```
enisoc: Not your fault, but can you fix this comment? It should say shards, not keyspaces.

pkg/controller/vitessshard/reconcile_topo.go
```diff
@@ -63,40 +63,22 @@ func (r *ReconcileVitessShard) reconcileTopology(ctx context.Context, vts *plane
...
-	if shard, err := ts.GetShard(ctx, keyspaceName, vts.Spec.Name); err == nil {
-		vts.Status.HasMaster = k8s.ConditionStatus(shard.HasMaster())
-		if shard.MasterAlias != nil {
-			vts.Status.MasterAlias = topoproto.TabletAliasString(shard.MasterAlias)
-		}
-
-		// Is the shard in the serving partition for any cell or tablet type?
-		if servingCells, err := ts.GetShardServingCells(ctx, shard); err == nil {
-			vts.Status.Idle = k8s.ConditionStatus(len(servingCells) == 0)
+	if *vts.Spec.TopologyReconciliation.PruneShardCells {
```
enisoc: We need to pull out the parts that read topo and update `vts.Status` and still do that. This flag should only make us skip the mutation of topo.

pkg/controller/vitessshard/reconcile_topo.go
```diff
@@ -63,40 +63,22 @@ func (r *ReconcileVitessShard) reconcileTopology(ctx context.Context, vts *plane
...
-				}
-			}
-		} else {
-			r.recorder.Eventf(vts, corev1.EventTypeWarning, "TopoGetFailed", "failed to get shard serving cells: %v", err)
-			resultBuilder.RequeueAfter(topoRequeueDelay)
-		}
-	} else {
-		r.recorder.Eventf(vts, corev1.EventTypeWarning, "TopoGetFailed", "failed to get shard info: %v", err)
-		resultBuilder.RequeueAfter(topoRequeueDelay)
+	if *vts.Spec.TopologyReconciliation.PruneTablets {
```
enisoc: Same here. We need to still update tablet status even if we aren't pruning.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +109,39 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
 		backup.Engine = defaultBackupEngine
 	}
 }
+
+func defaultTopoReconcileConfig(confPtr **TopoReconcileConfig) {
+	// Fail fast with static defaults if user didn't supply any field.
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/bd53c9410b840112368abb39f270307cc0b39209.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -108,3 +109,39 @@ func defaultClusterBackup(backup *ClusterBackupSpec) {
...
+
+	// Defaulting registration code.
+	if conf.RegisterCells == nil {
+		conf.RegisterCells = pointer.BoolPtr(true)
+	}
+	if conf.RegisterCellsAliases == nil {
+		conf.RegisterCellsAliases = pointer.BoolPtr(true)
+	}
+
+	// Defaulting pruning code.
```
avaidyanatha: Could provide a comment about how this prevents panics unless you feel like it's self-evident.

pkg/controller/vitesskeyspace/reconcile_topo.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/bd53c9410b840112368abb39f270307cc0b39209.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: Fixed https://github.com/planetscale/vitess-operator/pull/26/commits/03448bb30f936614a4b9582b59fd7bcf24286f8a.

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -107,6 +107,45 @@ type VitessClusterSpec struct {
...
 	// set the string value to either "true" or "false".
 	ExtraVitessFlags map[string]string `json:"extraVitessFlags,omitempty"`
+
+	// TopologyReconciliation can be used to enable or disable registration or pruning of various vitess components to and from topo records.
+	TopologyReconciliation *TopoReconcileConfig `json:"topologyReconciliation,omitempty"`
+}
+
+// TopoReconcileConfig can be used to turn on or off registration or pruning of specific vitess components from topo records.
+// This should only be necessary if you need to override defaults, and shouldn't be required for the vast majority of use cases.
+type TopoReconcileConfig struct {
```
avaidyanatha: I don't know if this is just me being nitpicky, but I think that there should be a simple option to just disable pruning for all Vitess resources instead of having to specify for each one.

pkg/controller/vitesscell/reconcile_topo.go
PrismaPhonic: Refactored per our offline discussion. When you have a moment PTAL.

pkg/controller/vitessshard/reconcile_topo.go
PrismaPhonic: Done. PTAL.

pkg/controller/vitessshard/reconcile_topo.go
PrismaPhonic: Done. PTAL.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
PrismaPhonic: This is generally the reason for all of our defaulting code in general. Do you think the comment is still necessary given that info?

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: @enisoc thoughts here? My thought is that adding an additional helper field adds an extra level of complexity to the system, although it does make it nice in a way. Thoughts?

enisoc: We don't yet know of a use case for disabling all pruning, so let's start with just addressing the use cases we do know about.

What we need right now is, "Don't try to prune or register things about which you might have incomplete information if you're one operator instance among a group that are collaborating to form a cross-k8s vitess cluster." However, we can't define the field that way at this level because one of the things these operator instances will lack complete information about is whether or not they have complete information.

The individual fields for disabling certain types of pruning and registration allow us to maintain the abstraction boundary that each operator instance only needs to take care of what's happening in its local k8s cluster.

pkg/controller/vitesscell/reconcile_keyspaces.go
```diff
@@ -97,17 +98,45 @@ func (r *ReconcileVitessCell) reconcileKeyspaces(ctx context.Context, vtc *plane
...
+	}
+	defer ts.Close()
+
 	// We have the list of keyspaces that should exist in this cell.
 	// See if we need to clean up topology.
-	topoEntries, err := r.reconcileTopology(ctx, vtc, keyspaces)
+	result, err := r.reconcileTopology(ctx, vtc, ts, keyspaces)
+	resultBuilder.Merge(result, err)
 	if err != nil {
 		return resultBuilder.RequeueAfter(topoRequeueDelay)
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/1e6f5795be775f021e8d22a7df124e8228c35d4a.

pkg/controller/vitesscell/reconcile_keyspaces.go
```diff
@@ -97,17 +98,45 @@ func (r *ReconcileVitessCell) reconcileKeyspaces(ctx context.Context, vtc *plane
...
 
-	if topoEntries != nil {
+	// Get the list of keyspaces deployed (served) in this cell.
+	srvKeyspaceNames, err := ts.GetSrvKeyspaceNames(ctx, vtc.Spec.Name)
+	if err != nil {
+		r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoListFailed", "failed to list keyspaces in cell-local lockserver: %v", err)
+		return resultBuilder.Error(err)
+	}
+
+	if srvKeyspaceNames != nil {
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/60e56d9e2f40ae57fe5d474320743263a55fba77.

pkg/controller/vitesscell/reconcile_keyspaces.go
```diff
@@ -97,17 +98,45 @@ func (r *ReconcileVitessCell) reconcileKeyspaces(ctx context.Context, vtc *plane
...
+		}
+	}
+
+	// We actually know the address of the local lockserver already,
+	// but for now we'll follow the same rule as all Vitess components,
+	// which is to use the global lockserver to find the local ones.
+	ts, err := toposerver.Open(ctx, vtc.Spec.GlobalLockserver)
+	if err != nil {
+		r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoConnectFailed", "failed to connect to global lockserver: %v", err)
+		return resultBuilder.Error(err)
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/00798c69e2692a228ccb806fc4d326d2849e4736.

pkg/controller/vitesscell/reconcile_keyspaces.go
```diff
@@ -97,17 +98,45 @@ func (r *ReconcileVitessCell) reconcileKeyspaces(ctx context.Context, vtc *plane
...
 	if err != nil {
 		return resultBuilder.RequeueAfter(topoRequeueDelay)
 	}
 
-	if topoEntries != nil {
+	// Get the list of keyspaces deployed (served) in this cell.
+	srvKeyspaceNames, err := ts.GetSrvKeyspaceNames(ctx, vtc.Spec.Name)
+	if err != nil {
+		r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoListFailed", "failed to list keyspaces in cell-local lockserver: %v", err)
+		return resultBuilder.Error(err)
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/00798c69e2692a228ccb806fc4d326d2849e4736.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -83,17 +75,15 @@ func (r *ReconcileVitessCell) reconcileTopology(ctx context.Context, vtc *planet
...
 			continue
 		}
+
 		// It's not wanted. Try to delete it.
 		if err := ts.DeleteSrvKeyspace(ctx, vtc.Spec.Name, srvKeyspaceName); err != nil {
-			// We failed to delete it, so assume the entry still exists.
-			topoEntries = append(topoEntries, srvKeyspaceName)
-
 			r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoCleanupBlocked", "unable to remove keyspace %s from cell-local topology: %v", srvKeyspaceName, err)
 			resultBuilder.Error(err)
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/c9529a32ab5e984429a9a1c696de92d87710d3ea.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -36,44 +36,36 @@ const (
...
+
+func (r *ReconcileVitessCell) pruneSrvKeyspaces(ctx context.Context, vtc *planetscalev2.VitessCell, keyspaces []*planetscalev2.VitessKeyspace, ts *toposerver.Conn) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
 
 	// Get the list of keyspaces deployed (served) in this cell.
 	srvKeyspaceNames, err := ts.GetSrvKeyspaceNames(ctx, vtc.Spec.Name)
 	if err != nil {
 		r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoListFailed", "failed to list keyspaces in cell-local lockserver: %v", err)
-		return nil, err
+		return resultBuilder.Error(err)
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/c9529a32ab5e984429a9a1c696de92d87710d3ea.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -36,44 +36,36 @@ const (
...
 	}
 
 	// We don't need to create Keyspace entries in topo for keyspaces that should exist.
 	// The tablets we deploy take care of that automatically.
 	// However, only we know when it's time to remove a keyspace entry from topo,
 	// when the keyspace has been undeployed from this cell.
-	resultBuilder := &results.Builder{}
-	topoEntries = make([]string, 0, len(srvKeyspaceNames))
+	topoEntries := make([]string, 0, len(srvKeyspaceNames))
```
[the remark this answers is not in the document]
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/27e4fe47f37a4db7b797473754db02d44d8c9c4a.

PrismaPhonic:
@enisoc addressed all review comments. PTAL.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,43 +85,58 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
+	}
+
+	if *vt.Spec.TopologyReconciliation.RegisterCells {
+		result, err := r.registerCells(ctx, vt, ts, globalTopoImpl, desiredCells)
+		resultBuilder.Merge(result, err)
+	}
+
+	if *vt.Spec.TopologyReconciliation.PruneCells {
+		result, err := r.pruneCells(ctx, vt, ts, desiredCells)
+		if err != nil {
```
enisoc: I think we can remove this and just rely on the Merge. If there were cases inside `pruneCells` that used to be a requeue instead of an error, we should change `pruneCells` to requeue and return in those cases instead of returning an error.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,43 +85,58 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
-	// knows that it can route traffic between them.
-	// Currently, only one region is supported so we allow routing anywhere.
-	//
-	// We also need to create the aliases before we create the cells because we
-	// don't want any vtgates to start after the cells are created but before
-	// the alias exists.
-	desiredCellsAliases := buildCellsAliases(desiredCells)
-	currentCellsAliases, err := ts.GetCellsAliases(ctx, true)
+	if *vt.Spec.TopologyReconciliation.RegisterCellsAliases {
+		err := r.registerCellsAliases(ctx, vt, ts, desiredCells)
```
enisoc: Bring back the part of the old comment that explained why this has to happen before creating cells, since that explains why we bail out here whereas the other cases simply Merge and continue.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -85,43 +85,58 @@ func (r *ReconcileVitessCluster) reconcileCellTopology(ctx context.Context, vt *
...
+func (r *ReconcileVitessCluster) pruneCells(ctx context.Context, vt *planetscalev2.VitessCluster, ts *topo.Server, desiredCells map[string]*planetscalev2.VitessCellTemplate) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
+	// Get list of cells in topo.
+	cellNames, err := ts.GetCellInfoNames(ctx)
 	if err != nil {
-		r.recorder.Eventf(vt, corev1.EventTypeWarning, "TopoCellAlias",
-			"Failed to get current cell aliases: %v", err)
+		r.recorder.Eventf(vt, corev1.EventTypeWarning, "TopoListFailed", "failed to list cells in topology: %v", err)
 		return resultBuilder.Error(err)
```
enisoc: This used to be a return RequeueAfter so we should keep it that way.

pkg/controller/vitesscluster/reconcile_topo.go
```diff
@@ -190,6 +223,17 @@ func (r *ReconcileVitessCluster) reconcileKeyspaceTopology(ctx context.Context,
 		desiredKeyspaces[keyspace.Name] = keyspace
 	}
 
+	if *vt.Spec.TopologyReconciliation.PruneKeyspaces == true {
```
enisoc: It looks like we only use `desiredKeyspaces` inside `pruneKeyspaces` so we should move it there.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -36,64 +36,48 @@ const (
...
+
+func (r *ReconcileVitessCell) pruneSrvKeyspaces(ctx context.Context, vtc *planetscalev2.VitessCell, keyspaces []*planetscalev2.VitessKeyspace, ts *toposerver.Conn) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
 
 	// Get the list of keyspaces deployed (served) in this cell.
 	srvKeyspaceNames, err := ts.GetSrvKeyspaceNames(ctx, vtc.Spec.Name)
 	if err != nil {
 		r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoListFailed", "failed to list keyspaces in cell-local lockserver: %v", err)
-		return nil, err
+		return resultBuilder.Requeue()
```
enisoc: I meant `resultBuilder.RequeueAfter(topoRequeueDelay)` for all these. Sorry for not being clear. If we do just `Requeue()` it means "try again immediately", which will cause hot-looping whenever topo is down.

pkg/controller/vitesscell/reconcile_topo.go
```diff
@@ -36,64 +36,48 @@ const (
...
 		}
+
 		// It's not wanted. Try to delete it.
 		if err := ts.DeleteSrvKeyspace(ctx, vtc.Spec.Name, srvKeyspaceName); err != nil {
-			// We failed to delete it, so assume the entry still exists.
-			topoEntries = append(topoEntries, srvKeyspaceName)
-
 			r.recorder.Eventf(vtc, corev1.EventTypeWarning, "TopoCleanupBlocked", "unable to remove keyspace %s from cell-local topology: %v", srvKeyspaceName, err)
-			resultBuilder.Error(err)
+			resultBuilder.Requeue()
```
enisoc: `resultBuilder.RequeueAfter(topoRequeueDelay)`

pkg/controller/vitesscell/reconcile_keyspaces.go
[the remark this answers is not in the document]
enisoc: I'm still seeing this as Error instead of `resultBuilder.RequeueAfter(topoRequeueDelay)`.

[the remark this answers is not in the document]
PrismaPhonic: Huh, weird. I am too. I wonder why that didn't take - I tend to work very incrementally on PR review comments to avoid this.  Will fix, my apologies.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: I got you fam https://github.com/planetscale/vitess-operator/pull/26/commits/89d4572f7eda837d744c2fdb304573b5009f849d.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: I got you fam https://github.com/planetscale/vitess-operator/pull/26/commits/89d4572f7eda837d744c2fdb304573b5009f849d.

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: The comment was moved into inside of `registerCellsAliases` already. Is that sufficient?

pkg/controller/vitesscluster/reconcile_topo.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/94069f9c3becd597518b1d7f7407b37f51282dc2.

pkg/controller/vitesscell/reconcile_topo.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/af0eb34ebb84a1973e1cd5daec2a21eb3dc32cc5.

pkg/controller/vitesscell/reconcile_topo.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/af0eb34ebb84a1973e1cd5daec2a21eb3dc32cc5.

pkg/controller/vitesscell/reconcile_keyspaces.go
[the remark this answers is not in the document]
PrismaPhonic: Fixed https://github.com/planetscale/vitess-operator/pull/26/commits/f4cc06fdd60ffaef518214cd46379ab50e18d2fc.

PrismaPhonic:
@enisoc @avaidyanatha addressed all PR review feedback. PTAL.

pkg/controller/vitesscluster/reconcile_topo.go
enisoc: I was thinking that part of the comment should stay here because we need to explain why this looks different: here we return immediately (bail out) whereas below we merely Merge. It's harder to make the connection if the comment is buried in a subroutine.

PrismaPhonic: Alright, I'll move it then.

PrismaPhonic: If this must happen before creating cells, then theoretically we shouldn't registercells unless we also allow registration of cells aliases, correct?

PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/26/commits/ec1001eb2eee3336b9978ab3bd5a9536a2bcda21.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: This is fine with me. For some reason I assumed that we would need to disable all pruning for our use-case.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
avaidyanatha: Nope!

PR #29, opened 2020-01-25
Add integration test for topology reconciliation.

enisoc:
As part of this, I also had to add periodic resync to the other controllers; previously only some controllers had it. Without this periodic resync, the test would not reconcile topology within a reasonable time since the test modifies topo directly, instead of going through CRDs, in order to set up some special test cases.

Since we don't really need to recheck these objects often in a real cluster, I set the default resync to 30min and override it to 5s in the test environment. The 30min resync is also not completely useless in the real world; it acts as a safety net in case someone makes direct topo edits like this that the operator otherwise wouldn't notice.

### DIFF
--- a/test/integration/vitesscluster/topology_test.go
+++ b/test/integration/vitesscluster/topology_test.go
@@ -1,0 +1,233 @@
+/*
+Copyright 2020 PlanetScale Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vitesscluster
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+
+	topodatapb "vitess.io/vitess/go/vt/proto/topodata"
+	"vitess.io/vitess/go/vt/topo"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/test/integration/framework"
+)
+
+func TestTopology(t *testing.T) {
+	ctx := context.Background()
+
+	f := framework.NewFixture(ctx, t)
+	defer f.TearDown()
+
+	ns := "default"
+	cluster := "test-topology"
+
+	// Start with the same basic VitessCluster as the regular test.
+	vt := &planetscalev2.VitessCluster{}
+	framework.MustDecodeYAML(basicVitessCluster, vt)
+
+	// Set the cluster to use an external lockserver, namely the same etcd that
+	// we use for our test k8s apiserver.
+	//
+	// This lets us actually test the operator's topology interaction.
+	// Otherwise, there is no actual etcd since this integration test
+	// environment doesn't actually run any Pods.
+	params := &planetscalev2.VitessLockserverParams{
+		Implementation: "etcd2",
+		Address:        framework.EtcdURL(),
+		RootPath:       fmt.Sprintf("/vitess/%s/", cluster),
+	}
+	vt.Spec.GlobalLockserver.External = params
+
+	f.CreateVitessCluster(ns, cluster, vt)
+
+	ts, err := topo.OpenServer(params.Implementation, params.Address, params.RootPath)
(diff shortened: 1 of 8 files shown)


test/integration/vitesscluster/topology_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// Start with the same basic VitessCluster as the regular test.
+	vt := &planetscalev2.VitessCluster{}
+	framework.MustDecodeYAML(basicVitessCluster, vt)
+
+	// Set the cluster to use an external lockserver, namely the same etcd that
+	// we use for our test k8s apiserver.
+	//
+	// This lets us actually test the operator's topology interaction.
+	// Otherwise, there is no actual etcd since this integration test
+	// environment doesn't actually run any Pods.
```
PrismaPhonic: Is `Pods` intentionally capitalized in this comment?
```suggestion
	// environment doesn't actually run any pods.
```

test/integration/vitesscluster/topology_test.go
```diff
@@ -0,0 +1,233 @@
...
+	params := &planetscalev2.VitessLockserverParams{
+		Implementation: "etcd2",
+		Address:        framework.EtcdURL(),
+		RootPath:       fmt.Sprintf("/vitess/%s/", cluster),
```
PrismaPhonic: In case others want to add more paths onto the `RootPath`, would it be better to not have a tail forward slash? I'm just thinking about the format of an addition to the path. It would require that any additions do *not* start with a forward slash. Thoughts?

test/integration/vitesscluster/topology_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	vt.Spec.GlobalLockserver.External = params
+
+	f.CreateVitessCluster(ns, cluster, vt)
+
+	ts, err := topo.OpenServer(params.Implementation, params.Address, params.RootPath)
+	if err != nil {
+
```
PrismaPhonic: Is something supposed to happen here?

test/integration/vitesscluster/topology_test.go
```diff
@@ -0,0 +1,233 @@
...
+	})
+}
+
+func populateTopo(f *framework.Fixture, ts *topo.Server) {
+	// Add an orphan SrvKeyspace.
+	if err := ts.UpdateSrvKeyspace(f.Context(), "cell1", "delete_me", &topodatapb.SrvKeyspace{}); err != nil {
+		f.Fatalf("Can't create SrvKeyspace: %v", err)
+	}
+
+	// Add an orphan Cell.
```
PrismaPhonic: Same here. Is `Cell` intentionally capitalized? 
```suggestion
	// Add an orphan cell.
```

test/integration/vitesscluster/topology_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// We have to use a working address for the cell because Vitess tries to
+	// connect to the cell to verify it's empty before letting us delete it.
+	if err := ts.CreateCellInfo(f.Context(), "delete_me", &topodatapb.CellInfo{
+		ServerAddress: framework.EtcdURL(),
+		Root:          "/vitess/delete_me/",
```
PrismaPhonic: Same question here regarding tail `/`.

test/integration/vitesscluster/topology_test.go
enisoc: By habit I always capitalize k8s API objects like Pods, Nodes, and Services. This was beaten into me by tech writers who reviewed the docs I wrote for the k8s website. I can't stop now.

PrismaPhonic: Alright, fair enough.

test/integration/vitesscluster/topology_test.go
enisoc: The idiomatic way to join paths in Go is `path.Join()` (for abstract paths) or `filepath.Join()` (for filesystem paths) which ensures separators are consistent regardless of input. It's also a good practice to design interfaces that accept paths to work either with or without a trailing slash whenever possible. So if it matters what people give as input, we need to fix the receiver rather than the input.

test/integration/vitesscluster/topology_test.go
enisoc: This is inline Whitespace. j/k good catch.

test/integration/vitesscluster/topology_test.go
PrismaPhonic: Good to know, thanks!

test/integration/vitesscluster/topology_test.go
enisoc: Yeah it's intentional. It's sort of carrying over the documentation habit of capitalizing Official API Concepts into the Vitess world.

PR #31, opened 2020-01-29
Add Custom VitessShardCondition Object and Refactor Status for Use

PrismaPhonic:
This PR adds a custom VitessShardCondition object that follows the convention of k8s condition objects. The purpose of this addition is to allow external consumers to add any custom conditions they care about observing the state of in relation to Vitess shards.


pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -305,37 +305,68 @@ type VitessShardStatus struct {
 	// Cells is a list of cells in which any tablets for this shard are deployed.
 	Cells []string `json:"cells,omitempty"`
 
-	// HasMaster is a condition indicating whether the Vitess topology
```
enisoc: We can't get rid of these, unfortunately, because it's a breaking change. Upon upgrading the operator, existing clusters may be disrupted because one controller looks for the new conditions but they haven't been set yet by another controller. That could lead to bad things like thinking there is no master yet or that we haven't taken the initial backup yet.

The reason these were not full condition structs to begin with is because some in the k8s project considered status condition structs to be a mistake and that plain fields were better for anything that isn't "dynamic" - fields that you expect to always be there. I was trying to learn from those mistakes.

The reason we're now adding a conditions list is because we want to support dynamic conditions - things added by outside controllers that we allow and preserve. It's sort of like a more structured and narrow form of annotations.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -305,37 +305,68 @@ type VitessShardStatus struct {
...
+// VitessShardCondition contains details for the current condition of this VitessShard.
+type VitessShardCondition struct {
+	// Type is the type of the condition.
+	Type VitessShardConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=VitessShardConditionType"`
+	// Status is the status of the condition.
+	// Can be True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status,casttype=ConditionStatus"`
+	// Last time we probed the condition.
+	// +optional
+	LastProbeTime metav1.Time `json:"lastProbeTime,omitempty" protobuf:"bytes,3,opt,name=lastProbeTime"`
```
enisoc: Omit this field. It is deprecated and not recommended for new objects because it's too noisy.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -305,37 +305,68 @@ type VitessShardStatus struct {
...
+	// Last time the condition transitioned from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,4,opt,name=lastTransitionTime"`
```
enisoc: Since this is an optional struct, we should make it a pointer so `omitempty` actually works.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -305,37 +305,68 @@ type VitessShardStatus struct {
...
+	Idle VitessShardConditionType = "Idle"
+)
 
-	// BackupLocations reports information about the backups for this shard in
-	// each backup location.
-	BackupLocations []*ShardBackupLocationStatus `json:"backupLocations,omitempty"`
+// VitessShardCondition contains details for the current condition of this VitessShard.
+type VitessShardCondition struct {
+	// Type is the type of the condition.
+	Type VitessShardConditionType `json:"type" protobuf:"bytes,1,opt,name=type,casttype=VitessShardConditionType"`
```
enisoc: Let's leave off the protobuf tags. CRDs don't support protobuf anyway so it's just confusing. If they ever do support protobuf, this dark code suddenly being used might bite us since it could have bit-rotted by then.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -305,37 +305,68 @@ type VitessShardStatus struct {
...
 	Cells []string `json:"cells,omitempty"`
 
-	// HasMaster is a condition indicating whether the Vitess topology
-	// reflects a master for this shard.
-	HasMaster corev1.ConditionStatus `json:"hasMaster,omitempty"`
-	// HasInitialBackup is a condition indicating whether the initial backup
-	// has been seeded for the shard.
-	HasInitialBackup corev1.ConditionStatus `json:"hasInitialBackup,omitempty"`
+	// Conditions is a map of all VitessShard specific conditions we want to set and monitor.
+	Conditions map[VitessShardConditionType]*VitessShardCondition `json:"conditions,omitempty"`
```
enisoc: Status conditions are conventionally a list rather than a map. Maybe a map is better but we should think about whether it's worthwhile to be conventional, and also whether there is a reason the convention avoids maps.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -71,7 +71,7 @@ func (r *ReconcileVitessKeyspace) reconcileShards(ctx context.Context, vtk *plan
 
 			status := vtk.Status.Shards[curObj.Spec.KeyRange.String()]
 			status.Cells = curObj.Status.Cells
-			status.HasMaster = curObj.Status.HasMaster
+			status.HasMaster = curObj.Status.Conditions[planetscalev2.HasMaster].CurrentStatus()
```
enisoc: You can't assume the status of your children has been filled in already. That happens asynchronously when the child controller processes it. You might read it back before that happens.

PrismaPhonic: Reverted back to old API, and fixed so we only set it if the child object status has been set (not empty string).

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: We chatted about this offline, and it sounds like we're sticking to a map because it's a better choice and worth breaking with convention?

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: I had copied this from PodStatus object in k8s, so I assume there must be some valid empty state for a metav1.Time object? Or do you think their code was simply wrong?

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Will do.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Brought them all back and reverted code back to depending directly on these fields https://github.com/planetscale/vitess-operator/pull/31/commits/55d6ac920b9cd1e39e4d37bcda6d56a073a4c317.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Ok, will remove those.

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: Making it a non-pointer was a mistake that we were stuck with for compatibility reasons.

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: We'll do it live.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Removed https://github.com/planetscale/vitess-operator/pull/31/commits/4d12ec3679395986ef1d9ffbec7fd3feee478a36.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Updated https://github.com/planetscale/vitess-operator/pull/31/commits/c2ab065a32f85fea3a98ddaa240fc4192a920475.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Removed https://github.com/planetscale/vitess-operator/pull/31/commits/65eedd30e0b3e3ae634b2b1317934d450c66a926.

PrismaPhonic:
@enisoc all comments addressed. PTAL.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +153,47 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
 	_, inZoneMap := s.ZoneMap[cellName]
 	return inZoneMap
 }
+
+// NewVitessShardCondition returns a VitessShardCondition object based on the provided type and initial state.
+func NewVitessShardCondition(ty VitessShardConditionType, initState corev1.ConditionStatus) *VitessShardCondition {
+	now := v1.NewTime(time.Now())
+	return &VitessShardCondition{
+		Type:               ty,
+		Status:             initState,
```
enisoc: I think we should always make Status start as Unknown, and Reason and Message should be empty. That's the idiomatic "zero value" of a condition struct. This also would encourage all updates to go through ChangeStatus() rather than calling New and passing in a status but not a reason nor message.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +153,47 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
...
+		LastTransitionTime: &now,
+		Reason:             "initState",
+		Message:            "The initial state for this VitessShardCondition.",
+	}
+}
+
+// ChangeStatus changes the status if the current status is not the same as the new status, and updates the
+// last transition time.
+func (c *VitessShardCondition) ChangeStatus(newStatus corev1.ConditionStatus, reason, message string) {
```
enisoc: SetStatus feels more natural to me, but I don't feel strongly.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +153,47 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
...
+
+	c.Status = newStatus
+	now := v1.NewTime(time.Now())
+	c.LastTransitionTime = &now
+}
+
+// Duration returns the duration since LastTransitionTime. It represents how long we've been in the current status for
+// this condition.
+func (c *VitessShardCondition) Duration() time.Duration {
+	return time.Now().Sub(c.LastTransitionTime.Time)
```
enisoc: There's a nice helper that makes these easier to read:

```suggestion
	return time.Since(c.LastTransitionTime.Time)
```

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +153,47 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
...
+	}
+
+	c.Status = newStatus
+	now := v1.NewTime(time.Now())
+	c.LastTransitionTime = &now
+}
+
+// Duration returns the duration since LastTransitionTime. It represents how long we've been in the current status for
+// this condition.
+func (c *VitessShardCondition) Duration() time.Duration {
```
enisoc: To me, the name `Duration()` feels incomplete. A Condition doesn't feel like something that itself has a duration, so I'm left wondering, "Duration of what?" It's really the duration since the condition's status changed, so maybe `StatusDuration()`?

We should also do something about the case when LastTransitionTime is nil. It seems awkward to expect callers to check some other field (LastTransitionTime) for nil before calling this method, but I also don't know what this method should return if the time is nil (unknown). Any ideas?

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,24 +308,50 @@ type VitessShardStatus struct {
...
 	// has been seeded for the shard.
 	HasInitialBackup corev1.ConditionStatus `json:"hasInitialBackup,omitempty"`
+
 	// MasterAlias is the tablet alias of the master according to the global
 	// shard record. This could be empty either because there is no master,
 	// or because the shard record could not be read. Check the HasMaster
 	// condition whenever the distinction is important.
 	MasterAlias string `json:"masterAlias,omitempty"`
 
+	// BackupLocations reports information about the backups for this shard in
```
enisoc: Any reason for moving this? If anything, I'd probably leave this below Idle but move MasterAlias down so all the ConditionStatus-type fields are together.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,24 +308,50 @@ type VitessShardStatus struct {
...
+
+// VitessShardConditionType is a valid value for VitessShardCondition.Type
+type VitessShardConditionType string
+
+// VitessShardCondition contains details for the current condition of this VitessShard.
+type VitessShardCondition struct {
+	// Type is the type of the condition.
+	Type VitessShardConditionType `json:"type"`
+	// Status is the status of the condition.
+	// Can be True, False, Unknown.
```
enisoc: Perhaps we should enforce this with an enum validation, especially since we're opening this up for dynamic use by other controllers. If any other values sneak in, it could cause difficult-to-diagnose bugs.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,24 +308,50 @@ type VitessShardStatus struct {
...
+	Conditions map[VitessShardConditionType]*VitessShardCondition `json:"conditions,omitempty"`
+}
+
+// VitessShardConditionType is a valid value for VitessShardCondition.Type
+type VitessShardConditionType string
+
+// VitessShardCondition contains details for the current condition of this VitessShard.
+type VitessShardCondition struct {
+	// Type is the type of the condition.
+	Type VitessShardConditionType `json:"type"`
```
enisoc: This is mostly redundant since we decided to use a map. I know it's just following the convention, but we already broke with convention by using a map, so I think we should only include this if it makes sense from first principles in a map-based design.

Do you think it's generally worthwhile to duplicate the map key inside the structured value of an entry? In my experience, there are pros and cons of both options.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,24 +308,50 @@ type VitessShardStatus struct {
...
+	// Status is the status of the condition.
+	// Can be True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status"`
+	// Last time the condition transitioned from one status to another.
+	// +optional
```
enisoc: These `+optional` tags also don't do anything, I believe, since we don't run the same tooling as core k8s. Let's remove them to avoid the possibility of getting out of sync.

pkg/controller/vitessshard/vitessshard_controller.go
```diff
@@ -171,6 +171,8 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
 	// Reset status, since that's all out of date info that we will recompute now.
 	oldStatus := vts.Status
 	vts.Status = planetscalev2.NewVitessShardStatus()
+	// Make sure we don't rinse old conditions - these states need to persist.
```
enisoc: Add justification for why we need to persist these: because they may have been added by some other controller.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -18,7 +18,9 @@ package v2
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"time"
```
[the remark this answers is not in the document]
PrismaPhonic: Fixed https://github.com/planetscale/vitess-operator/pull/31/commits/e3c8b88175e50e08664a032b57fb26d5423df0d3.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -18,7 +18,9 @@ package v2
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
PrismaPhonic: Adjusted https://github.com/planetscale/vitess-operator/pull/31/commits/e3c8b88175e50e08664a032b57fb26d5423df0d3.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: I had named it `ChangeStatus` because it bailed immediately if the statuses were the same. Now that we set the reason and message every time regardless, I think `SetStatus` is a better name. I'll change it.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Good suggestion. Changed: https://github.com/planetscale/vitess-operator/pull/31/commits/19a40e6a420384f2e9fe11fe824dc076054f8df1.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: If you compare against the original file it's actually not moved. In Master right now BackupLocations is directly underneath MasterAlias.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: That's a good name suggestion - I'll change it. As for LastTransitionTime being nil, I think we should return a Duration of 0.

PrismaPhonic: Method name update: https://github.com/planetscale/vitess-operator/pull/31/commits/89eb420dbd304ab7b6a9b440d46fc4432a93127f.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Updated: https://github.com/planetscale/vitess-operator/pull/31/commits/89eb420dbd304ab7b6a9b440d46fc4432a93127f.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: I think that's a good idea. I'll add Kubebuilder enum validation.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/31/commits/7090ba557e9ebe152b564df4b9180278be40ee57.

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Added nil check and zero return https://github.com/planetscale/vitess-operator/pull/31/commits/7090ba557e9ebe152b564df4b9180278be40ee57.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/31/commits/5011b4b6b77b9d5a6a9d82a45cc04ce04ff3e14c.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: I think if we don't care about conventions at all, then it doesn't provide any extra benefit. In fact, having it seems to encourage mutability, and it would be very odd if the type of a condition mutated but its key stays the same. So I would be in favor of removing it if you are as well. It seems error-prone. WDYT?

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: In our codebase, I've often seen `Optional.` on its own line. Do you think it would be useful to change to that?

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: It's at the bottom on master: https://github.com/planetscale/vitess-operator/blob/master/pkg/apis/planetscale/v2/vitessshard_types.go#L326-L328

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: OK let's remove then.

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: sgtm

pkg/controller/vitessshard/vitessshard_controller.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/31/commits/448517eb5d5944f4f2749ee076669f504707f29b.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: oh, was apparently not looking at master then. My bad - I'll move it back. No reason why it got moved. Think it was the shuffle of removing conditions and then puttting them back.

PrismaPhonic: Moved back to the bottom and grouped all condition fields together https://github.com/planetscale/vitess-operator/pull/31/commits/c192f7b252505bee0b39f77b9bdb99cb235632fd.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/31/commits/bf051469919e3a3e7686f689c0766983e7852b11.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Removed https://github.com/planetscale/vitess-operator/pull/31/commits/ef506e72fe8f258c9628fd1d9d2c9d15c5f12acc.

PrismaPhonic:
@enisoc Addressed all changes, PTAL.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +154,51 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
...
+	// We should update reason and message regardless of whether the status type is different.
+	c.Reason = reason
+	c.Message = message
+	if c.Status == newStatus {
+		return
+	}
+
+	c.Status = newStatus
+	now := metav1.NewTime(time.Now())
+	c.LastTransitionTime = &now
```
avaidyanatha: This may never happen, but is it possible for the status to be set and then fail before updating the `LastTransitionTime`?

deploy/crds/planetscale_v2_vitessshard_crd.yaml
```diff
@@ -679,6 +679,34 @@ spec:
               items:
                 type: string
               type: array
+            conditions:
+              additionalProperties:
+                properties:
```
avaidyanatha: The `additionalProperties` -> `properties` tree confuses me a little, it feels like `additionalProperties` should be renamed or removed (I assume it serves a purpose though)

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: If it did then we would panic. I would say this is an extremely low likelihood. In general, the question is a chicken or an egg problem. If we set the transition time first and then somehow we panic, we are stuck with a `LastTransitionTime` that is updated but an object that hasn't been updated yet. In general, I don't think this kind of an issue is worth worrying about. I would surmise that the number of cases where this causes a problem are next to none even under high use.

deploy/crds/planetscale_v2_vitessshard_crd.yaml
PrismaPhonic: This is auto-generated :-P

pkg/apis/planetscale/v2/vitessshard_methods.go
PrismaPhonic: Although, I could move grabbing the time to happen right before we set status, so if somehow grabbing time causes a panic, then both setting status and last transition fail.

PrismaPhonic: Moved up one line: https://github.com/planetscale/vitess-operator/pull/31/commits/05a44f961b1b6cf6603cf00ffc4232819a6d767d.

enisoc: This is just filling an in-memory struct with proposed new values. Ultimately all that matters is the state of the object in the apiserver. When we later make the UpdateStatus call, the server will ensure that the object is updated atomically - either all fields will be updated or none of them will be.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,26 +308,52 @@ type VitessShardStatus struct {
...
-	// condition whenever the distinction is important.
-	MasterAlias string `json:"masterAlias,omitempty"`
 
 	// Idle is a condition indicating whether the shard can be turned down.
 	// If Idle is True, the shard is not part of the active shard set
 	// (partitioning) for any tablet type in any cell, so it should be safe
 	// to turn down the shard.
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
 
+	// Conditions is a map of all VitessShard specific conditions we want to set and monitor.
```
enisoc: Let's document the API guarantee that it's ok for other controllers to add conditions here, and we will preserve them, since we are going to rely on that guarantee elsewhere.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -308,26 +308,52 @@ type VitessShardStatus struct {
...
 
+// VitessShardConditionType is a valid value for the key of a VitessShardCondition map where the key is a
+// VitessShardConditionType and the value is a VitessShardCondition.
+type VitessShardConditionType string
+
+// VitessShardCondition contains details for the current condition of this VitessShard.
+type VitessShardCondition struct {
+	// Status is the status of the condition.
+	// Can be True, False, Unknown.
+	// +kubebuilder:validation:Enum=True;False;Unknown
```
enisoc: The values should be comma-separated.

```suggestion
	// +kubebuilder:validation:Enum=True,False,Unknown
```

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -151,3 +154,52 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
...
 	return inZoneMap
 }
+
+// NewVitessShardCondition returns an init VitessShardCondition object.
+func NewVitessShardCondition() *VitessShardCondition {
+	now := metav1.NewTime(time.Now())
+	return &VitessShardCondition{
+		Status:             corev1.ConditionUnknown,
+		LastTransitionTime: &now,
+		Reason:             "initState",
```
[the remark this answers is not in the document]
PrismaPhonic: The comment documentation around `PodCondition` says: `"Unique, one-word, CamelCase reason for the condition's last transition."`. The really odd thing is that they put `CamelCase` in `PascalCase` in their code comment. Is this just a mistake on their part? For reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#podcondition-v1-core. Did they really mean `PascalCase` even though it says `CamelCase`?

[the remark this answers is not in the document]
PrismaPhonic: I'm happy to leave them empty - I think that's a good point. Will update. If they truly meant `PascalCase` then I would like to update my code comment above the `Reason` field in `VitessShardCondition`.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Good point. Will add that now.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: According to the official documentation, they should be semicolon-separated. For reference, their example for `type Alias string`: https://book.kubebuilder.io/reference/generating-crd.html

PrismaPhonic: Ahh, this totally different example is comma-separated: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html

PrismaPhonic: I'm really confused. Is the first example just flat wrong from their docs?

PrismaPhonic: Here's another one that's semicolon-separated. search for `type ConcurrencyPolicy string`: https://book.kubebuilder.io/cronjob-tutorial/api-design.html

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Added https://github.com/planetscale/vitess-operator/pull/31/commits/3851de380e22eed9628ca7734f94443f22db3c52.

pkg/apis/planetscale/v2/vitessshard_methods.go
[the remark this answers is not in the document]
PrismaPhonic: Switched them to empty strings: https://github.com/planetscale/vitess-operator/pull/31/commits/66606939f676555f3314a2c94eff114f82970a3d.

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: My guess is they changed the format in a later version of kubebuilder, but we are still using an old version. In any case, you can see from the generated output that semicolon doesn't work with our current generator version. I guess we'll have to keep an eye on this when we upgrade to a newer operator-sdk (and hence pick up a new kubebuilder version) since it looks like a breaking change.

pkg/apis/planetscale/v2/vitessshard_methods.go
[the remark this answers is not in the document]
enisoc: The docs intended for `CamelCase` to imply upper camel case, which is a variant of the original camel case that starts with capital letters (equivalent to PascalCase). There is unfortunately not strong agreement on what "camel case" means in terms of the starting letter, so I agree that PascalCase is a less ambiguous way to describe it.

deploy/crds/planetscale_v2_vitessshard_crd.yaml
```diff
@@ -679,6 +679,35 @@ spec:
...
+                    type: string
+                  reason:
+                    description: Unique, one-word, CamelCase reason for the condition's
+                      last transition. Optional.
+                    type: string
+                  status:
+                    description: Status is the status of the condition. Can be True,
+                      False, Unknown.
+                    enum:
+                    - True;False;Unknown
```
enisoc: FYI this is how I can tell that semicolons don't work in this version of the generator. Each of these should be its own list item in the generated output.

PrismaPhonic: Got it, will update.

pkg/apis/planetscale/v2/vitessshard_methods.go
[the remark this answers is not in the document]
PrismaPhonic: Alright, I'll update the doc comment as well then.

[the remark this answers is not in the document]
PrismaPhonic: Updated comment for clarity https://github.com/planetscale/vitess-operator/pull/31/commits/ce1844e7eed81af4c78a7e26b1f70670214dfe82.

deploy/crds/planetscale_v2_vitessshard_crd.yaml
PrismaPhonic: Updated https://github.com/planetscale/vitess-operator/pull/31/commits/20698920e71c38d09894ba8b1dfadbe395d879d26.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Updated https://github.com/planetscale/vitess-operator/pull/31/commits/20698920e71c38d09894ba8b1dfadbe395d879d2.

PR #32, opened 2020-01-29
Only prune ShardCells for known/local cells.

enisoc:
This makes it safe to leave ShardCell pruning enabled in the case when multiple vitess-operator instances are combining to manage pieces of a multi-region Vitess cluster.

Each operator instance knows which Cells it should handle, so it can prune ShardCells corresponding to those Cells.

### DIFF
--- a/pkg/controller/vitessshard/reconcile_topo.go
+++ b/pkg/controller/vitessshard/reconcile_topo.go
@@ -137,16 +137,25 @@ func (r *ReconcileVitessShard) pruneShardCells(ctx context.Context, vts *planets
 
 	// Clean up cells from the shard record that we don't deploy to anymore.
 	for _, cellName := range servingCells {
-		if !topo.InCellList(cellName, vts.Status.Cells) {
-			// The cell is listed in topo, but we don't deploy there anymore.
-			// We use the Vitess wrangler (multi-step command executor) to remove the cell from that shard.
-			// This is equivalent to `vtctl RemoveShardCell`.
-			if err := wr.RemoveShardCell(ctx, keyspaceName, vts.Spec.Name, cellName, false /* force*/, false /* recursive */); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/vitessshard/reconcile_topo.go
```diff
@@ -137,16 +137,24 @@ func (r *ReconcileVitessShard) pruneShardCells(ctx context.Context, vts *planets
...
-			// The cell is listed in topo, but we don't deploy there anymore.
-			// We use the Vitess wrangler (multi-step command executor) to remove the cell from that shard.
-			// This is equivalent to `vtctl RemoveShardCell`.
-			if err := wr.RemoveShardCell(ctx, keyspaceName, vts.Spec.Name, cellName, false /* force*/, false /* recursive */); err != nil {
-				r.recorder.Eventf(vts, corev1.EventTypeWarning, "TopoCleanupFailed", "unable to remove cell %s from shard: %v", cellName, err)
-				resultBuilder.RequeueAfter(topoRequeueDelay)
-			} else {
-				r.recorder.Eventf(vts, corev1.EventTypeNormal, "TopoCleanup", "removed unwanted cell %s from shard", cellName)
-			}
+		if _, inZoneMap := vts.Spec.ZoneMap[cellName]; !inZoneMap {
```
PrismaPhonic: The only issue I have with this approach is that while it's technically correct, it relies on knowledge that the ZoneMap is formed by a relationship between cell names to cell zones that gets generated at the cluster spec level and then passed down. The comment below helps some, but on the first read, it's not obvious what's happening here.

enisoc: Added helper method as discussed offline. PTAL.

PrismaPhonic: LGTM. Approved and merged.

PR #33, opened 2020-01-31
Refactor GlobalConnectionParams to Take Shared Types

PrismaPhonic:
This PR refactors the `GlobalConnectionParams` function to take shared types. In PlanetScale's proprietary operator we have an analog to `VitessCluster` that extends the functionality of `VitessCluster`. Translating from the native parent type in that operator to a `VitessCluster` type just to use this helper function is expensive. To avoid code duplication I'm adjusting the helper method to take shared types rather than the entire root type of `VitessCluster`.

### DIFF
--- a/pkg/controller/vitesscluster/reconcile_vtctld.go
+++ b/pkg/controller/vitesscluster/reconcile_vtctld.go
@@ -18,6 +18,7 @@ package vitesscluster
 
 import (
 	"context"
+
 	"planetscale.dev/vitess-operator/pkg/operator/update"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -133,7 +134,7 @@ func (r *ReconcileVitessCluster) vtctldSpecs(vt *planetscalev2.VitessCluster, pa
 		}
 	}
 
-	glsParams := lockserver.GlobalConnectionParams(vt)
+	glsParams := lockserver.GlobalConnectionParams(&vt.Spec.GlobalLockserver, vt.Name)
 
 	// Make a vtctld Deployment spec for each cell.
 	specs := make([]*vtctld.Spec, 0, len(cells))
--- a/pkg/operator/lockserver/params.go
+++ b/pkg/operator/lockserver/params.go
@@ -33,15 +33,15 @@ const (
 
 // GlobalConnectionParams returns the Vitess connection parameters for a
 // VitessCluster's global lockserver.
-func GlobalConnectionParams(vt *planetscalev2.VitessCluster) *planetscalev2.VitessLockserverParams {
+func GlobalConnectionParams(lockSpec *planetscalev2.LockserverSpec, name string) *planetscalev2.VitessLockserverParams {
 	switch {
-	case vt.Spec.GlobalLockserver.External != nil:
-		return vt.Spec.GlobalLockserver.External
-	case vt.Spec.GlobalLockserver.Etcd != nil:
+	case lockSpec.External != nil:
+		return lockSpec.External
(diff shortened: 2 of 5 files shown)


pkg/operator/lockserver/params.go
```diff
@@ -33,15 +33,15 @@ const (
 
 // GlobalConnectionParams returns the Vitess connection parameters for a
 // VitessCluster's global lockserver.
-func GlobalConnectionParams(vt *planetscalev2.VitessCluster) *planetscalev2.VitessLockserverParams {
+func GlobalConnectionParams(lockSpec *planetscalev2.LockserverSpec, name string) *planetscalev2.VitessLockserverParams {
```
enisoc: Let's be more clear what name this refers to:

```suggestion
func GlobalConnectionParams(lockSpec *planetscalev2.LockserverSpec, clusterName string) *planetscalev2.VitessLockserverParams {
```

pkg/controller/vitesscluster/reconcile_vtctld.go
```diff
@@ -18,6 +18,7 @@ package vitesscluster
 
 import (
 	"context"
+
```
avaidyanatha: I assume your IDE did this?

pkg/operator/lockserver/params.go
avaidyanatha: Is this pointer necessary?

pkg/controller/vitesscluster/reconcile_vtctld.go
PrismaPhonic: Yep. Due to the new sorting stuff we turned on that keeps stdlib as its own group at the top.

pkg/operator/lockserver/params.go
PrismaPhonic: I might be wrong, but I believe without a pointer then we have copy semantics, meaning the struct is copied when the function is called. The pointer eliminates unnecessary struct copying. Perhaps though with such a small struct it doesn't save us much in terms of memory and might be slower due to heap allocation. @enisoc, thoughts on this?I might be wrong, but I believe without a pointer then we have copy semantics, meaning the struct is copied when the function is called. The pointer eliminates unnecessary struct copying. Perhaps though with such a small struct it doesn't save us much in terms of memory and might be slower due to heap allocation. @enisoc, thoughts on this?

PrismaPhonic: Good idea, will update.

PrismaPhonic: Updated https://github.com/planetscale/vitess-operator/pull/33/commits/bc4749d003bca3d7843a076883ed0d9ce37495c4.

pkg/controller/vitesscluster/reconcile_vtctld.go
enisoc: It's a good thing.

pkg/operator/lockserver/params.go
enisoc: I prefer to pass k8s API structs as pointers in most cases, if only because these structs very often contain pointers themselves, which means a shallow copy only sets you up to mistakenly believe you have a copy when you actually don't. It's also a habit because the scalability team in k8s hounded us to avoid copies in the core controllers, since they identified the copies and extra GC load as being a problem for controllers that handle lots of objects.

PrismaPhonic: Sounds like this was the right approach then. I'll mark this resolved for now and go ahead and merge.

PR #36, opened 2020-02-07
Make Conditions Access Safe

PrismaPhonic:
I should have already done this in my previous PR where I introduced the `VitessShardCondition` type. We need to allocate a map for the Conditions field in the defaulting code so access is safe.

Note that the intended design is for consumers to be able to use the Conditions field (but no other part of the status object). This means that we might have a race condition where another controller tries to access the conditions before this controller has allocated a map for them.

That means that it's up to a consumer to nil check the conditions field before trying to access or modify conditions.

One idea I have for that, is that in any controller trying to use conditions, they must have a line such as this that happens before attempted access of conditions:

```go
	// Add empty condition map if condition map isn't there already.
	if vts.Status.Conditions == nil {
		vts.Status.Conditions = make(map[v2.VitessShardConditionType]*v2.VitessShardCondition)
	}
```

Thoughts on best practice for handling this in other controllers? Another strategy is that we force consumers to use methods on the status object itself. Those getters/setters could nil check the conditions field and allocate a new map if nil.

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -363,6 +363,7 @@ func NewVitessShardStatus() VitessShardStatus {
 		HasMaster:        corev1.ConditionUnknown,
 		HasInitialBackup: corev1.ConditionUnknown,
 		Idle:             corev1.ConditionUnknown,
+		Conditions:       make(map[VitessShardConditionType]*VitessShardCondition),
 	}
 }
 
--- a/pkg/apis/planetscale/v2/vitessshard_methods.go
+++ b/pkg/apis/planetscale/v2/vitessshard_methods.go
@@ -155,6 +155,32 @@ func (s *VitessShardSpec) CellInCluster(cellName string) bool {
 	return inZoneMap
 }
 
+// SetConditionStatus first ensures we have allocated a conditions map, and also ensures we have allocated a ShardCondition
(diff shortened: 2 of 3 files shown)


PrismaPhonic:
I just realized that another issue is if we have multiple consumers trying to allocate an empty map on nil, there might be a race condition where one succeeds and then very quickly tries to insert a condition, and then gets overwritten by another controller that also allocates an empty map. Maybe the best approach then in other controllers is to nil check, and if nil, re-queue but don't attempt any more of the reconcile loop that would attempt access.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -363,6 +363,7 @@ func NewVitessShardStatus() VitessShardStatus {
 		HasMaster:        corev1.ConditionUnknown,
 		HasInitialBackup: corev1.ConditionUnknown,
 		Idle:             corev1.ConditionUnknown,
+		Conditions:       make(map[VitessShardConditionType]*VitessShardCondition),
```
enisoc: This makes sense at a conceptual level (`NewX()` should allocate the map for you), but it actually doesn't have any practical effect right now. We currently only call `NewVitessShardStatus()` in one place, and then immediately overwrite `Conditions` with something else.

PrismaPhonic: Oh hmm, that's true I suppose. What do you suggest then as a better approach?

PR #41, opened 2020-02-19
Custom Condition Max Durations

PrismaPhonic:
This PR adds an optional field for consumers to provide max durations for known `VitessShardConditionType`s. An example has been shown of checking for the existence of a max duration. The logic is that if there doesn't exist a max duration for a condition type, then we assume there is no max duration and we should always act immediately on repair. If a max duration is found, then we check that the conditions duration has exceeded the max duration before attempting to repair.

### DIFF
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -141,7 +141,7 @@ func newVitessShard(key client.ObjectKey, vtk *planetscalev2.VitessKeyspace, par
 			BackupEngine:           vtk.Spec.BackupEngine,
 			ExtraVitessFlags:       vtk.Spec.ExtraVitessFlags,
 			TopologyReconciliation: vtk.Spec.TopologyReconciliation,
-			MaxConditionDurations:  make(map[planetscalev2.VitessShardConditionType]metav1.Duration),
+			ConditionMaxDurations:  make(map[planetscalev2.VitessShardConditionType]metav1.Duration),
 		},
 	}
 }
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -141,6 +141,7 @@ func newVitessShard(key client.ObjectKey, vtk *planetscalev2.VitessKeyspace, par
 			BackupEngine:           vtk.Spec.BackupEngine,
 			ExtraVitessFlags:       vtk.Spec.ExtraVitessFlags,
 			TopologyReconciliation: vtk.Spec.TopologyReconciliation,
+			MaxConditionDurations:  make(map[planetscalev2.VitessShardConditionType]metav1.Duration),
 		},
 	}
 }
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -141,7 +141,7 @@ func newVitessShard(key client.ObjectKey, vtk *planetscalev2.VitessKeyspace, par
 			BackupEngine:           vtk.Spec.BackupEngine,
 			ExtraVitessFlags:       vtk.Spec.ExtraVitessFlags,
 			TopologyReconciliation: vtk.Spec.TopologyReconciliation,
-			ConditionMaxDurations:  make(map[planetscalev2.VitessShardConditionType]metav1.Duration),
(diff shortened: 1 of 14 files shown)


pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -141,6 +141,7 @@ func newVitessShard(key client.ObjectKey, vtk *planetscalev2.VitessKeyspace, par
 			BackupEngine:           vtk.Spec.BackupEngine,
 			ExtraVitessFlags:       vtk.Spec.ExtraVitessFlags,
 			TopologyReconciliation: vtk.Spec.TopologyReconciliation,
+			MaxConditionDurations:  make(map[planetscalev2.VitessShardConditionType]metav1.Duration),
```
avaidyanatha: Nit: I think it reads better as `ConditionMaxDurations`

pkg/controller/vitessshardreplication/repair_replication.go
```diff
@@ -151,7 +151,13 @@ func (r *ReconcileVitessShard) canRepairReplication(ctx context.Context, vts *pl
...
 	}
 	if masterReadOnly {
-		return true, nil
+		vts.Status.SetConditionStatus(planetscalev2.ReadOnlyMasterConditionType, corev1.ConditionTrue, "MasterReadOnlyDetected", "Master ReadOnly State Detected.")
+		shouldFix := masterRdOnlyShouldBeFixed(vts)
+		if shouldFix {
+			return true, nil
+		}
+	} else {
+		vts.Status.SetConditionStatus(planetscalev2.ReadOnlyMasterConditionType, corev1.ConditionFalse, "MasterNotReadOnly", "Master is not in ReadOnly State.")
```
avaidyanatha: Maybe I'm misunderstanding, but why do we have a false condition for the master not being in read only? Shouldn't we just need to check that the master doesn't have a "MasterReadOnly" condition?

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: I don't feel strongly either way so I'm happy to change it.

pkg/controller/vitessshardreplication/repair_replication.go
```diff
@@ -187,6 +193,21 @@ func (r *ReconcileVitessShard) canRepairReplication(ctx context.Context, vts *pl
 	return foundFixableTablet, nil
 }
 
+func masterRdOnlyShouldBeFixed(vts *planetscalev2.VitessShard) bool {
```
avaidyanatha: Check that the condition exists at some point just to be safe.

PrismaPhonic: Can do.

pkg/controller/vitessshardreplication/repair_replication.go
PrismaPhonic: We could either clear the condition entirely from the map, or set it to ConditionFalse. The convention as far as I know is to set the conditions to false when they are false rather than clear them entirely.

pkg/controller/vitessshardreplication/repair_replication.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/41/commits/b643bd6815b6c3128cea04b544dfb204c7a9b5dc.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Updated: https://github.com/planetscale/vitess-operator/pull/41/commits/f9dc9ba9dab7e0588af7a5bea78c9759dabbd61d.

enisoc:
@PrismaPhonic Are you ok closing this based on our offline conversation?

To summarize the discussion for posterity: I don't think we should center the UX for configuring settings of this form ("wait a bit before acting on an observed state") around the concept of k8s status conditions. Instead, we should express those settings in the natural language of a given domain, such as a delay field within the part of a VT config where you enable a given feature like replication repair. The fact that we use k8s status conditions to implement the delay is an implementation detail that users shouldn't have to think about.

PrismaPhonic:
@enisoc I'm fine with closing this for now as long as the branch doesn't get deleted. For now, I need the remote branch to stay alive for some upstream work that currently depends on it. I can delete the remote branch myself once we migrate to static fields in structs.

PR #44, opened 2020-02-28
Add support for pod labels

### LINKED ISSUE #35: Allow extra Labels to be specified like the Annotations
We currently support adding arbitrary Annotations to be added to all the components. In our company we require having specific labels added to our K8s pods to indicate things like the owning team names. 

Can we follow the annotations pattern and allow another map[string][string] for labels as well?

evalsocket:
## Main Changes

    Adding support for applying custom pod labels through the CRD

Fixed #35


pkg/controller/etcdlockserver/reconcile_members.go
```diff
@@ -129,6 +130,7 @@ func memberSpecs(ls *planetscalev2.EtcdLockserver, parentLabels map[string]strin
 		if _, hasClusterLabel := ls.Labels[planetscalev2.ClusterLabel]; hasClusterLabel {
 			labels[planetscalev2.ClusterLabel] = ls.Labels[planetscalev2.ClusterLabel]
 		}
+		update.Labels(&labels, ls.Spec.UserLabels)
```
avaidyanatha: Since we have a new field for this, add this in the append statement below `Annotations` like so:

```
Annotations:       ls.Spec.Annotations,
UserLabels:        ls.Spec.UserLabels,
		})
```

Do the same for the function in reconcile_tablets.go and the vtgate/vtctld deployments.

avaidyanatha:
As a reminder, with my above suggested changes, you'll need to change the `UpdatePod` and `UpdateDeployment` functions to use the new UserLabels, otherwise they will just remain unused.

For example, here is the `UpdatePod` function for the etcd members: https://github.com/planetscale/vitess-operator/blob/e57a7a97e71e8a4c7b4e5b1e55ad918352aec3ab/pkg/operator/etcd/pod.go#L110

You would need to merge `UserLabels` and `Labels` here. However, since we would called `UpdateLabels()` to merge them, old extraneous labels may still exist, so we want to force a pod restart. We do this by updating a "key-hash" whenever other labels change as done here: https://github.com/planetscale/vitess-operator/blob/1a01fc3f9f68c27fcaa80d7c3a5b74626f90e8fc/pkg/operator/etcd/pod.go#L113

This pattern only needs to be done for Pods not created by a Deployment, as Deployments automatically perform the restarts for us on label/annotation change.

pkg/operator/vttablet/pod.go
```diff
@@ -70,6 +70,8 @@ func UpdatePodInPlace(obj *corev1.Pod, spec *Spec) {
 func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
+	// Update UserLabels, but ignore existing ones we don't set.
+	update.Labels(&obj.Labels, spec.UserLabels)
```
avaidyanatha:
```suggestion
// Record a hash of desired label keys to force the Pod
// to be recreated if a key disappears from the desired list.
update.Labels(&obj.Annotations, map[string]string{
	"planetscale.com/labels-keys-hash": contenthash.StringMapKeys(spec.UserLabels),
})
```

pkg/operator/vtgate/deployment.go
```diff
@@ -106,6 +107,10 @@ func UpdateDeployment(obj *appsv1.Deployment, spec *Spec) {
 	update.Labels(&obj.Labels, spec.Labels)
 	// Tell Deployment to set the same labels on the Pods it creates.
 	update.Labels(&obj.Spec.Template.Labels, spec.Labels)
+	// Set user labels on the Deployment object.
```
avaidyanatha: Since we only care about setting labels on the pods that the user creates, we don't need to set user labels on the Deployment object, this can be removed.

pkg/operator/vtctld/deployment.go
```diff
@@ -93,6 +94,9 @@ func UpdateDeploymentImmediate(obj *appsv1.Deployment, spec *Spec) {
 	// Set labels on the Deployment object.
 	update.Labels(&obj.Labels, spec.Labels)
 
+	// Set user labels on the Deployment object.
```
avaidyanatha: Move this logic into `UpdateDeployment`, this is a change that we want to put behind a rollout and isn't necessarily safe to change immediately. Additionally, similarly to what I suggested with the vtgate deployment logic, we only need to ask our deployment to set `UserLabels` on the pods, not the deployment object.

pkg/operator/etcd/pod.go
```diff
@@ -110,6 +111,8 @@ func UpdatePodInPlace(obj *corev1.Pod, spec *Spec) {
 func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
+	// Update UserLabels, but ignore existing ones we don't set.
+	update.Labels(&obj.Labels, spec.UserLabels)
```
avaidyanatha:
```suggestion
// Record a hash of desired label keys to force the Pod
// to be recreated if a key disappears from the desired list.
update.Labels(&obj.Annotations, map[string]string{
	"planetscale.com/labels-keys-hash": contenthash.StringMapKeys(spec.UserLabels),
})
```

pkg/operator/vttablet/pod.go
```diff
@@ -71,6 +71,12 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
```
avaidyanatha:
```suggestion
// Update desired labels
	update.Labels(&obj.Labels, spec.UserLabels)
```

pkg/operator/etcd/pod.go
```diff
@@ -111,6 +112,12 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
```
avaidyanatha:
```suggestion
// Update desired labels
	update.Labels(&obj.Labels, spec.UserLabels)
```

PR #46, opened 2020-03-04
Only repair replication in the region containing the current master.

enisoc:
When deploying multiple VitessCluster objects (e.g. across multiple Kubernetes clusters) that join into a single Vitess topology, we only want one operator instance to attempt replication repair.

### DIFF
--- a/pkg/controller/vitessshardreplication/repair_replication.go
+++ b/pkg/controller/vitessshardreplication/repair_replication.go
@@ -84,6 +84,19 @@ func (r *ReconcileVitessShard) repairReplication(ctx context.Context, vts *plane
 		return resultBuilder.Result()
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/vitessshardreplication/repair_replication.go
```diff
@@ -84,6 +84,19 @@ func (r *ReconcileVitessShard) repairReplication(ctx context.Context, vts *plane
 		return resultBuilder.Result()
 	}
 
+	// If the current master is in a cell that this VitessCluster doesn't
```
avaidyanatha: In theory, couldn't we just place a lock on replication repair if one of the operators gets to it? Or do we not want the cells to interact at all? Otherwise, LGTM.

PR #47, opened 2020-03-06
Changes to support azblob backup storage engine

dkhenry:
This adds an azblob field to the storage spec. It also implements the
azblob backup flags and adds the ENV vars for the account name and key


pkg/operator/vitessbackup/storage.go
```diff
@@ -42,6 +42,8 @@ func StorageFlags(backupLocation *planetscalev2.VitessBackupLocation, clusterNam
 		return gcsBackupFlags(backupLocation.GCS, clusterName)
 	case backupLocation.S3 != nil:
 		return s3BackupFlags(backupLocation.S3, clusterName)
+	case backupLocation.Azblob != nil:
+		return azblobBackupFlags(backupLocation.Azblob, clusterName)
```
avaidyanatha: Are these functions from something that hasn't been merged yet?

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -67,6 +67,8 @@ type VitessBackupLocation struct {
 	GCS *GCSBackupLocation `json:"gcs,omitempty"`
 	// S3 specifies a backup location in Amazon S3.
 	S3 *S3BackupLocation `json:"s3,omitempty"`
+	// Azblob specifies a backup location in Azure Blob Storage.
+	Azblob *AzblobBackupLocation `json:"azblob,omitempty"`
```
avaidyanatha: Nit - Probably worth spelling it out, even though Azblob sounds great.
```suggestion
	AzureBlob *AzureBlobBackupLocation `json:"azureblob,omitempty"`
```

pkg/operator/vitessbackup/storage.go
dkhenry: Woops didn't add the file, yes comming in on this Commit now

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
dkhenry: The API in golang is azblob, so thats what I called it in vitess and I feel the name should be the same here. It would be like saying SimpleStorageService instead of S3, if everyone knows it as azblob we should just use that name

pkg/operator/vitessbackup/storage_azblob.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func azblobBackupEnvVars(azblob *planetscalev2.AzblobBackupLocation) []corev1.EnvVar {
+	if azblob.AuthSecret == nil {
+		return nil
+	}
+	return []corev1.EnvVar{
+		{
+			Name: "VT_AZBLOB_ACCOUNT_NAME",
+			ValueFrom: &corev1.EnvVarSource{
```
enisoc: We can move this up to the command-line flag section if you're ok converting it to a regular field instead of a Secret as mentioned above.

pkg/operator/vitessbackup/storage_azblob.go
```diff
@@ -0,0 +1,61 @@
...
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: azblob.AuthSecret.Name,
+					},
+					Key: azblobAccountName,
+				},
+			},
+		}, {
+			Name: "VT_AZBLOB_ACCOUNT_KEY",
+			ValueFrom: &corev1.EnvVarSource{
```
enisoc: I forgot that we can't use an EnvVarSource because it breaks an important feature of SecretSource, which is that you can specify the file source as an arbitrary Volume. EnvVarSource supports pulling from a Secret, but it doesn't support pulling from an arbitrary Volume.

We have to convert this to use the command-line flag that specifies a file path, and mount the secret as a volume like in `storage_gcs.go`.

dkhenry:
This is the relevant backup section I used to test this 
```
  backup:
    locations:
    - azblob:
        account: azurestaginguswest
        container: backup
        authSecret:
          name: azure-dev-vitess
          key: key1
```

If you are interested please contact me and I can get you the storage account key

pkg/apis/planetscale/v2/defaults.go
```diff
@@ -88,7 +88,7 @@ const (
 	// DefaultMysqlPortName is the name for the MySQL port.
 	DefaultMysqlPortName = "mysql"
 
-	defaultVitessLiteImage = "us.gcr.io/planetscale-vitess/lite:2020-01-14.6b55d845"
+	defaultVitessLiteImage = "us.gcr.io/planetscale-vitess/lite:2020-03-07.f569401b"
```
avaidyanatha: Is this needed for this PR or are you just updating it to keep our image up to date? Since this may affect other things, I'd include it in the PR description for posterity's sake.

dkhenry: This is needed for this PR. That is the first vitess version that has the azblob backup engine

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -116,6 +118,27 @@ type S3BackupLocation struct {
 	AuthSecret *SecretSource `json:"authSecret,omitempty"`
 }
 
+// AzblobBackupLocation specifies a backup location in Amazon S3
```
enisoc:
```suggestion
// AzblobBackupLocation specifies a backup location in Azure Blob Storage.
```

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -116,6 +118,27 @@ type S3BackupLocation struct {
...
+type AzblobBackupLocation struct {
+	// Account is the name of the Azure storage account to use
```
enisoc:
```suggestion
	// Account is the name of the Azure storage account to use.
```

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -116,6 +118,27 @@ type S3BackupLocation struct {
...
+	// This is only needed if the same container is also used for something other
+	// than backups for VitessClusters. Backups from different clusters,
+	// keyspaces, or shards will automatically avoid colliding with each other
+	// within a container, regardless of this setting.
+	// +kubebuilder:validation:Pattern=^[^\r\n]*$
+	// +kubebuilder:validation:MaxLength=256
+	KeyPrefix string `json:"keyPrefix,omitempty"`
+	// AuthSecret is a reference to the Secret to use for the Azure authentication
+	// The secret must contain two keys; azblob-account-key, and azblob-account-name
+	AuthSecret *SecretSource `json:"authSecret"`
```
enisoc: In AWS and GCP, we make the AuthSecret optional (`omitempty`) for those who choose to give implicit blob store permissions to the Nodes themselves. Does the Vitess Azure plugin support such a mode, or is this truly required?

If it's truly required, let's make this a non-pointer type. If we decide to make it optional instead, we'll have to add `nil` checks down in `storage_azblob.go`.

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -67,6 +67,8 @@ type VitessBackupLocation struct {
 	GCS *GCSBackupLocation `json:"gcs,omitempty"`
 	// S3 specifies a backup location in Amazon S3.
 	S3 *S3BackupLocation `json:"s3,omitempty"`
+	// Azblob specifies a backup location in Azure Blob Storage.
```
enisoc: Run `make generate` to update generated code and docs.

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
dkhenry: It is in fact required

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
dkhenry: I get this error and I don't know how to convince it to get that dependency
```
operator-sdk-v0.10.0 generate k8s
Error: chdir /home/dk/go/src/planetscale.dev/vitess-operator/vendor/k8s.io/code-generator: no such file or directory
```

dkhenry: and this 
```
> operator-sdk-v0.10.0 generate openapi --verbose
DEBU[0000] Debug logging is set                         
DEBU[0000] Running []string{"go", "build", "-o", "/home/dk/go/src/planetscale.dev/vitess-operator/build/_output/bin/openapi-gen", "./cmd/openapi-gen"} 
can't load package: package ./cmd/openapi-gen: cannot find package "." in:
	/home/dk/go/src/planetscale.dev/vitess-operator/vendor/k8s.io/kube-openapi/cmd/openapi-gen
Error: failed to exec []string{"go", "build", "-o", "/home/dk/go/src/planetscale.dev/vitess-operator/build/_output/bin/openapi-gen", "./cmd/openapi-gen"}: exit status 1
```

enisoc: Are you trying to run those outside the Makefile? If so, you need to `export GO111MODULE=on` yourself.

dkhenry: With the Makefile I don't get the error message. they just Exit 1

PR #49, opened 2020-03-11
Updated all copyrights to 2020

PrismaPhonic:
All of our copyrights (except for the newest package added in the most recent PR) have copyrights set to 2019. This PR updates all copyright text to 2020.


PrismaPhonic:
Removing per offline discussion. Sounds like year is only ever set on the initial file creation date and not updated.

PR #51, opened 2020-03-13
Add Annotations for VitessShardTemplate to VitessShard Object

PrismaPhonic:
This PR adds the ability for users to inject annotations into the VitessShard object via a new annotations field in the VitessShardTemplate. This is useful for passing down metadata from other controllers that vitess-operator can be agnostic to.


pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -128,6 +128,7 @@ func newVitessShard(key client.ObjectKey, vtk *planetscalev2.VitessKeyspace, par
 			Namespace: key.Namespace,
 			Name:      key.Name,
 			Labels:    labels,
+			Annotations: shard.VitessShardTemplate.Annotations,
```
enisoc: nit: use the version from the deep copy since we have one already:

```suggestion
			Annotations: template.Annotations,
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -151,6 +152,27 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
...
+	update.Annotations(&vts.Annotations, newShard.Annotations)
+
 	// For now, everything in Spec is safe to update.
 	vts.Spec = newShard.Spec
 }
+
+// differentKeys returns keys from an older map instance that are no longer in a newer map instance.
+func differentKeys(oldMap, newMap map[string]string) []string {
+	var differentKeys []string
+	for k, _ := range oldMap {
```
enisoc: You can just leave out the value if you don't need it:

```suggestion
	for k := range oldMap {
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -151,6 +152,27 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vts.Labels, newShard.Labels)
 
+	// Remove old annotations that shouldn't be there that we injected previously.
+	differentAnnotations := differentKeys(newShard.Annotations, vts.Spec.Annotations)
```
enisoc: These should be reversed, right? Also, I think we should compare against `newShard.Spec.Annotations` so we don't get confused if later there is an extra annotation in `newShard.Annotations` that is not part of the set that comes from `newShard.Spec.Annotations`.

```suggestion
	differentAnnotations := differentKeys(vts.Spec.Annotations, newShard.Spec.Annotations)
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -151,6 +152,27 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vts.Labels, newShard.Labels)
 
+	// Remove old annotations that shouldn't be there that we injected previously.
```
enisoc: Add another line to document the implicit ordering dependency:

```suggestion
	// Remove old annotations that shouldn't be there that we injected previously.
	// This must be done before we update vts.Spec.
```

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/51/commits/96319b058fcd2e98c87277e0877950b0e885ccf4

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/51/commits/cfce0b22ddf5451c2b02a79f6083963644a2a65a.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Yep, good catch! https://github.com/planetscale/vitess-operator/pull/51/commits/e12e40a8d7d1f5b6127b6a00b1f8bbd4481cb875

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Added explanation comment about implicit ordering: https://github.com/planetscale/vitess-operator/pull/51/commits/e12e40a8d7d1f5b6127b6a00b1f8bbd4481cb875.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -151,6 +152,28 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
...
+	// This must be done before we update vts.Spec.
+	differentAnnotations := differentKeys(vts.Spec.Annotations, newShard.Annotations)
```
enisoc: I think you missed the second part of my 2-part comment:

> Also, I think we should compare against newShard.Spec.Annotations so we don't get confused if later there is an extra annotation in newShard.Annotations that is not part of the set that comes from newShard.Spec.Annotations.

PrismaPhonic: Ahh, I did miss that, and good idea. Will update.

PrismaPhonic: Done https://github.com/planetscale/vitess-operator/pull/51/commits/5b88f8f334dc462e1e9c98efcfbac34097c7adf1.

PR #52, opened 2020-03-17
Share Prune Cells

PrismaPhonic:
This PR moves prune cells logic to a shared (public) library function, so that consumers can re-use cell pruning logic as they see fit.

### DIFF
--- a/pkg/operator/vitesstopo/prune_cells.go
+++ b/pkg/operator/vitesstopo/prune_cells.go
@@ -54,10 +54,10 @@ func PruneCells(ctx context.Context, p PruneCellsParams) (reconcile.Result, erro
 
 	for _, cellName := range candidates {
 		if err := p.TopoServer.DeleteCellInfo(ctx, cellName); err != nil {
-			p.Recorder.Eventf(p.EventObj, corev1.EventTypeWarning, "TopoCleanupFailed", "unable to remove cell %s from topology: %v", name, err)
+			p.Recorder.Eventf(p.EventObj, corev1.EventTypeWarning, "TopoCleanupFailed", "unable to remove cell %s from topology: %v", cellName, err)
 			resultBuilder.RequeueAfter(topoRequeueDelay)
 		} else {
-			p.Recorder.Eventf(p.EventObj, corev1.EventTypeNormal, "TopoCleanup", "removed unwanted cell %s from topology", name)
+			p.Recorder.Eventf(p.EventObj, corev1.EventTypeNormal, "TopoCleanup", "removed unwanted cell %s from topology", cellName)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


PrismaPhonic:
Looks like this PR is now blocked until this gets approved and merged: https://github.com/planetscale/planetscale-operator/pull/3009. The reasoning is that right now psc is not a superset of vt because of new annotations field in VitessShardTemplate, and required logic to copy annotations down from psc -> vt that are solved in the dead master recovery PR, waiting to be merged.

pkg/operator/vitesstopo/prune_cells.go
```diff
@@ -0,0 +1,92 @@
...
+			// We should add it to the list of candidates to prune.
+			candidates = append(candidates, name)
+		}
+	}
+
+	return candidates
+}
+
+// PruneCellCandidates takes in a list of cell candidates (cell names) and prunes them from topology.
+func PruneCellCandidates(ctx context.Context, ts *topo.Server, recorder record.EventRecorder, eventObj runtime.Object, cellCandidates []string) (reconcile.Result, error) {
```
enisoc: They aren't just candidates anymore at the point when we decide to pass them here. Also, "pruning" refers to a process that selectively deletes unneeded parts; this function is indiscriminately deletes everything passed in, so we should make that clear.

```suggestion
// DeleteCells takes in a list of cell names and deletes their CellInfo records from topology.
func DeleteCells(ctx context.Context, ts *topo.Server, recorder record.EventRecorder, eventObj runtime.Object, cells []string) (reconcile.Result, error) {
```

pkg/operator/vitesstopo/prune_cells.go
```diff
@@ -0,0 +1,92 @@
...
+	candidates := CandidatesFromCellsList(cellNames, p.DesiredCells, p.OrphanedCells)
+
+	result, err := PruneCellCandidates(ctx, p.TopoServer, p.Recorder, p.EventObj, candidates)
+	resultBuilder.Merge(result, err)
+
+	return resultBuilder.Result()
+}
+
+// CandidatesFromCellsList returns a list of cell candidates for pruning, based on a provided list of cells to consider.
+func CandidatesFromCellsList(cellNames []string, desiredCells map[string]*planetscalev2.LockserverSpec, orphanedCells map[string]*planetscalev2.OrphanStatus) []string {
```
enisoc: This package is only called `vitesstopo` so there's no context to know that a function called `vitesstopo.CandidatesFromCellsList` is about pruning - candidates for what?

```suggestion
// CellsToPrune returns a list of cell candidates for pruning, based on a provided list of cells to consider.
func CellsToPrune(cellNames []string, desiredCells map[string]*planetscalev2.LockserverSpec, orphanedCells map[string]*planetscalev2.OrphanStatus) []string {
```

PR #56, opened 2020-03-20
Make DeleteCells Idempotent

PrismaPhonic:
Right now DeleteCells is not idempotent because ts.DeleteCellInfo may return a topo.NoNode error type in the case that the cell was not found while attempting to delete it. To ensure DeleteCells is idempotent, we should not treat "Not Found" as an error type, even if Vitess does.

### DIFF
--- a/pkg/operator/vitesstopo/prune_cells.go
+++ b/pkg/operator/vitesstopo/prune_cells.go
@@ -80,7 +80,8 @@ func DeleteCells(ctx context.Context, ts *topo.Server, recorder record.EventReco
 	resultBuilder := &results.Builder{}
 
 	for _, cellName := range cellNames {
-		if err := ts.DeleteCellInfo(ctx, cellName); err != nil {
+		// topo.NoNode is the error type returned if we can't find the cell when deleting. This ensures that this operation is idempotent.
+		if err := ts.DeleteCellInfo(ctx, cellName); err != nil && !topo.IsErrType(err, topo.NoNode) {
 			recorder.Eventf(eventObj, corev1.EventTypeWarning, "TopoCleanupFailed", "unable to remove cell %s from topology: %v", cellName, err)
 			resultBuilder.RequeueAfter(topoRequeueDelay)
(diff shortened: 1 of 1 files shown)


pkg/operator/vitesstopo/prune_cells.go
```diff
@@ -80,7 +80,8 @@ func DeleteCells(ctx context.Context, ts *topo.Server, recorder record.EventReco
 	resultBuilder := &results.Builder{}
 
 	for _, cellName := range cellNames {
-		if err := ts.DeleteCellInfo(ctx, cellName); err != nil {
+		// topo.NoNode is the error type returned if we can't find the cell when deleting. This ensures that this operation is idempotent.
```
enisoc: I think we should avoid logging that we deleted it if we actually did nothing.

PrismaPhonic: Ok, pushed up a new commit to make it so we don't log in this case. I think by definition though that may make this non-idempotent, because don't the same thing multiple times results in different side effects (the first time logging an event, and then not logging every again). I'm not a purist though and may be wrong. The end outcome is idempotent enough I think regardless.

pkg/operator/vitesstopo/prune_cells.go
```diff
@@ -80,10 +80,11 @@ func DeleteCells(ctx context.Context, ts *topo.Server, recorder record.EventReco
...
```
avaidyanatha: Do you think that it would be safer to handle the `NoNode` error distinctly? I feel that it would be cleaner and would read better if this function was modified in the future.

PrismaPhonic: My feeling is that the first conditional arm is our error path - and fundamentally a "not found" error should be a non-error in a delete operation, because delete operations should be idempotent. In this case, I think it makes more sense to combine them together to very clearly illustrate that an error is only an error if it's NOT a "not found" error specifically in relation to delete operations.

PR #58, opened 2020-03-23
etcdls: Add fields to help run etcd across Kubernetes clusters.

enisoc:
These are some necessary but not sufficient changes to support running etcd across Kubernetes clusters with the EtcdLockserver controller.

These new fields allow you to:

- Disable PDB since it can't see Pods in other clusters.
- Disable Services since they need to be managed separately.
- Deploy only one etcd member in each cluster.
- Override peer URLs so they can find each other across clusters.

### DIFF
--- a/pkg/apis/planetscale/v2/etcdlockserver_types.go
+++ b/pkg/apis/planetscale/v2/etcdlockserver_types.go
@@ -126,6 +126,52 @@ type EtcdLockserverTemplate struct {
 	// ExtraLabels can optionally be used to attach custom labels to Pods
 	// created for this component.
 	ExtraLabels map[string]string `json:"extraLabels,omitempty"`
+
+	// CreatePDB sets whether to create a PodDisruptionBudget (PDB) for etcd
+	// member Pods.
+	//
+	// Note: Disabling this will NOT delete a PDB that was previously created.
+	//
+	// Default: true
+	CreatePDB *bool `json:"createPDB,omitempty"`
+
+	// CreateClientService sets whether to create a Service for the client port
+	// of etcd member Pods.
+	//
+	// Note: Disabling this will NOT delete a Service that was previously created.
+	//
+	// Default: true
+	CreateClientService *bool `json:"createClientService,omitempty"`
+
+	// CreatePeerService sets whether to create a Service for the peer port
+	// of etcd member Pods.
+	//
+	// Note: Disabling this will NOT delete a Service that was previously created.
+	//
+	// Default: true
+	CreatePeerService *bool `json:"createPeerService,omitempty"`
+
+	// AdvertisePeerURLs can optionally be used to override the URLs that etcd
+	// members use to find each other for peer-to-peer connections.
+	//
+	// If specified, the list must contain exactly 3 entries, one for each etcd
+	// member index (1,2,3) respectively.
+	//
+	// Default: Build peer URLs automatically based on Kubernetes built-in DNS.
+	// +kubebuilder:validation:MinItems=3
+	// +kubebuilder:validation:MaxItems=3
+	AdvertisePeerURLs []string `json:"advertisePeerURLs,omitempty"`
+
+	// LocalMemberIndex can optionally be used to specify that only one etcd
+	// member should actually be deployed. This can be used to spread members
+	// across multiple Kubernetes clusters by configuring the EtcdLockserver CRD
+	// in each cluster to deploy a different member index. If specified, the
+	// index must be 1, 2, or 3.
+	//
+	// Default: Deploy all etcd members locally.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=3
+	LocalMemberIndex *int32 `json:"localMemberIndex,omitempty"`
 }
 
 // EtcdLockserverStatus defines the observed state of an EtcdLockserver.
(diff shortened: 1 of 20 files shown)


pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -126,6 +126,42 @@ type EtcdLockserverTemplate struct {
 	// ExtraLabels can optionally be used to attach custom labels to Pods
 	// created for this component.
 	ExtraLabels map[string]string `json:"extraLabels,omitempty"`
+
+	// CreatePDB sets whether to create a PDB for etcd member Pods.
```
teejae: spell out what PDB means
```suggestion
	// CreatePDB sets whether to create a PodDisruptionBudget (PDB) for etcd member Pods.
```

pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -126,6 +126,42 @@ type EtcdLockserverTemplate struct {
...
+	// CreatePeerService sets whether to create a Service for the peer port
+	// of etcd member Pods.
+	// Note: Disabling this will NOT delete a Service that was previously created.
+	// Default: true
+	CreatePeerService *bool `json:"createPeerService,omitempty"`
+
+	// AdvertisePeerURLs can optionally be used to override the URLs that etcd
+	// members use to find each other for peer-to-peer connections.
+	// If specified, the list must contain exactly 3 entries, one for each etcd
+	// member index (1,2,3) respectively.
```
teejae: do we document that there must be exactly 3 instances somewhere?

pkg/controller/etcdlockserver/reconcile_members.go
```diff
@@ -104,8 +104,15 @@ func (r *ReconcileEtcdLockserver) reconcileMembers(ctx context.Context, ls *plan
 	}
 
 	// Update status of the lockserver.
-	// We should be available for queries if the number of Ready replicas is a majority.
-	ls.Status.Available = k8s.ConditionStatus(numPodsReady >= etcd.QuorumSize)
+	if ls.Spec.LocalMemberIndex == nil {
+		// We're deploying all members locally, so we can see all of them.
+		// We should be available for queries if the number of Ready replicas is a majority.
```
teejae:
```suggestion
		// We should be available for queries if the number of Ready replicas has reached quorum.
```

pkg/controller/etcdlockserver/reconcile_members.go
```diff
@@ -114,6 +121,11 @@ func (r *ReconcileEtcdLockserver) reconcileMembers(ctx context.Context, ls *plan
 func memberSpecs(ls *planetscalev2.EtcdLockserver, parentLabels map[string]string) []*etcd.Spec {
 	members := make([]*etcd.Spec, 0, etcd.NumReplicas)
 	for i := 1; i <= etcd.NumReplicas; i++ {
+		// If we're only deploying one member locally, skip the others.
+		if ls.Spec.LocalMemberIndex != nil && int32(i) != *ls.Spec.LocalMemberIndex {
```
teejae:
```suggestion
		if ls.Spec.LocalMemberIndex != nil && (*ls.Spec.LocalMemberIndex != int32(i)) {
```

pkg/apis/planetscale/v2/etcdlockserver_types.go
enisoc: I try to mention it in every place it's relevant (like here). If you find a place where it's missing, we can add it.

pkg/apis/planetscale/v2/etcdlockserver_types.go
enisoc: Done.

pkg/controller/etcdlockserver/reconcile_members.go
enisoc: Done.

pkg/controller/etcdlockserver/reconcile_members.go
enisoc: As a matter of consistent style, I always try to write conditional expressions with the dynamic variable first, and the "constant" search value second. In this case, `i` is the value that's changing each time through the loop. `LocalMemberIndex` is a fixed value that we're searching for.

pkg/operator/etcd/pod.go
```diff
@@ -311,17 +312,31 @@ func (spec *Spec) Args() []string {
...
+			advertisePeerURLs = append(advertisePeerURLs, fmt.Sprintf("http://%s.%s:%d", peerName, subdomain, PeerPortNumber))
+		}
+	}
+
+	// Set the address that this peer will advertise for itself.
+	initialAdvertisePeerURLs := advertisePeerURLs[spec.Index-1]
+
+	// Create list of peer addresses.
 	initialCluster := make([]string, 0, NumReplicas)
 	for i := 1; i <= NumReplicas; i++ {
```
PrismaPhonic: If possible with manual loops, I think it's better to index from 0. In this case we use i twice, and the second time we index at `i-1`. I personally would prefer PodName be with i+1 (which is clearer anyways, since it's not *really* the index), and index into advertisePeerURLs with just `i`.

enisoc: Done.

PR #59, opened 2020-03-24
Allow disabling of InitReplication through VitessCluster CRD

avaidyanatha:
## Main Changes
- We want to be able to disable the initialization of replication so that it can be manually configured for cases in which multiple Vitess operators are racing to initialize replication.


pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -86,6 +86,9 @@ type VitessShardSpec struct {
 
 	// TopologyReconciliation is inherited from the parent's VitessClusterSpec.
 	TopologyReconciliation *TopoReconcileConfig `json:"topologyReconciliation,omitempty"`
+
+	// InitReplication is inherited from the parent's VitessClusterSpec.
```
PrismaPhonic: If it's inherited from the parent's VitessClusterSpec then we should not be running defaulting code in `DefaultVitessShard()`. Defaulting should be up to the `DefaultVitessCluster()` and the value should simply be copied down.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -193,22 +193,25 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
-	// If we are not using external MySQL we will bail out early.
-	terResult, err := r.tabletExternallyReparent(ctx, vts, wr)
-	resultBuilder.Merge(terResult, err)
-
-	// Check if we need to start replication on a shard that's been restored
-	// from backup. If it's already initialized, this will be a no-op.
-	irsResult, err := r.initRestoredShard(ctx, vts, wr)
-	resultBuilder.Merge(irsResult, err)
+	// If we specified the operator to handle the initialization of replication, make the appropriate checks.
+	if *vts.Spec.InitReplication {
```
PrismaPhonic: At first glance, I wanted to suggest using a bailout rather than putting everything into this conditional, but I see that this only applies to a select few items. I feel like something that might help here is to make a private method called `r.initReplication(ctx, vts, wr)` where these three functions are simply run inside. Then you can simply put the bailout inside of `r.initReplication()` since you are passing `vts` into it anyways.

pkg/apis/planetscale/v2/vitessshard_types.go
avaidyanatha: Done - Just updated it to copy the value down. And makes sense on the Defaulting code side, we are just setting one value and passing it down.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: I put it into its own method, which makes sense... but I'm unsure what you mean about bailing within the method, can you elaborate?

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -193,22 +193,10 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
+		r.initReplication(ctx, vts, wr, resultBuilder)
```
PrismaPhonic: See my comments below. Once you are returning a `(reconcile.Result, error)`, you should merge here and take away the conditional entirely:
```suggestion
		initReplicationResult, err := r.initReplication(ctx, vts, wr, resultBuilder)
		resultBuilder.Merge(initReplicationResult, err)
```

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Cool, I'm happy with that part so resolving this conversation so it doesn't clutter up space.

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -110,6 +110,9 @@ type VitessClusterSpec struct {
 
 	// TopologyReconciliation can be used to enable or disable registration or pruning of various vitess components to and from topo records.
 	TopologyReconciliation *TopoReconcileConfig `json:"topologyReconciliation,omitempty"`
+
+	// InitReplication can be used to enable or disable the operator initializing replication on a new or restored Vitess Shard object.
+	InitReplication *bool `json:"initReplication,omitempty"`
```
enisoc: I think we should put this in the existing `VitessReplicationSpec` struct since it's another replication-related setting. Since the parent field will already be called `replication`, we won't need to repeat that in the name of the child field. Perhaps we could call it `InitializeMaster` since this is about whether we choose the initial master for a shard that has no master?

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +214,22 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
 	reconcileCount.WithLabelValues(metricLabels(vts, err)...).Inc()
 	return result, err
 }
+
+func (r *ReconcileVitessShard) initReplication(ctx context.Context, vts *planetscalev2.VitessShard, wr *wrangler.Wrangler, rb *results.Builder) {
```
enisoc: We don't usually pass resultBuilders around. Make a new one inside this function and return `resultBuilder.Result()`. It's up to the caller if they want to merge that with their own resultBuilder.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
enisoc: He means move the `if *vts.Spec.InitReplication {` check inside of `initReplication()` instead of doing it here. IOW we always call initReplication but it might return immediately if it decides it shouldn't run. This is a convention we have to keep logic out of the main `Reconcile()` function in controllers where we break things out into subroutines. We let the subroutines themselves decide whether they need to do anything.

go.mod
```diff
@@ -16,17 +16,19 @@ require (
...
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
+	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9 // indirect
 	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.0.0-20190612125737-db0771252981
 	k8s.io/apimachinery v0.0.0-20190612125636-6a5db36e93ad
 	k8s.io/apiserver v0.0.0-20190228174905-79427f02047f // indirect
 	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/code-generator v0.17.3 // indirect
+	k8s.io/gengo v0.0.0-20190822140433-26a664648505 // indirect
```
enisoc: This PR shouldn't need to change deps, right? Can you try reverting `go.*` and just run `make build` without any manual `go mod` commands?

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: Since we're adding a field that's default-on, I think we do need to run defaulting even for inherited fields. Come to think of it, I might have messed this up too in my last PR. Let me go take another look.

enisoc: Oh looks like I did remember to add this in my other PR:

https://github.com/planetscale/vitess-operator/pull/58/files#diff-8d96129ffdb514abe933c2a0416b261bR150

We'll need to do something similar here, to make sure this new vitessshardreplication controller code will see an existing VitessShard object (made by the old vitesscluster controller) as having these default-on fields on.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +214,22 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
+	// Check if we need to initialize the shard.
```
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/apis/planetscale/v2/vitessshard_types.go
avaidyanatha: Done, added a line to materialize defaults and put the default function back in vitessshard_defaults.go

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: Ah okay, that convention makes sense, and so does the explanation now. Thanks!

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +214,22 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
+	// If it's already initialized, this will be a no-op.
+	// If we are using external MySQL we will bail out early.
+	ismResult, err := r.initShardMaster(ctx, vts, wr)
+	rb.Merge(ismResult, err)
```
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +214,22 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
+	// Check if we need to externally reparent
+	// in the case of external MySQL.
+	// If we are not using external MySQL we will bail out early.
+	terResult, err := r.tabletExternallyReparent(ctx, vts, wr)
+	rb.Merge(terResult, err)
+
+	// Check if we need to start replication on a shard that's been restored
+	// from backup. If it's already initialized, this will be a no-op.
+	irsResult, err := r.initRestoredShard(ctx, vts, wr)
+	rb.Merge(irsResult, err)
```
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: Done.

go.mod
avaidyanatha: Build runs fine with reverting the go.sum and go.mod files.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: I agree, although the definition of `VitessReplicationSpec` makes it sound like those options only modify Vitess/MySQL level parameters, whereas our new parameter is an operator action.

avaidyanatha: I can put it in there and slightly modify the description?

pkg/apis/planetscale/v2/vitessshard_defaults.go
```diff
@@ -16,10 +16,24 @@ limitations under the License.
...
 // DefaultVitessShard fills in VitessShard defaults for unspecified fields.
 // Note: This should only be used for nillable fields passed down from a parent because controllers run in parallel,
 // and the defaulting code for a parent object may not have been run yet, meaning the values passed down from that parent
 // might not be safe to deref.
 func DefaultVitessShard(dst *VitessShard) {
 	DefaultTopoReconcileConfig(&dst.Spec.TopologyReconciliation)
+	defaultInitializeMaster(&dst.Spec.VitessShardTemplate.Replication.InitializeMaster)
 }
+
+func defaultInitializeMaster(enabled *bool) {
```
PrismaPhonic: I'm a little confused. Wouldn't this be a `**bool` at this point because `InitializeMaster` is a pointer bool already, so we need to pass the address to that pointer bool?

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -134,6 +134,10 @@ type VitessReplicationSpec struct {
 	//
 	// Default: Semi-sync is not enforced.
 	EnforceSemiSync bool `json:"enforceSemiSync,omitempty"`
+
+	// InitializeMaster can be used to enable or disable the operator initializing replication
+	// on a new or restored Vitess Shard object.
+	InitializeMaster bool `json:"initializeMaster,omitempty"`
```
PrismaPhonic: I must have missed something - Why is this a different name from `initReplication` which still exists on `VitessKeyspaceSpec`? Did you mean to update all of them to `InitializeMaster`? Also, why is this one no longer a pointer type?

avaidyanatha: `initReplication` still existing on `VitessKeyspaceSpec` was just a mistake. I just cleaned that up.

PrismaPhonic: Cool, will read through one more time.

pkg/apis/planetscale/v2/vitessshard_defaults.go
avaidyanatha: The only time it will be called, it will be defaulted, so we just call the address of that defaulted template.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: I believe you still will want to copy the setting on down with the following logic:
1. If child has setting (non-nil), use child setting, 
2. else copy parent down.

this allows someone to specify the setting at the cluster level and have overrides at the shard level. In other words, I could turn it off globally, and then turn it on for select keyspaces, or select shards.

pkg/apis/planetscale/v2/vitessshard_defaults.go
PrismaPhonic: I see my confusion - you changed it to a normal bool. I think it should be a pointer bool so we can change the default over time. This also has been our general pattern for code that we default (making them pointer types).

pkg/apis/planetscale/v2/vitessshard_defaults.go
```diff
@@ -16,10 +16,24 @@ limitations under the License.
...
+	// Enable initialization of replication by default.
+	enableInitMaster := enabled
+
+	if enableInitMaster == nil {
```
PrismaPhonic: Also, now that you made the type non-pointer this will never be nil. You are always taking a valid address when you call this defaulting function, so now it doesn't ever default.

pkg/apis/planetscale/v2/vitessshard_defaults.go
enisoc: Peter is right; this would have to be a `**bool`.

However, I recommend changing the function to be `DefaultVitessReplicationSpec(*VitessReplicationSpec)` because we don't usually create a whole defaulter function for just one bool.

We should also then call `DefaultVitessReplicationSpec()` on every `VitessShardTemplate` nested within a `VitessCluster`. This is the first time we've had a non-zero default in that struct, so you'll need to create a few of the intermediate methods. We currently only go as deep as `DefaultVitessKeyspaceTemplate()`.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -134,6 +134,10 @@ type VitessReplicationSpec struct {
 	//
 	// Default: Semi-sync is not enforced.
 	EnforceSemiSync bool `json:"enforceSemiSync,omitempty"`
+
+	// InitializeMaster can be used to enable or disable the operator initializing replication
+	// on a new or restored Vitess Shard object.
```
enisoc:
```suggestion
	// InitializeMaster specifies whether to choose an initial master for a
	// new or restored shard that has no master yet.
	//
	// Default: true.
```

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: This has to be a pointer so we can tell the difference between true, false, and unset.

```suggestion
	InitializeMaster *bool `json:"initializeMaster,omitempty"`
```

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +216,29 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
 	reconcileCount.WithLabelValues(metricLabels(vts, err)...).Inc()
 	return result, err
 }
+
+func (r *ReconcileVitessShard) initReplication(ctx context.Context, vts *planetscalev2.VitessShard, wr *wrangler.Wrangler) (reconcile.Result, error){
+	resultBuilder := &results.Builder{}
+
+	// If we have configured the operator to not initialize replication, bail.
+	if !vts.Spec.VitessShardTemplate.Replication.InitializeMaster {
```
enisoc: VitessShardTemplate is an anonymous field so we usually omit it for clarity.

```suggestion
	if !*vts.Spec.Replication.InitializeMaster {
```

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
```diff
@@ -226,3 +216,29 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
...
+	// Check if we need to externally reparent
+	// in the case of external MySQL.
+	// If we are not using external MySQL we will bail out early.
+	terResult, err := r.tabletExternallyReparent(ctx, vts, wr)
+	resultBuilder.Merge(terResult, err)
+
+	// Check if we need to start replication on a shard that's been restored
+	// from backup. If it's already initialized, this will be a no-op.
+	irsResult, err := r.initRestoredShard(ctx, vts, wr)
+	return resultBuilder.Merge(irsResult, err)
```
enisoc: Just call Merge without returning, then have a separate return at the end. This maintains symmetry of all the parts of the function, which makes it easy to add/remove/rearrange in the future.

pkg/apis/planetscale/v2/vitessshard_types.go
enisoc: We don't have a cluster-level Replication field in VitessCluster yet. For now, you have to specify this individually for every shard template.

It would make sense to add a cluster-level field in the future, but the override logic will live in the controller that generates VitessShard objects, not in defaulting code. Defaulting code needs to avoid taking values from one place and sticking them somewhere else; we learned from mistakes in core k8s that that breaks compatibility.

pkg/apis/planetscale/v2/vitessshard_defaults.go
avaidyanatha: Fixed.

pkg/apis/planetscale/v2/vitessshard_types.go
avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: Done.

avaidyanatha: Done.

pkg/controller/vitessshardreplication/vitessshardreplication_controller.go
avaidyanatha: Sounds good, done.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: I put the comment here because I wasn't talking about defaulting code.  Sorry if my comment was unclear.

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -99,6 +99,31 @@ func DefaultVitessKeyspaceTemplate(keyspace *VitessKeyspaceTemplate) {
...
+	}
+}
+
+func defaultCustomPartitioning(customPartition *VitessKeyspaceCustomPartitioning) {
+	if customPartition == nil {
+		return
+	}
+
+	for i := range customPartition.Shards {
+		shard := &customPartition.Shards[i]
```
PrismaPhonic: Super small nit, but I'm not sure this line buys us much in code clarity. Up to you but you can probably just one line this:
```suggestion
		DefaultVitessShardTemplate(&customPartition.Shards[i].VitessShardTemplate)
```

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -134,6 +134,12 @@ type VitessReplicationSpec struct {
 	//
 	// Default: Semi-sync is not enforced.
 	EnforceSemiSync bool `json:"enforceSemiSync,omitempty"`
+
+	// InitializeMaster specifies whether to choose an initial master for a
+	// new or restored shard that has no master yet.
```
PrismaPhonic: Because the VitessReplicationSpec is used in shard, keyspace and cluster specs, I would advise maybe changing the comment a bit so it makes sense no matter where it gets used.
```suggestion
	// InitializeMaster specifies whether to elect an initial master for
	// new or restored shards that have no master yet.
```

pkg/apis/planetscale/v2/vitesscluster_defaults.go
```diff
@@ -99,6 +99,31 @@ func DefaultVitessKeyspaceTemplate(keyspace *VitessKeyspaceTemplate) {
...
+}
+
+func defaultCustomPartitioning(customPartition *VitessKeyspaceCustomPartitioning) {
+	if customPartition == nil {
+		return
+	}
+
+	for i := range customPartition.Shards {
+		DefaultVitessShardTemplate(&customPartition.Shards[i].VitessShardTemplate)
+		DefaultVitessShardTemplate(&shard.VitessShardTemplate)
```
enisoc: Extra?

avaidyanatha: Fixing that as we speak :)

PR #61, opened 2020-03-25
Allow disabling of the VitessShard initial backup job

avaidyanatha:
## Main Changes
- Allow disabling of the initial backup job to allow for the manual configuration of multiple actors racing to initialize it.

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -141,7 +141,9 @@ type VitessReplicationSpec struct {
 	// Default: true.
 	InitializeMaster *bool `json:"initializeMaster,omitempty"`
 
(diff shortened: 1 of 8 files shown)


pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -140,6 +140,11 @@ type VitessReplicationSpec struct {
 	//
 	// Default: true.
 	InitializeMaster *bool `json:"initializeMaster,omitempty"`
+
+	// InitializeBackup specifies whether to take an initial placeholder backup before a real one is created.
```
enisoc: Let's add some more context for why this is related to replication:

```suggestion
	// InitializeBackup specifies whether to take an initial placeholder backup
	// as part of preparing tablets to begin replication. This only takes effect
	// if a backup location is defined in the VitessCluster.
```

PR #62, opened 2020-03-25
Make Prune Keyspaces Re-usable

PrismaPhonic:
This PR moves prune keyspace logic into shared public functions that can be re-used by consumers.

### DIFF
--- a/pkg/operator/vitesstopo/prune_keyspaces.go
+++ b/pkg/operator/vitesstopo/prune_keyspaces.go
@@ -1,0 +1,92 @@
+package vitesstopo
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"vitess.io/vitess/go/vt/logutil"
+	"vitess.io/vitess/go/vt/topo"
+	"vitess.io/vitess/go/vt/wrangler"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+)
+
+type PruneKeyspacesParams struct {
+	// EventObj holds the object type that the recorder will use when writing events.
+	EventObj   runtime.Object
+	TopoServer *topo.Server
+	Recorder   record.EventRecorder
+	// Keyspaces is a current list of KeySpaceTemplates from the cluster spec.
+	Keyspaces []planetscalev2.VitessKeyspaceTemplate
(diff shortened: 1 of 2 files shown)


pkg/operator/vitesstopo/prune_keyspaces.go
```diff
@@ -0,0 +1,92 @@
...
+
+// PruneKeyspaces will prune keyspaces that exist but shouldn't anymore.
+func PruneKeyspaces(ctx context.Context, p PruneKeyspacesParams) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
+	// Make a map from keyspace name (as Vitess calls them) back to the keyspace spec.
+	desiredKeyspaces := make(map[string]*planetscalev2.VitessKeyspaceTemplate, len(p.Keyspaces))
+	for i := range p.Keyspaces {
+		keyspace := &p.Keyspaces[i]
+		desiredKeyspaces[keyspace.Name] = keyspace
```
avaidyanatha:
```suggestion
		desiredKeyspaces[keyspace.Name] = &p.Keyspaces[i]
```

avaidyanatha: Nit: Combine these two lines?

pkg/operator/vitesstopo/prune_keyspaces.go
```diff
@@ -0,0 +1,92 @@
...
+	return resultBuilder.Result()
+}
+
+// KeyspacesToPrune returns a list of keyspace candidates for pruning, based on a provided list of keyspaces to consider.
+func KeyspacesToPrune(keyspaceNames []string, desiredKeyspaces map[string]*planetscalev2.VitessKeyspaceTemplate, orphanedKeyspaces map[string]*planetscalev2.OrphanStatus) []string {
+	var candidates []string
+
+	for _, name := range keyspaceNames {
+		if desiredKeyspaces[name] == nil && orphanedKeyspaces[name] == nil {
+			// The cell exists in topo, but not in the VT spec.
```
avaidyanatha: Do you mean keyspace instead of cell?

pkg/operator/vitesstopo/prune_keyspaces.go
PrismaPhonic: good idea - I had simply moved the code but I agree that it's an unnecessary extra line.

pkg/operator/vitesstopo/prune_keyspaces.go
PrismaPhonic: Yep, good catch. Fixed.

pkg/operator/vitesstopo/prune_keyspaces.go
PrismaPhonic: Done.

PR #63, opened 2020-03-26
Make Prune Shards Public

PrismaPhonic:
This PR makes logic within prune shards public, so it can be re-used by external consumers.

### DIFF
--- a/pkg/controller/vitesskeyspace/reconcile_topo.go
+++ b/pkg/controller/vitesskeyspace/reconcile_topo.go
@@ -21,14 +21,13 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	"vitess.io/vitess/go/vt/logutil"
-	"vitess.io/vitess/go/vt/wrangler"
 
 	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
 	"planetscale.dev/vitess-operator/pkg/operator/results"
 	"planetscale.dev/vitess-operator/pkg/operator/toposerver"
+	"planetscale.dev/vitess-operator/pkg/operator/vitesstopo"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


PrismaPhonic:
Should have rebased off master first. Just rebased right now - the extra commits are left over in github land unfortunately.

pkg/controller/vitesskeyspace/reconcile_topo.go
```diff
@@ -56,38 +55,21 @@ func (r *ReconcileVitessKeyspace) reconcileTopology(ctx context.Context, vtk *pl
...
+		for k := range vtk.Status.Shards {
+			desiredShards.Insert(k)
+		}
 
-	// Get list of shards in topo.
-	shardNames, err := ts.GetShardNames(ctx, vtk.Spec.Name)
-	if err != nil {
-		r.recorder.Eventf(vtk, corev1.EventTypeWarning, "TopoListFailed", "failed to list shards in topology: %v", err)
-		return resultBuilder.RequeueAfter(topoRequeueDelay)
+		result, err := vitesstopo.PruneShards(ctx, vitesstopo.PruneShardsParams{
```
avaidyanatha: I know I didn't really contest this on the `PruneKeyspace` PR but what was ultimately the readability decision to have `PruneShards` take in its own custom type? Wouldn't it be cleaner in the main reconcile path just to pass in ctx, vtk, etc.?

pkg/operator/vitesstopo/prune_shards.go
```diff
@@ -0,0 +1,100 @@
...
+			// It's also not being kept around by a blocked turn-down.
+			candidates = append(candidates, name)
+		}
+	}
+
+	return candidates
+}
+
+// DeleteShards takes in a list of shard names and deletes their records from topology.
+func DeleteShards(ctx context.Context, ts *topo.Server, recorder record.EventRecorder, eventObj runtime.Object, keyspaceName string, shardNames []string) (reconcile.Result, error) {
```
avaidyanatha: Nit: Also make this private if it isn't called elsewhere and you agree with the above one.

pkg/controller/vitesskeyspace/reconcile_topo.go
```diff
@@ -56,38 +55,21 @@ func (r *ReconcileVitessKeyspace) reconcileTopology(ctx context.Context, vtk *pl
...
-		result, err := r.pruneShards(ctx, vtk, ts)
-		resultBuilder.Merge(result, err)
-	}
-
-	return resultBuilder.Result()
-}
-
-func (r *ReconcileVitessKeyspace) pruneShards(ctx context.Context, vtk *planetscalev2.VitessKeyspace, ts *toposerver.Conn) (reconcile.Result, error) {
-	resultBuilder := &results.Builder{}
+		desiredShards := make(sets.String, len(vtk.Status.Shards))
```
avaidyanatha: If you agree with the below comment, then this `desiredShards` logic could be entirely isolated inside the prune_shards.go file.

pkg/operator/vitesstopo/prune_shards.go
```diff
@@ -0,0 +1,100 @@
...
+	candidates := ShardsToPrune(shardNames, p.DesiredShards, p.OrphanedShards)
+
+	result, err := DeleteShards(ctx, p.TopoServer, p.Recorder, p.EventObj, p.KeyspaceName, candidates)
+	resultBuilder.Merge(result, err)
+
+	return resultBuilder.Result()
+}
+
+// ShardsToPrune returns a list of shard candidates for pruning, based on a provided list of shards to consider.
+func ShardsToPrune(currentShards []string, desiredShards sets.String, orphanedShards map[string]*planetscalev2.OrphanStatus) []string {
```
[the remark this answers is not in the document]
PrismaPhonic: This PR is intended to make this logic shared for external consumers to use.

pkg/operator/vitesstopo/prune_shards.go
PrismaPhonic: This PR is intended to make this logic shared for external consumers to use.

pkg/controller/vitesskeyspace/reconcile_topo.go
PrismaPhonic: Per our offline discussion, I'll resolve this for now, with the idea that we table this discussion for later :smile_cat:

pkg/controller/vitesskeyspace/reconcile_topo.go
PrismaPhonic: Desired shards might be calculated in different ways by consumers of `ShardsToPrune`, and this PR hopes to make `ShardsToPrune` and `DeleteShards` as easy to use by external consumers as possible.

PR #66, opened 2020-03-29
Added support for injection of custom sidecar containers.

PrismaPhonic:
This adds the feature of sidecar container injection anywhere we have ExtraVolumes and updates SidecarContainers in relevant pod specs.

### DIFF
--- a/pkg/apis/planetscale/v2/etcdlockserver_types.go
+++ b/pkg/apis/planetscale/v2/etcdlockserver_types.go
@@ -113,6 +113,9 @@ type EtcdLockserverTemplate struct {
 	// that will be run to completion one after another before any app containers are started.
 	InitContainers []corev1.Container `json:"initContainers,omitempty"`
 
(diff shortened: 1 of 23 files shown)


pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -113,6 +113,9 @@ type EtcdLockserverTemplate struct {
 	// that will be run to completion one after another before any app containers are started.
 	InitContainers []corev1.Container `json:"initContainers,omitempty"`
 
+	// SidecarContainers can optionally be used to supply extra sidecar containers
```
enisoc:
```suggestion
	// SidecarContainers can optionally be used to supply extra containers that run alongside the main containers.
```

PrismaPhonic: Cool, will update everywhere and then merge.

PR #69, opened 2020-03-31
Add configurable reconcile timeout for all top level Reconcile functions

avaidyanatha:
## Main Changes
- For safety, we should always set time-outs on controller reconciles.
- For start, they will default to 10 minutes long, but can be changed through a command line flag.

## Notes
- This PR adds an environment variables package for the operator to allow for setting global environment variables.

### DIFF
--- a/pkg/operator/environment/environment.go
+++ b/pkg/operator/environment/environment.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 PlanetScale Inc.
+Copyright 2020 PlanetScale Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


pkg/operator/environment/environment.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2019 PlanetScale Inc.
```
PrismaPhonic, line 2:
```suggestion
Copyright 2020 PlanetScale Inc.
```

avaidyanatha: Funny catch, nice eye - done.

PR #70, opened 2020-04-01
Tolerate config dir missing from Vitess image.

enisoc:
Newer Vitess images no longer contain the config dir because the
necessary files are baked into the binaries instead. To support both
old and new images, we now copy the dir if it exists, but we don't
require it to exist.

The RBR setting has also become default in Vitess without needing an
override, but to maintain compatibility with older images we continue to
override it by writing out our own rbr.cnf.


morgo:
LGTM

PR #71, opened 2020-04-01
BugFix: Ensure updateDrainStatus Only Updates on Diff

PrismaPhonic:
We were seeing an issue where pods were being needlessly updated and discovered that it was due to `updateDrainStatus()` calling a pod update everytime regardless of whether status had actually updated or not. This PR seeks to fix this problem by ensuring that we only call for a pod update if status has actually updated.


PrismaPhonic:
@avaidyanatha potentially, although by doing it this way we encapsulate all the logic necessary to ensure that all callers will not call pod updates needlessly.

avaidyanatha:
> @avaidyanatha potentially, although by doing it this way we encapsulate all the logic necessary to ensure that all callers will not call pod updates needlessly.

We should make all of our `client.Update` calls in a consistent way so that we don't run into this next time. Just a thought - I can follow up later on this.

PrismaPhonic:
Cool, sounds good. Merging for now.

PR #75, opened 2020-04-10
Add option to configure storage endpoint for AWS

dkhenry:
Allow the User to configure the endpoint for backups to S3 compatible systems. 

ref: https://developers.digitalocean.com/documentation/spaces/

### DIFF
--- a/pkg/apis/planetscale/v2/vitessbackupstorage_types.go
+++ b/pkg/apis/planetscale/v2/vitessbackupstorage_types.go
@@ -103,8 +103,8 @@ type S3BackupLocation struct {
 	// Bucket is the name of the S3 bucket to use.
 	// +kubebuilder:validation:MinLength=1
 	Bucket string `json:"bucket"`
-	// Endpoint is the url endpoint for the S3 bucket.
-	// Default: User the endpoint associated with `Region` by the driver.
+	// Endpoint is the `host:port` (port is required) for the S3 backend.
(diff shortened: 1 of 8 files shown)


pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -103,6 +103,9 @@ type S3BackupLocation struct {
 	// Bucket is the name of the S3 bucket to use.
 	// +kubebuilder:validation:MinLength=1
 	Bucket string `json:"bucket"`
+	// Endpoint is the url endpoint for the S3 bucket.
+	// Default: User the endpoint associated with `Region` by the driver.
```
enisoc, line 107: We should refer to the JSON field name since these comments go into user-facing docs for the JSON API.

```suggestion
	// Default: Use the endpoint associated with `region` by the driver.
```

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -103,6 +103,9 @@ type S3BackupLocation struct {
 	// Bucket is the name of the S3 bucket to use.
 	// +kubebuilder:validation:MinLength=1
 	Bucket string `json:"bucket"`
+	// Endpoint is the url endpoint for the S3 bucket.
```
enisoc, line 106:
```suggestion
	// Endpoint is the `host:port` (port is required) for the S3 backend.
```

PR #76, opened 2020-04-10
Update VitessShard controller to perform rolling restarts when schedl…

avaidyanatha:
## Main Changes
- The `VitessShard` controller now will attempt to perform a rolling restart if it is given an annotation to perform rolling restarts
- Nothing actually applies the annotations yet in our operator, but can be manually applied to the shard object with `kubectl edit`
- This is a prerequisite for resizing resources

### DIFF
--- a/pkg/controller/vitessshard/reconcile_rollout.go
+++ b/pkg/controller/vitessshard/reconcile_rollout.go
@@ -80,11 +80,11 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 	if tabletKey == "" {
 		// If we have no more scheduled tablets, uncascade the shard.
 		if err := r.uncascadeShard(ctx, vts); err != nil {
-			r.recorder.Eventf(vts, v1.EventTypeWarning, "UncascadeFailed", "Client failed to mark shard as uncascading.")
+			r.recorder.Eventf(vts, v1.EventTypeWarning, "UncascadeFailed", "Failed to mark cascading shard rollout as complete: %v", err)
 			return resultBuilder.Error(err)
 		}
 
-		r.recorder.Eventf(vts, v1.EventTypeNormal, "CascadeComplete", "Cascading changes have been successfully applied to shard.")
+		r.recorder.Eventf(vts, v1.EventTypeNormal, "RollingRestartComplete", "Cascading rollout of tablets is complete.")
 		return resultBuilder.Result()
 	}
 
@@ -99,7 +99,7 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 
 	if err := r.releaseTabletPod(ctx, pod, deletePod); err != nil {
 		r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartBlocked", "release of Pod %v (tablet %v) failed: %v", pod.Name, tabletKey, err)
-		return resultBuilder.Error(err)
+		resultBuilder.Error(err)
 	}
 
 	return resultBuilder.Result()
@@ -122,16 +122,14 @@ func (r *ReconcileVitessShard) uncascadeShard(ctx context.Context, vts *planetsc
 }
 
 func getNextScheduledTablet(tabletKeys []string, tabletPods map[string]*v1.Pod) (string, *v1.Pod){
-	nextTabletKey := ""
 	for _, tabletKey := range tabletKeys {
 		pod := tabletPods[tabletKey]
 		if !rollout.Scheduled(pod) {
 			continue
 		}
 
-		nextTabletKey = tabletKey
-		return nextTabletKey, pod
+		return tabletKey, pod
 	}
 
-	return nextTabletKey, nil
+	return "", nil
 }
--- a/pkg/controller/vitessshard/reconcile_rollout.go
+++ b/pkg/controller/vitessshard/reconcile_rollout.go
@@ -35,7 +35,7 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 		}).AsSelector(),
 	}
 
-	if err := r.client.List(ctx, podList, listOpts); err != nil {
+	if err := r.client.List(ctx, listOpts, podList); err != nil {
 		return resultBuilder.Error(err)
 	}
 
--- a/pkg/controller/vitessshard/reconcile_rollout.go
+++ b/pkg/controller/vitessshard/reconcile_rollout.go
@@ -19,7 +19,7 @@ import (
 func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
 	resultBuilder := &results.Builder{}
 
-	if !rollout.Cascaded(vts) {
+	if !rollout.Cascading(vts) {
 		// If the shard is not scheduled for a cascading update, silently bail out and do nothing.
 		return resultBuilder.Result()
 	}
@@ -48,8 +48,8 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 		tabletPods[tabletKey] = pod
 	}
 
-	tabletKeys := make([]string, 0, len(tabletPods))
-	for key := range tabletPods {
+	tabletKeys := make([]string, 0, len(vts.Status.Tablets))
+	for key := range vts.Status.Tablets {
 		tabletKeys = append(tabletKeys, key)
 	}
 	sort.Strings(tabletKeys)
@@ -57,49 +57,49 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 	for _, tabletKey := range tabletKeys {
 		tablet := vts.Status.Tablets[tabletKey]
 		if tablet.Available != v1.ConditionTrue {
-			// If all tablets aren't healthy, we should bail and not perform a rolling restart.
+			// If any tablets are unhealthy, we should bail and not perform a rolling restart.
 			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to be Available.", tabletKey)
 			return resultBuilder.Result()
 		}
 
-		pod := tabletPods[tabletKey]
-		if rollout.Released(pod) {
-			// If any tablet has already been released, we should wait until it is finished to release another one.
-			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to finish release.", tabletKey)
-		}
-	}
-
-	scheduledTablets := false
-	masterEligibleTablets := vts.Spec.MasterEligibleTabletCount()
-
-	for _, tabletKey := range tabletKeys {
-		pod := tabletPods[tabletKey]
-		if !rollout.Scheduled(pod) {
-			continue
-		}
-
-		scheduledTablets = true
-		deletePod := false
-		tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
-		// These two conditions guarantee that the tablet is a lone master.
-		if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
-			// If we have a lone master, we must delete it since reparenting is impossible.
-			deletePod = true
-		}
-
-		if err := r.releaseTabletPod(ctx, pod, deletePod); err != nil {
-			r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartFailed", "deletion of pod %v with tablet %v failed", pod.Name, tabletKey)
-			return resultBuilder.Error(err)
-		} else {
-			// We must only release at most one tablet per reconcile loop, so exit if we do.
+		pod, ok := tabletPods[tabletKey]
+		if !ok {
+			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for desired tablet %v to be created.", tabletKey)
 			return resultBuilder.Result()
 		}
 
+		if rollout.Released(pod) {
+			// If any tablet has already been released, we should wait until it is finished to release another one.
+			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to finish release.", tabletKey)
+			return resultBuilder.Result()
+		}
 	}
 
-	// If we have no more scheduled tablets, uncascade the shard.
-	if !scheduledTablets {
-		r.uncascadeShard(vts)
+	// Retrieve tablet pod to be released during this reconcile.
+	tabletKey, pod := getNextScheduledTablet(tabletKeys, tabletPods)
+	if tabletKey == "" {
+		// If we have no more scheduled tablets, uncascade the shard.
+		if err := r.uncascadeShard(ctx, vts); err != nil {
+			r.recorder.Eventf(vts, v1.EventTypeWarning, "UncascadeFailed", "Client failed to mark shard as uncascading.")
+			return resultBuilder.Error(err)
+		}
+
+		r.recorder.Eventf(vts, v1.EventTypeNormal, "CascadeComplete", "Cascading changes have been successfully applied to shard.")
+		return resultBuilder.Result()
+	}
+
+	masterEligibleTablets := vts.Spec.MasterEligibleTabletCount()
+	deletePod := false
+	tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
+	// These two conditions guarantee that the tablet is a lone master.
+	if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
+		// If we have a lone master, we must delete it since reparenting is impossible.
+		deletePod = true
+	}
+
+	if err := r.releaseTabletPod(ctx, pod, deletePod); err != nil {
+		r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartBlocked", "release of Pod %v (tablet %v) failed: %v", pod.Name, tabletKey, err)
+		return resultBuilder.Error(err)
 	}
 
 	return resultBuilder.Result()
@@ -108,32 +108,30 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
 func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, pod *v1.Pod, deletePod bool) error {
 	if deletePod {
 		// TODO: Evict pods instead of deleting them directly, to respect PDBs.
-		if err := r.client.Delete(ctx, pod); err != nil {
-			return err
-		}
-	} else {
-		// Release the pod to be recreated with updates.
-		releasePod(pod)
+		return r.client.Delete(ctx, pod)
 	}
 
-	return nil
-}
-
-func releasePod(pod *v1.Pod) {
-	if !rollout.Scheduled(pod) || rollout.Released(pod) {
-			// If there's nothing scheduled, or already released, we're done.
-			return
-	}
-
+	// Release the pod to be recreated with updates.
 	rollout.Release(pod)
-	return
+	return r.client.Update(ctx, pod)
 }
 
-func (r *ReconcileVitessShard) uncascadeShard(vts *planetscalev2.VitessShard) {
-	if !rollout.Scheduled(vts) || rollout.Released(vts) {
-		// If there's nothing scheduled, or already released, we're done.
-		return
+func (r *ReconcileVitessShard) uncascadeShard(ctx context.Context, vts *planetscalev2.VitessShard) error {
+	rollout.Uncascade(vts)
+	return r.client.Update(ctx, vts)
+}
+
+func getNextScheduledTablet(tabletKeys []string, tabletPods map[string]*v1.Pod) (string, *v1.Pod){
+	nextTabletKey := ""
+	for _, tabletKey := range tabletKeys {
+		pod := tabletPods[tabletKey]
+		if !rollout.Scheduled(pod) {
+			continue
+		}
+
+		nextTabletKey = tabletKey
+		return nextTabletKey, pod
 	}
 
-	rollout.Uncascade(vts)
+	return nextTabletKey, nil
 }
--- a/pkg/controller/vitessshard/reconcile_rollout.go
+++ b/pkg/controller/vitessshard/reconcile_rollout.go
@@ -1,0 +1,214 @@
+package vitessshard
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apilabels "k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+	"planetscale.dev/vitess-operator/pkg/operator/rollout"
+)
+
+func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
+	if !rollout.Scheduled(vts) {
+		// If the shard is not scheduled for a rolling update, silently bail out and do nothing.
+		return resultBuilder.Result()
+	}
+
+	for _, tablet := range vts.Status.Tablets {
+		if tablet.Ready != corev1.ConditionTrue {
+			if tablet.Ready != corev1.ConditionTrue {
+				// If all tablets aren't healthy, we should bail and not perform a rolling restart.
+				r.recorder.Eventf(vts, corev1.EventTypeWarning, "RollingRestartFailed", "all tablets are not healthy")
+				return resultBuilder.Result()
+			}
+		}
+	}
+
+	podList := &corev1.PodList{}
+	listOpts := &client.ListOptions{
+		Namespace: vts.Namespace,
+		LabelSelector: apilabels.Set(map[string]string{
+			planetscalev2.ComponentLabel: planetscalev2.VttabletComponentName,
+			planetscalev2.ClusterLabel:   vts.Labels[planetscalev2.ClusterLabel],
+			planetscalev2.KeyspaceLabel:  vts.Labels[planetscalev2.KeyspaceLabel],
+			planetscalev2.ShardLabel:     vts.Spec.KeyRange.SafeName(),
+		}).AsSelector(),
+	}
+	if err := r.client.List(ctx, listOpts, podList); err != nil {
+		return resultBuilder.Error(err)
+	}
+
+	scheduledTablets := false
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		if !rollout.Scheduled(pod) {
+			continue
+		}
+
+		scheduledTablets = true
+		if err := r.releaseTabletPod(ctx, log, pod); err != nil {
+			r.recorder.Eventf(vts, corev1.EventTypeWarning, "RollingRestartFailed", "tablet deletion failed")
+			return resultBuilder.Error(err)
+		}
+	}
+
+	// If we have no more scheduled tablets, unschedule the shard.
+	if !scheduledTablets {
+		if err := r.unscheduleObject(ctx, vts); err != nil {
+			return resultBuilder.Error(err)
+		}
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, log *logrus.Entry, pod *corev1.Pod) error {
+	// Release the pod to be recreated with updates.
+	if err := r.releaseObject(ctx,pod); err != nil {
+		return err
+	}
+
+	// TODO: Evict pods before deleting them.
+	if err := r.client.Delete(ctx, pod); err != nil {
+		return err
+	}
+	if err := r.waitObjectUpdated(ctx, pod); err != nil {
+		return err
+	}
+	if err := r.waitPodReady(ctx, pod); err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


avaidyanatha:
Closes https://github.com/planetscale/planetscale-operator/issues/4334

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+	"planetscale.dev/vitess-operator/pkg/operator/rollout"
+)
+
+func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
+	if !rollout.Scheduled(vts) {
```
enisoc, line 28: This is checking whether the VitessShard's parent wants to apply changes to the VitessShard, which is not what we're interested in here. We want to handle the case after changes have been applied to the VitessShard, and the shard controller needs to then roll out changes to its child tablets.

There is not yet an annotation defined for triggering rolling restart. We should add one in the `rollout` package along with the other annotations. Maybe we can call the new annotation `rollout.planetscale.com/propagate` since it directs the object to propagate the rollout to its children?

We can then change this to check for `if !rollout.Propagated(vts) {`. Once we find that there are no children left with scheduled changes, we can remove the annotation with `rollout.Unpropagate(vts)` since we are done at that point.

pkg/controller/vitessshard/vitessshard_controller.go
```diff
@@ -184,6 +186,10 @@ func (r *ReconcileVitessShard) Reconcile(request reconcile.Request) (reconcile.R
 	tabletResult, err := r.reconcileTablets(ctx, vts)
 	resultBuilder.Merge(tabletResult, err)
 
+	// Perform rolling updates on tablets if needed.
```
enisoc, line 189: Copy the NOTE comment below to here as well since it also applies.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+		// If the shard is not scheduled for a rolling update, silently bail out and do nothing.
+		return resultBuilder.Result()
+	}
+
+	for _, tablet := range vts.Status.Tablets {
+		if tablet.Ready != corev1.ConditionTrue {
```
enisoc, line 34: Let's look like `tablet.Available` like the drain controller does, so we are naturally rate-limited by the time it takes for a new tablet to be considered stable after first becoming Ready.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			if tablet.Ready != corev1.ConditionTrue {
```
enisoc, line 35: Accidental doubling?

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+				// If all tablets aren't healthy, we should bail and not perform a rolling restart.
+				r.recorder.Eventf(vts, corev1.EventTypeWarning, "RollingRestartFailed", "all tablets are not healthy")
```
enisoc, line 37: This is going to happen as an expected, normal occurrence during rollouts, so we shouldn't make it sound bad per se. Also let's give some more detail of what we see that made us pause.

```suggestion
				r.recorder.Eventf(vts, corev1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to be Available.", tabletAlias)
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+
+	if !rollout.Scheduled(vts) {
+		// If the shard is not scheduled for a rolling update, silently bail out and do nothing.
+		return resultBuilder.Result()
+	}
+
+	for _, tablet := range vts.Status.Tablets {
```
enisoc, line 33: Since we're going to be sending an event based on the first tablet that isn't Available, we should go ahead and establish a deterministic ordering and check status in that order. Otherwise we may spam events for a different tablet each time, defeating k8s event consolidation.

We can just take all the keys (tablet aliases) from the `vts.Status.Tablets` map and sort them, and use that same order for the actual rollout.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+		if tablet.Ready != corev1.ConditionTrue {
+			if tablet.Ready != corev1.ConditionTrue {
+				// If all tablets aren't healthy, we should bail and not perform a rolling restart.
+				r.recorder.Eventf(vts, corev1.EventTypeWarning, "RollingRestartFailed", "all tablets are not healthy")
+				return resultBuilder.Result()
+			}
+		}
+	}
```
enisoc, line 41: In this same loop, we should also similarly bail out if any tablet has already been released. We have to wait until nothing is released before we decide to release anything else. I guess that means we'll have to list the pods before this loop and put them in a map by alias string.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			planetscalev2.KeyspaceLabel:  vts.Labels[planetscalev2.KeyspaceLabel],
+			planetscalev2.ShardLabel:     vts.Spec.KeyRange.SafeName(),
+		}).AsSelector(),
+	}
+	if err := r.client.List(ctx, listOpts, podList); err != nil {
+		return resultBuilder.Error(err)
+	}
+
+	scheduledTablets := false
+	for i := range podList.Items {
```
enisoc, line 58: I don't think we can rely on the pod list to be in deterministic order. Let's use the sorted alias strings as mentioned above.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+		pod := &podList.Items[i]
+		if !rollout.Scheduled(pod) {
+			continue
+		}
+
+		scheduledTablets = true
+		if err := r.releaseTabletPod(ctx, log, pod); err != nil {
+			r.recorder.Eventf(vts, corev1.EventTypeWarning, "RollingRestartFailed", "tablet deletion failed")
```
enisoc, line 66: Include the tablet alias and Pod name in the event.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			return resultBuilder.Error(err)
+		}
+	}
```
enisoc, line 69: We should break the loop if we actually released a tablet. We only ever release at most one tablet per reconcile.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// If we have no more scheduled tablets, unschedule the shard.
```
enisoc, line 71:
```suggestion
	// If we have no more scheduled tablets, unpropagate the shard.
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}
+
+		if !rollout.Scheduled(objectMeta(obj)) {
+			// This is what we're waiting for.
+			return nil
+		}
+
+		// Keep waiting.
+		time.Sleep(1 * time.Second)
```
enisoc, line 120: There's no sleeping allowed in controllers. If you need to wait for something, check again in a subsequent reconcile pass.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, log *logrus.Entry, pod *corev1.Pod) error {
+	// Release the pod to be recreated with updates.
+	if err := r.releaseObject(ctx,pod); err != nil {
+		return err
+	}
+
+	// TODO: Evict pods before deleting them.
```
enisoc, line 87: Eviction is an alternative to deletion; you wouldn't do both.

```suggestion
	// TODO: Evict Pods instead of deleting them directly, to respect PDBs.
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			return resultBuilder.Error(err)
+		}
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, log *logrus.Entry, pod *corev1.Pod) error {
+	// Release the pod to be recreated with updates.
+	if err := r.releaseObject(ctx,pod); err != nil {
```
enisoc, line 83: We should either release or delete, not both.

If the tablet is a master and there are no other master-eligible tablets in the shard (reparent is impossible), then we need to evict/delete.

In the normal case, marking it as released is enough; the reconciler will automatically drain it (reparent if necessary) and then delete it.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+		return err
+	}
+
+	// TODO: Evict pods before deleting them.
+	if err := r.client.Delete(ctx, pod); err != nil {
+		return err
+	}
+	if err := r.waitObjectUpdated(ctx, pod); err != nil {
```
enisoc, line 91: Unlike the CLI tool, we don't need to wait for anything in this controller. We simply return and check again on the next reconcile.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			return nil
+		}
+
+		// Keep waiting.
+		time.Sleep(1 * time.Second)
+	}
+}
+
+func (r *ReconcileVitessShard) releaseObject(ctx context.Context, obj runtime.Object) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
enisoc, line 141: Unlike the CLI tool, we don't usually RetryOnConflict in controllers. Instead, we just return the error and let that trigger a requeue.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+			return nil
+		}
+		rollout.Unschedule(objMeta)
+		return r.client.Update(ctx, obj)
+	})
+}
+
+// regetObject does a fresh Get from the server to update the contents of an
+// existing object.
+func regetObject(ctx context.Context, kubeClient client.Client, obj runtime.Object) error {
```
enisoc, line 172: Unlike in the CLI tool, we shouldn't get doing "reget" in a controller. This is related to the fact that we don't retry on conflict. If there would be a conflict without a reget, we should not hide that. We should requeue and let the whole controller reconcile again based on the new state, instead of partially observing the new state only here.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,214 @@
...
+		if !rollout.Scheduled(objMeta) || rollout.Released(objMeta) {
+			// If there's nothing scheduled, or already released, we're done.
+			return nil
+		}
+		rollout.Release(objMeta)
+		return r.client.Update(ctx, obj)
+	})
+}
+
+func (r *ReconcileVitessShard) unscheduleObject(ctx context.Context, obj runtime.Object) error {
```
enisoc, line 155: Make this `unpropagateShard` and just pass `*planetscalev2.VitessShard` so we don't have to use the `objectMeta` casting. We only ever need this on that one type here.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Yep. This should be a smell check for deleting this entire thing - thanks for the reminder.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Makes sense. Done.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Done.

pkg/controller/vitessshard/vitessshard_controller.go
```diff
@@ -135,6 +135,8 @@ type ReconcileVitessShard struct {
 	resync     *resync.Periodic
 	recorder   record.EventRecorder
 	reconciler *reconciler.Reconciler
+
+	// This client is required
```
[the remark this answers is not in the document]
avaidyanatha: Yep, fixed.

pkg/controller/vitessshard/vitessshard_controller.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+			planetscalev2.ShardLabel:     vts.Spec.KeyRange.SafeName(),
+		}).AsSelector(),
+	}
+
+	if err := r.client.List(ctx, listOpts, podList); err != nil {
+		return resultBuilder.Error(err)
+	}
+
+	// Safety checks and rolling updates must be deterministically ordered, so we access and sort pods by tablet alias.
+	tabletPods := make(map[string]v1.Pod)
```
PrismaPhonic, line 43: Since this is a k8s type, I would suggest using a pointer type:
```suggestion
	tabletPods := make(map[string]*v1.Pod)
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+	for i := range podList.Items {
+		pod := podList.Items[i]
```
PrismaPhonic, line 45:
```suggestion
		pod := &podList.Items[i]
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+		tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
+		// These two conditions guarantee that the tablet is a lone master.
+		if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
+			// If we have a lone master, we must delete it since reparenting is impossible.
+			deletePod = true
+		}
+
+		if err := r.releaseTabletPod(ctx, &pod, deletePod); err != nil {
+			r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartFailed", "deletion of pod %v with tablet %v failed", pod.Name, tabletKey)
+			return resultBuilder.Error(err)
```
PrismaPhonic, line 92: I'm not sure if it makes sense to fail the entire process here. I would defer to @enisoc 's judgment on this one. Potential suggestion: 
```suggestion
			resultBuilder.Error(err)
```

And then check for error after the loop and bail then so we at least get through as many pods as possible before a retry.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Yeah this makes more sense, done.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_rollout.go
avaidyanatha: `reconcileRollout` will at most release one Pod for a rolling update no matter what. Even if we don't error out here, we would release only the next scheduled pod. This might not be worth losing our deterministic rollout, but I would also defer to @enisoc's judgment.

pkg/controller/vitessshard/reconcile_rollout.go
enisoc: I like "cascade" even better!

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+	// Safety checks and rolling updates must be deterministically ordered, so we access and sort pods by tablet alias.
+	tabletPods := make(map[string]*v1.Pod)
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		tabletAlias := vttablet.AliasFromPod(pod)
+		tabletKey := topoproto.TabletAliasString(&tabletAlias)
+		tabletPods[tabletKey] = pod
+	}
+
+	tabletKeys := make([]string, 0, len(tabletPods))
```
enisoc, line 51: We should form the `tabletKeys` list based on the keys in the `vts.Status.Tablets` map. That status map is guaranteed to have an entry for every _desired_ tablet, whereas `tabletPods` will only list tablets that currently exist. We need to use the desired list so we don't forget to account for tablets whose Pods are temporarily missing, such as when we ourselves delete them for update.

If we find that a desired key is missing from `tabletPods`, we'll emit an event saying we're waiting for that Pod to be recreated and bail out.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+	for key := range tabletPods {
+		tabletKeys = append(tabletKeys, key)
+	}
+	sort.Strings(tabletKeys)
+
+	for _, tabletKey := range tabletKeys {
+		tablet := vts.Status.Tablets[tabletKey]
+		if tablet.Available != v1.ConditionTrue {
+			// If all tablets aren't healthy, we should bail and not perform a rolling restart.
```
enisoc, line 60: Just to be a little clearer:

```suggestion
			// If any tablets are unhealthy, we should bail and not perform a rolling restart.
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to be Available.", tabletKey)
+			return resultBuilder.Result()
+		}
+
+		pod := tabletPods[tabletKey]
+		if rollout.Released(pod) {
+			// If any tablet has already been released, we should wait until it is finished to release another one.
+			r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutPaused", "Waiting for tablet %v to finish release.", tabletKey)
```
enisoc, line 68: Also bail out after recording the event.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+		deletePod := false
+		tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
+		// These two conditions guarantee that the tablet is a lone master.
+		if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
+			// If we have a lone master, we must delete it since reparenting is impossible.
+			deletePod = true
+		}
+
+		if err := r.releaseTabletPod(ctx, pod, deletePod); err != nil {
+			r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartFailed", "deletion of pod %v with tablet %v failed", pod.Name, tabletKey)
```
enisoc, line 91: The error might be transient, so let's soften the event a bit. Also we didn't necessarily try to delete the pod. We might just be trying to annotate it. The `err` will have more detail on what we tried to do and what happened.

```suggestion
			r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartBlocked", "release of Pod %v (tablet %v) failed: %v", pod.Name, tabletKey, err)
```

pkg/controller/vitessshard/reconcile_rollout.go
enisoc: > check for error after the loop and bail then so we at least get through as many pods as possible

We actually don't want to get through any other Pods in this case. We are looking for the one Pod whose turn is up next, and then we try to process that one Pod.

Maybe we should refactor the code a bit to enforce this more clearly: instead of doing this inside a loop, make a helper like `getFirstScheduledTabletKey` which returns a single tabletKey, or `""` if no tablets are scheduled. Then it's clear that we only ever look at one Pod per reconcile.

It also makes it easier to treat `uncascadeShard` as a bail-out so the real meat of the function (releasing the next tablet) is less indented.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+		r.uncascadeShard(vts)
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, pod *v1.Pod, deletePod bool) error {
+	if deletePod {
+		// TODO: Evict pods instead of deleting them directly, to respect PDBs.
+		if err := r.client.Delete(ctx, pod); err != nil {
```
enisoc, line 111: You can just `return r.client.Delete(...)` so that the rest of the function is unindented.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	} else {
+		// Release the pod to be recreated with updates.
+		releasePod(pod)
+	}
+
+	return nil
+}
+
+func releasePod(pod *v1.Pod) {
```
enisoc, line 122: This helper is unnecessary. We already check that that tablet is scheduled and not released elsewhere.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (r *ReconcileVitessShard) releaseTabletPod(ctx context.Context, pod *v1.Pod, deletePod bool) error {
+	if deletePod {
+		// TODO: Evict pods instead of deleting them directly, to respect PDBs.
+		if err := r.client.Delete(ctx, pod); err != nil {
+			return err
+		}
+	} else {
+		// Release the pod to be recreated with updates.
+		releasePod(pod)
```
enisoc, line 116: The `rollout.Release()` function only adds the annotation to our in-memory copy of the Pod. We then need to call `client.Update()` to send it to the server.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+	if !rollout.Scheduled(pod) || rollout.Released(pod) {
+			// If there's nothing scheduled, or already released, we're done.
+			return
+	}
+
+	rollout.Release(pod)
+	return
+}
+
+func (r *ReconcileVitessShard) uncascadeShard(vts *planetscalev2.VitessShard) {
```
enisoc, line 132: This also needs to call `client.Update()` and return the `error`.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,139 @@
...
+	if !rollout.Scheduled(vts) || rollout.Released(vts) {
```
enisoc, line 133: This is checking whether the shard's parent has scheduled or released changes on us, which is irrelevant to this file. Here we only care about our own children.

pkg/operator/rollout/rollout.go
```diff
@@ -53,6 +53,10 @@ const (
 	// ReleasedAnnotation is the annotation whose presence indicates that the
 	// object's controller may now apply any pending changes to the object.
 	ReleasedAnnotation = AnnotationPrefix + "/" + "released"
+
+	// CascadeAnnotation is the annotation whose presence indicates that the
+	// object's controller may now propagate cascading changes to its children.
```
enisoc, line 58:
```suggestion
	// object's controller should now release any scheduled changes to its children.
	// The controller will remove the annotation when all children are updated.
```

pkg/operator/rollout/rollout.go
```diff
@@ -73,6 +77,15 @@ func Released(obj metav1.Object) bool {
 	return present
 }
 
+func Cascaded(obj metav1.Object) bool {
```
enisoc, line 80: Maybe this should be `Cascading`? The other annotations are telling us "this thing happened in the past" -- a change _was scheduled_ or _was released_. This annotation is more of an on/off switch -- "is this parent currently cascading or not?"

Also add a doc comment.

pkg/operator/rollout/rollout.go
```diff
@@ -146,3 +159,42 @@ func Unrelease(obj metav1.Object) {
 	delete(ann, ReleasedAnnotation)
 	obj.SetAnnotations(ann)
 }
+
+/*
+Cascade annotates an object as being ready to have cascading changes
+applied to its children.
```
enisoc, line 165:
```suggestion
Cascade annotates an object to tell its controller to release
any scheduled changes to its children.
```

pkg/operator/rollout/rollout.go
```diff
@@ -146,3 +159,42 @@ func Unrelease(obj metav1.Object) {
...
+
+If the object has already been marked as cascaded, this has no effect.
```
enisoc, line 167:
```suggestion
If the object has already been marked as cascading, this has no effect.
```

pkg/operator/rollout/rollout.go
```diff
@@ -146,3 +159,42 @@ func Unrelease(obj metav1.Object) {
...
+
+Note that this only mutates the provided, in-memory object to add the
+annotation; the caller is responsible for sending the updated object to
+the server.
+*/
+func Cascade(obj metav1.Object) {
+	ann := obj.GetAnnotations()
+	if _, present := ann[CascadeAnnotation]; present {
+		// The object has already been released.
```
enisoc, line 176:
```suggestion
		// The object has already been marked as cascading.
```

pkg/controller/vitessshard/reconcile_rollout.go
PrismaPhonic: Ahhh yeah, then I would say the code could be a bit more clear with a helper like that which filters first. As a reader this came across as if we were releasing more than one pod.

pkg/operator/rollout/rollout.go
avaidyanatha: Done.

pkg/operator/rollout/rollout.go
avaidyanatha: Yeah that makes sense, done and done.

pkg/operator/rollout/rollout.go
avaidyanatha: Done.

pkg/operator/rollout/rollout.go
avaidyanatha: Done.

pkg/operator/rollout/rollout.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,137 @@
...
+			return resultBuilder.Result()
+		}
+	}
+
+	// Retrieve tablet pod to be released during this reconcile.
+	tabletKey, pod := getNextScheduledTablet(tabletKeys, tabletPods)
+	if tabletKey == "" {
+		// If we have no more scheduled tablets, uncascade the shard.
+		if err := r.uncascadeShard(ctx, vts); err != nil {
+			r.recorder.Eventf(vts, v1.EventTypeWarning, "UncascadeFailed", "Client failed to mark shard as uncascading.")
```
enisoc, line 83: Let's make this easier for users to understand, and also log the error:

```suggestion
			r.recorder.Eventf(vts, v1.EventTypeWarning, "UncascadeFailed", "Failed to mark cascading shard rollout as complete: %v", err)
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,137 @@
...
+			return resultBuilder.Error(err)
+		}
+
+		r.recorder.Eventf(vts, v1.EventTypeNormal, "CascadeComplete", "Cascading changes have been successfully applied to shard.")
```
enisoc, line 87:
```suggestion
		r.recorder.Eventf(vts, v1.EventTypeNormal, "RollingRestartComplete", "Cascading rollout of tablets is complete.")
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,137 @@
...
+func getNextScheduledTablet(tabletKeys []string, tabletPods map[string]*v1.Pod) (string, *v1.Pod){
+	nextTabletKey := ""
+	for _, tabletKey := range tabletKeys {
+		pod := tabletPods[tabletKey]
+		if !rollout.Scheduled(pod) {
+			continue
+		}
+
+		nextTabletKey = tabletKey
+		return nextTabletKey, pod
```
enisoc, line 133:
```suggestion
		return tabletKey, pod
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return nextTabletKey, nil
```
enisoc, line 136:
```suggestion
	return "", nil
```

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -0,0 +1,137 @@
...
+	tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
+	// These two conditions guarantee that the tablet is a lone master.
+	if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
+		// If we have a lone master, we must delete it since reparenting is impossible.
+		deletePod = true
+	}
+
+	if err := r.releaseTabletPod(ctx, pod, deletePod); err != nil {
+		r.recorder.Eventf(vts, v1.EventTypeWarning, "RollingRestartBlocked", "release of Pod %v (tablet %v) failed: %v", pod.Name, tabletKey, err)
+		return resultBuilder.Error(err)
```
PrismaPhonic, line 102: At this point, now that we aren't in a loop, anymore, so I think preferring less exit points will be clearer. We will return `resultBuilder.Result()` in a moment anyways, so we can just do `resultBuilder.Error(err)` with no return here and then return (as we already will) at the final exit on line 105. Calling `resultBuilder.Result()` on line 105 will carry with it any errors from `resultBuilder.Error(err)`.

PR #78, opened 2020-04-21
Update disk size in place when reconfigured

avaidyanatha:
## Main Changes
- Disk size updates are now not held behind manual rollouts and can be updated in place.


pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+		return
+	}
+
+	updateDiskSize(dst, src)
+}
+
+func updateDiskSize(dst *planetscalev2.VitessKeyspaceTemplate, src planetscalev2.VitessKeyspaceTemplate) {
+	for i := range dst.Partitionings {
+		dstPartitioning := dst.Partitionings[i]
+		if dstPartitioning.Equal != nil {
```
enisoc, line 37: We also need to handle Custom partitionings.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+	if !validateKeyspacePartitionings(*dst,src) {
+		return
+	}
+
+	updateDiskSize(dst, src)
+}
+
+func updateDiskSize(dst *planetscalev2.VitessKeyspaceTemplate, src planetscalev2.VitessKeyspaceTemplate) {
+	for i := range dst.Partitionings {
+		dstPartitioning := dst.Partitionings[i]
```
enisoc, line 36: The purpose of the index trick is so we can make this a pointer instead of a copy:

```suggestion
		dstPartitioning := &dst.Partitionings[i]
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+		if dstPartitioning.Equal != nil {
+			updatePartitioningDiskSize(dstPartitioning.Equal, *src.Partitionings[i].Equal)
+		}
+	}
+}
+
+func updatePartitioningDiskSize(dst *planetscalev2.VitessKeyspaceEqualPartitioning, src planetscalev2.VitessKeyspaceEqualPartitioning) {
```
enisoc, line 43:
```suggestion
func updateEqualPartitioningDiskSize(dst, src *planetscalev2.VitessKeyspaceEqualPartitioning) {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+	srcLoop:
+		for i := range dst.ShardTemplate.TabletPools {
+			dstTablet := dst.ShardTemplate.TabletPools[i]
+			var requestedTablet *planetscalev2.VitessShardTabletPool
+
+			for j := range src.ShardTemplate.TabletPools {
```
enisoc, line 49: Break this out into a subroutine that finds the matching tablet pool, if any. Then we won't need the loop label either.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+		if dstPartitioning.Equal != nil {
+			updatePartitioningDiskSize(dstPartitioning.Equal, *src.Partitionings[i].Equal)
+		}
+	}
+}
+
+func updatePartitioningDiskSize(dst *planetscalev2.VitessKeyspaceEqualPartitioning, src planetscalev2.VitessKeyspaceEqualPartitioning) {
+	srcLoop:
+		for i := range dst.ShardTemplate.TabletPools {
+			dstTablet := dst.ShardTemplate.TabletPools[i]
```
enisoc, line 46:
```suggestion
			dstTablet := &dst.ShardTemplate.TabletPools[i]
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+			var requestedTablet *planetscalev2.VitessShardTabletPool
+
+			for j := range src.ShardTemplate.TabletPools {
+				srcTablet := src.ShardTemplate.TabletPools[j]
```
enisoc, line 50:
```suggestion
				srcTablet := &src.ShardTemplate.TabletPools[j]
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+				// Match each dst tablet pool with its unique Type, Cell tuple in src.
+				if srcTablet.Type == dstTablet.Type && srcTablet.Cell == dstTablet.Cell {
```
enisoc, line 52: Let's pull this out into a method defined on VitessShardTabletPool to determine whether two objects are talking about the same pool. We may need to add more values to this tuple later, so we should make sure the comparison logic isn't scattered around.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+		if dst.Shards[i].KeyRange != src.Shards[i].KeyRange {
+			return false
+		}
+	}
+
+	return true
+}
+
+// DiskResource updates disk size entries in 'dst' based on the values in 'src'.
+func DiskResource(dst, src *corev1.ResourceList) {
```
enisoc, line 140: We only need dst to be a pointer, in case we need to allocate it. We don't allocate src, and ResourceList is already a map value, which is essentially a pointer already.

Also, since the ResourceList calls the entry Storage, we should reflect that name.

```suggestion
func StorageResource(dst *corev1.ResourceList, src corev1.ResourceList) {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+			*dst = *src
+		}
+		return
+	}
+	for srcKey, srcVal := range *src {
+		if srcKey != corev1.ResourceStorage {
+			continue
+		}
+		(*dst)[srcKey] = srcVal
+	}
```
enisoc, line 152: Let's allocate a new ResourceList instead of sharing, to avoid accidental mutation of inputs down the line. Also, you don't need to iterate to find the storage value because ResourceList is an alias for a map type.

```suggestion
	srcVal, ok := src[corev1.ResourceStorage]
	if !ok {
		return
	}
	if *dst == nil {
		*dst = make(corev1.ResourceList)
	}
	(*dst)[corev1.ResourceStorage] = srcVal
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+
+			srcRequests := &requestedTablet.DataVolumeClaimTemplate.Resources.Requests
+			dstRequests := &dstTablet.DataVolumeClaimTemplate.Resources.Requests
+
+			DiskResource(dstRequests, srcRequests)
+		}
+}
+
+func validateKeyspacePartitionings(dst planetscalev2.VitessKeyspaceTemplate, src planetscalev2.VitessKeyspaceTemplate) bool {
```
enisoc, line 73: It seems like we only need the actual lists of partitionings.

```suggestion
func validateKeyspacePartitionings(dst, src []planetscalev2.VitessKeyspacePartitioning) bool {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return valid
+}
+
+func validatePartitioningTypes(dst *planetscalev2.VitessKeyspacePartitioning, src *planetscalev2.VitessKeyspacePartitioning) bool {
+	if dst.Equal != nil && src.Equal == nil {
+		return false
+	} else if dst.Equal == nil && src.Equal != nil {
+		return false
+	}
```
enisoc, line 109: We also need to check Custom partitionings. It's not guaranteed that one is always nil and one is always non-nil.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+
+	valid := false
+	if dst.Equal != nil {
+		valid = validateEqualPartitioning(dst.Equal, src.Equal)
+	} else {
+		valid = validateCustomPartitioning(dst.Custom, src.Custom)
+	}
+
+	return valid
```
enisoc, line 101:
```suggestion
	if dst.Equal != nil {
		return validateEqualPartitioning(dst.Equal, src.Equal)
	}
	if dst.Custom != nil {
		return validateCustomPartitioning(dst.Custom, src.Custom)
	}

	return false
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,153 @@
...
+
+package update
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+)
+
+func KeyspaceDiskSize(dst *planetscalev2.VitessKeyspaceTemplate, src planetscalev2.VitessKeyspaceTemplate) {
```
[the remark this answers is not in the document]
enisoc: This function is a good candidate for a unit test. We should check that various buried disk sizes get updated, and check that invalid cases (not matching) are left untouched. We can also check that nothing besides the disk sizes is touched.

enisoc:
This handles propagation of disk size changes from VT to VTK, but we then need to propagate from VTK to VTS. We'll need to add an UpdateInPlace handler to reconcile_shards.go.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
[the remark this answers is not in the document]
avaidyanatha: Will add unit tests in the next commit cycle, also 1 and 2 of the first comment are done.

pkg/operator/update/resources.go
enisoc: It _should_ be the case that exactly one of them is non-nil, but we don't yet have a validating webhook to prevent invalid CRDs from being persisted, so we need to be liberal in what we accept.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+	for i := range dst {
+		srcTabletDiskSize := src[i].DataVolumeClaimTemplate.Resources.Requests[corev1.ResourceStorage]
+		dst[i].DataVolumeClaimTemplate.Resources.Requests[corev1.ResourceStorage] = srcTabletDiskSize
+	}
+}
+
+// KeyspaceDiskSize updates values in 'dst' based on values in 'src'.
+// It leaves extra entries (found in 'dst' but not in 'src') untouched,
+// since those might be set by mutating admission webhooks or other controllers.
+// TODO: Add unit test to check that various buried disk sizes get updated and
```
PrismaPhonic, line 35: Can we please flip the order of this so the todo is above the function document, with a space between each to clearly separate them?

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+		srcTablet := &src[i]
+		if srcTablet.IsMatch(dst) {
+			return srcTablet
+		}
+	}
+
+	return nil
+}
+
+func validateKeyspacePartitionings(dst, src []planetscalev2.VitessKeyspacePartitioning) bool {
```
PrismaPhonic, line 108: I think changing the name would make this read better at call sights:

```suggestion
func keyspacePartitioningsAreValid(dst, src []planetscalev2.VitessKeyspacePartitioning) bool {
```

Example:

```
if !keyspacePartitioningsAreValid {
    // bail
}
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+		// Validate that partitionings have the same type and shard count.
+		if !validatePartitionings(&dst[i], &src[i]) {
+			return false
+		}
+	}
+
+	return true
+}
+
+func validatePartitionings(dst *planetscalev2.VitessKeyspacePartitioning, src *planetscalev2.VitessKeyspacePartitioning) bool {
```
PrismaPhonic, line 124: Same here. Functions that return bools should flow when used as conditional checks by callers:

```suggestion
func partitioningsMatch(dst *planetscalev2.VitessKeyspacePartitioning, src *planetscalev2.VitessKeyspacePartitioning) bool {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+	if dst.Equal != nil {
+		return validateEqualPartitioning(dst.Equal, src.Equal)
+	}
+	if dst.Custom != nil {
+		return validateCustomPartitioning(dst.Custom, src.Custom)
+	}
+	return false
+}
+
+func validatePartitioningTypes(dst *planetscalev2.VitessKeyspacePartitioning, src *planetscalev2.VitessKeyspacePartitioning) bool {
```
PrismaPhonic, line 138:
```suggestion
func partitioningTypesMatch(dst *planetscalev2.VitessKeyspacePartitioning, src *planetscalev2.VitessKeyspacePartitioning) bool {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+	if dst.Custom != nil && src.Custom == nil {
+		return false
+	} else if dst.Custom == nil && src.Custom != nil  {
+		return false
+	}
+
+	return true
+}
+
+func validateEqualPartitioning(dst *planetscalev2.VitessKeyspaceEqualPartitioning, src *planetscalev2.VitessKeyspaceEqualPartitioning) bool {
```
PrismaPhonic, line 154:
```suggestion
func equalPartitioningsMatch(dst *planetscalev2.VitessKeyspaceEqualPartitioning, src *planetscalev2.VitessKeyspaceEqualPartitioning) bool {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+	// Validate that the number of shards is the same.
+	if dst.Parts != src.Parts {
+		return false
+	}
+
+	return true
+ }
+
+func validateCustomPartitioning(dst *planetscalev2.VitessKeyspaceCustomPartitioning, src *planetscalev2.VitessKeyspaceCustomPartitioning) bool {
```
PrismaPhonic, line 163:
```suggestion
func customPartitioningsMatch(dst *planetscalev2.VitessKeyspaceCustomPartitioning, src *planetscalev2.VitessKeyspaceCustomPartitioning) bool {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+)
+
+func ShardDiskSize(dst []planetscalev2.VitessShardTabletPool, src []planetscalev2.VitessShardTabletPool) {
+	for i := range dst {
+		srcTabletDiskSize := src[i].DataVolumeClaimTemplate.Resources.Requests[corev1.ResourceStorage]
```
enisoc, line 27: We shouldn't assume the tablet pools are in the same order. We should look up matching pools like we discussed for the keyspace level.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+		dst[i].DataVolumeClaimTemplate.Resources.Requests[corev1.ResourceStorage] = srcTabletDiskSize
```
enisoc, line 28: We should use the `StorageResource` method, which allocates if necessary.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+
+	updateDiskSize(dst, src)
+}
+
+func updateDiskSize(dst, src *planetscalev2.VitessKeyspaceTemplate) {
+	for i := range dst.Partitionings {
+		dstPartitioning := &dst.Partitionings[i]
+		if dstPartitioning.Equal != nil {
+			updateEqualPartitioningDiskSize(dstPartitioning.Equal, src.Partitionings[i].Equal)
+		} else {
```
enisoc, line 52: Just to be as liberal as we can in what we accept:

```suggestion
		}
		if dstPartitioning.Custom != nil {
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,189 @@
...
+			updateCustomPartitioningDiskSize(dstPartitioning.Custom, src.Partitionings[i].Custom)
+		}
+	}
+}
+
+func updateEqualPartitioningDiskSize(dst, src *planetscalev2.VitessKeyspaceEqualPartitioning) {
```
enisoc, line 58: If you make this function directly accept the two slices of tablet pools, you can share it for both equal and custom partitionings, as well as for use by `ShardDiskSize`.

avaidyanatha: Hmmm my senses told me there was a way to do this... Done!

pkg/operator/update/resources.go
avaidyanatha: Yeah I agree, done!

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Gotcha!

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/controller/vitesscluster/reconcile_keyspaces.go
```diff
@@ -189,6 +189,9 @@ func updateVitessKeyspaceInPlace(key client.ObjectKey, vtk *planetscalev2.Vitess
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vtk.Labels, newKeyspace.Labels)
 
+	// Update disk size.
+	update.KeyspaceDiskSize(&vtk.Spec.VitessKeyspaceTemplate, keyspace)
```
enisoc, line 193: You had the source right to begin with:

```suggestion
	update.KeyspaceDiskSize(&vtk.Spec.VitessKeyspaceTemplate, &newKeyspace.Spec.VitessKeyspaceTemplate)
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,173 @@
...
+
+package update
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+)
+
+func ShardDiskSize(dst []planetscalev2.VitessShardTabletPool, src []planetscalev2.VitessShardTabletPool) {
```
enisoc, line 25: Add doc comment for exported function.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,173 @@
...
+	updateTabletPoolDiskSize(dst, src)
+}
+
+// TODO: Add unit test to check that various buried disk sizes get updated and
+// check that invalid cases (not matching) are left untouched. We can also check
+// that nothing besides the disk sizes is touched.
+
+// KeyspaceDiskSize updates values in 'dst' based on values in 'src'.
+// It leaves extra entries (found in 'dst' but not in 'src') untouched,
+// since those might be set by mutating admission webhooks or other controllers.
```
enisoc, line 35:
```suggestion
// KeyspaceDiskSize updates disk sizes in 'dst' based on values in 'src'.
// It does not update any other values besides disk sizes.
// It leaves extra entries (found in 'dst' but not in 'src') untouched.
```

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,173 @@
...
+	updateDiskSize(dst, src)
+}
+
+func updateDiskSize(dst, src *planetscalev2.VitessKeyspaceTemplate) {
+	for i := range dst.Partitionings {
+		dstPartitioning := &dst.Partitionings[i]
+		srcPartitioning := &src.Partitionings[i]
+
+		if dstPartitioning.Equal != nil {
+			updateTabletPoolDiskSize(dstPartitioning.Equal.ShardTemplate.TabletPools, src.Partitionings[i].Equal.ShardTemplate.TabletPools)
```
enisoc, line 51: Might as well use the var defined above:

```suggestion
			updateTabletPoolDiskSize(dstPartitioning.Equal.ShardTemplate.TabletPools, srcPartitioning.Equal.ShardTemplate.TabletPools)
```

pkg/controller/vitesscluster/reconcile_keyspaces.go
avaidyanatha: Done. Also I'm not sure why this confused me so much.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Done.

pkg/operator/update/resources.go
avaidyanatha: Yep makes sense, done.

pkg/operator/update/resources.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// ShardDiskSize updates values in 'dst' based on values in 'src'.
+// It does not update any other values besides disk sizes.
+// It leaves extra entries (found in 'dst' but not in 'src') untouched.
+func ShardDiskSize(dst []planetscalev2.VitessShardTabletPool, src []planetscalev2.VitessShardTabletPool) {
+	updateTabletPoolDiskSize(dst, src)
+}
+
+// TODO: Add unit test to check that various buried disk sizes get updateds and
```
enisoc, line 32: Just leaving this as a reminder to remove this comment when you add the test.

PR #83, opened 2020-05-02
Restart tablet Pods to complete pending PVC filesystem resizes.

avaidyanatha:
## Main Changes
- This marks pods as ready to be restarted if they have undergone a disk size change. It checks that their associated PVC has received the updates and only cascades the shard when all tablet pools within a shard are ready.

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_methods.go
+++ b/pkg/apis/planetscale/v2/vitessshard_methods.go
@@ -218,6 +218,7 @@ func (s *VitessShardStatus) DeepCopyConditions() map[VitessShardConditionType]*V
 	return out
 }
 
+// TabletAliases returns a sorted list of desired tablet aliases for the shard.
 func (s *VitessShardStatus) TabletAliases() []string {
 	tabletKeys := make([]string, 0, len(s.Tablets))
 	for key := range s.Tablets {
--- a/pkg/apis/planetscale/v2/vitessshard_methods.go
+++ b/pkg/apis/planetscale/v2/vitessshard_methods.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v2
 
 import (
+	"sort"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -215,4 +216,14 @@ func (s *VitessShardStatus) DeepCopyConditions() map[VitessShardConditionType]*V
 	}
 
 	return out
+}
+
+func (s *VitessShardStatus) TabletAliases() []string {
+	tabletKeys := make([]string, 0, len(s.Tablets))
+	for key := range s.Tablets {
+		tabletKeys = append(tabletKeys, key)
+	}
+	sort.Strings(tabletKeys)
+
+	return tabletKeys
 }
--- a/pkg/controller/vitessshard/reconcile_disk.go
+++ b/pkg/controller/vitessshard/reconcile_disk.go
@@ -1,0 +1,124 @@
+package vitessshard
+
+import (
+	"context"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	apitypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+	"planetscale.dev/vitess-operator/pkg/operator/rollout"
+)
+
+const (
+	pvcDiskSizeAnnotation = "pvc-disk-size"
+)
+
+func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+	changed := false
+	tabletKeys := tabletKeysFromShard(vts)
+	tabletPods, err := r.tabletPodsFromShard(ctx, vts)
+	if err != nil {
+		return resultBuilder.Error(err)
+	}
+
+	for i  := range vts.Spec.TabletPools {
+		tabletPool := &vts.Spec.TabletPools[i]
+		requestedDisk := tabletPool.DataVolumeClaimTemplate.Resources.Requests[v1.ResourceStorage].String()
+		poolTablets := tabletsForPool(vts, tabletKeys, tabletPool.Cell, string(tabletPool.Type))
+
+		for _, tabletKey := range poolTablets {
+			pod, ok := tabletPods[tabletKey]
+			if !ok {
+				continue
+			}
+
+			tabletDiskSize, ok := pod.Annotations[pvcDiskSizeAnnotation]
+			if !ok {
+				// If it's never been set before, apply the disk size annotation.
+				setDiskSizeAnnotation(pod, requestedDisk)
+				continue
+			}
+
+			// If disk size is unchanged, move on to the next tablet.
+			if tabletDiskSize == requestedDisk {
+				continue
+			}
+
+			// If there are disk size changes, mark the shard as changed to cascade later.
+			changed = true
+
+			pvc, err := r.claimForTabletPod(ctx, pod)
+			if err != nil {
+				return resultBuilder.Error(err)
+			}
+
+			// Make sure that the tablet's PVC has the updated disk size before proceeding.
+			if pvc.Spec.Resources.Requests[v1.ResourceStorage].String() != requestedDisk {
+				r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutCheckPaused", "Waiting for pvc %v to reflect updated disk.", pvc.Name)
+				return resultBuilder.Result()
+			}
+
+			// Set the disk size annotation to schedule this Pod for changes.
+			setDiskSizeAnnotation(pod, requestedDisk)
+		}
+	}
+
+	if changed {
+		rollout.Cascade(vts)
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) claimForTabletPod(ctx context.Context, pod *v1.Pod) (*v1.PersistentVolumeClaim, error) {
+	var claimName string
+	for i := range pod.Spec.Volumes {
+		volume := &pod.Spec.Volumes[i]
+		if volume.PersistentVolumeClaim != nil {
+			claimName = volume.Name
+			break
+		}
+	}
+
+	pvc := &v1.PersistentVolumeClaim{}
+	namespacedClaim := apitypes.NamespacedName{
+		Namespace: pod.Namespace,
+		Name:      claimName,
+	}
+	err := r.client.Get(ctx, namespacedClaim, pvc)
+	if err != nil {
+		return nil, err
+	}
+
+	return pvc, nil
+}
+
+func tabletsForPool(vts *planetscalev2.VitessShard, tabletKeys []string, poolCell string, poolType string) []string {
+	tabletsInCell := make([]string, 0, len(tabletKeys))
+	for _, tabletKey := range tabletKeys {
+		tablet := vts.Status.Tablets[tabletKey]
+		tabletCell := strings.Split(tabletKey, "-")[0]
+
+		if tablet.Type != poolType || tabletCell != poolCell{
+			continue
+		}
+
+		tabletsInCell = append(tabletsInCell, tabletKey)
+	}
+
+	return tabletsInCell
+}
+
+func setDiskSizeAnnotation(pod *v1.Pod, requestedDisk string) {
+	ann := pod.GetAnnotations()
+	if ann == nil {
+		ann = make(map[string]string, 1)
+	}
+	ann[pvcDiskSizeAnnotation] = requestedDisk
+	pod.SetAnnotations(ann)
+}
--- a/pkg/controller/vitessshard/reconcile_disk.go
+++ b/pkg/controller/vitessshard/reconcile_disk.go
@@ -43,7 +43,7 @@ func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetsca
 			continue
 		}
 
-		poolTablets, err := tabletKeysForPool(vts, tabletPool.Cell, string(tabletPool.Type))
+		poolTablets, err := tabletKeysForPool(vts, tabletPool.Cell, tabletPool.Type)
 		if err != nil {
 			return resultBuilder.Error(err)
 		}
@@ -119,7 +119,7 @@ func (r *ReconcileVitessShard) claimForTabletPod(ctx context.Context, pod *v1.Po
 	return pvc, nil
 }
 
-func tabletKeysForPool(vts *planetscalev2.VitessShard, poolCell string, poolType string) ([]string, error) {
+func tabletKeysForPool(vts *planetscalev2.VitessShard, poolCell string, poolType planetscalev2.VitessTabletPoolType) ([]string, error) {
 	tabletKeys := vts.Status.TabletAliases()
 
 	tabletsInCell := make([]string, 0, len(tabletKeys))
@@ -130,7 +130,7 @@ func tabletKeysForPool(vts *planetscalev2.VitessShard, poolCell string, poolType
 			return nil, err
 		}
 
-		if tablet.PoolType != poolType || tabletAlias.Cell != poolCell{
+		if tablet.PoolType != string(poolType) || tabletAlias.Cell != poolCell{
 			continue
 		}
 
--- a/pkg/controller/vitessshard/reconcile_disk.go
+++ b/pkg/controller/vitessshard/reconcile_disk.go
@@ -2,11 +2,11 @@ package vitessshard
 
 import (
 	"context"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	apitypes "k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"vitess.io/vitess/go/vt/topo/topoproto"
 
 	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
 	"planetscale.dev/vitess-operator/pkg/operator/results"
@@ -14,13 +14,12 @@ import (
 )
 
 const (
-	pvcDiskSizeAnnotation = "pvc-disk-size"
+	pvcDiskSizeAnnotation = "planetscale.com/pvc-filesystem-resize"
 )
 
 func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
 	resultBuilder := &results.Builder{}
-	changed := false
-	tabletKeys := tabletKeysFromShard(vts)
+	anythingChanged := false
 	tabletPods, err := r.tabletPodsFromShard(ctx, vts)
 	if err != nil {
 		return resultBuilder.Error(err)
@@ -30,7 +29,10 @@ func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetsca
 		tabletPool := &vts.Spec.TabletPools[i]
 		requestedDiskQuantity := tabletPool.DataVolumeClaimTemplate.Resources.Requests[v1.ResourceStorage]
 		requestedDisk := requestedDiskQuantity.String()
-		poolTablets := tabletsForPool(vts, tabletKeys, tabletPool.Cell, string(tabletPool.Type))
+		poolTablets, err := tabletsForPool(vts, tabletPool.Cell, string(tabletPool.Type))
+		if err != nil {
+			return resultBuilder.Error(err)
+		}
 
 		for _, tabletKey := range poolTablets {
 			pod, ok := tabletPods[tabletKey]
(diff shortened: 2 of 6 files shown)


pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+				continue
+			}
+
+			// If disk size is unchanged, move on to the next tablet.
+			if tabletDiskSize == requestedDisk {
+				continue
+			}
+
+			// If there are disk size changes, mark the shard as changed to cascade later.
+			changed = true
```
enisoc, line 54: This would begin cascading as soon as *any* tablet is ready. We need to wait until *all* tablets (that need restarts) are ready.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+
+			tabletDiskSize, ok := pod.Annotations[pvcDiskSizeAnnotation]
+			if !ok {
+				// If it's never been set before, apply the disk size annotation.
+				setDiskSizeAnnotation(pod, requestedDisk)
+				continue
+			}
+
+			// If disk size is unchanged, move on to the next tablet.
+			if tabletDiskSize == requestedDisk {
```
enisoc, line 49: We can't look at the actual annotation on the Pod to determine whether it's time to cascade. The annotation won't be there until after the Pod is restarted; until then, it is merely a scheduled change.

What we want to check for here is, among all the PVCs whose Status size differs from the desired size configured in the tablet pool:

1. Do all those PVCs have their Spec size set to the new desired size?
2. Do all those PVCs have the status condition `FileSystemResizePending` set to `True`?
3. Do all the corresponding Pods have some kind of update scheduled (`rollout.Scheduled(pod)`)?

If all these are true for all tablet pools with such pending PVCs, and if the number of pending PVCs is greater than zero, then we should trigger a cascade (if one is not already ongoing).

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return resultBuilder.Result()
+}
+
+func (r *ReconcileVitessShard) claimForTabletPod(ctx context.Context, pod *v1.Pod) (*v1.PersistentVolumeClaim, error) {
+	var claimName string
+	for i := range pod.Spec.Volumes {
+		volume := &pod.Spec.Volumes[i]
+		if volume.PersistentVolumeClaim != nil {
```
enisoc, line 84: We shouldn't assume the first PVC listed in the Pod spec is the one we're looking for. Since we are inside the controller that assigns the name for the PVC, we can assume we know the PVC name directly -- it's the same as the Pod name.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+const (
+	pvcDiskSizeAnnotation = "pvc-disk-size"
+)
+
+func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
+	changed := false
+	tabletKeys := tabletKeysFromShard(vts)
```
enisoc, line 23: I don't think tablet order matters here since we don't do anything until all tablets are ready. It would be simpler to just directly iterate vts.Status.Tablets when necessary, rather than passing this ordered list around, if we don't care about order.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return pvc, nil
+}
+
+func tabletsForPool(vts *planetscalev2.VitessShard, tabletKeys []string, poolCell string, poolType string) []string {
+	tabletsInCell := make([]string, 0, len(tabletKeys))
+	for _, tabletKey := range tabletKeys {
+		tablet := vts.Status.Tablets[tabletKey]
+		tabletCell := strings.Split(tabletKey, "-")[0]
```
enisoc, line 107: Use `topoproto.ParseTabletAlias()` instead of trying to parse it ourselves.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,126 @@
...
+
+		if tablet.Type != poolType || tabletCell != poolCell{
```
enisoc, line 109: The `tablet.Type` can change based on dynamic state. We should compare against `tablet.PoolType` which is static.

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Done.

avaidyanatha: As a note, I switched the order of checks 1 and 2 so that the logic with the anythingChanged variable stays the same, let me know if that breaks anything.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (r *ReconcileVitessShard) claimForTabletPod(ctx context.Context, pod *v1.Pod) (*v1.PersistentVolumeClaim, error) {
+	var claimName string
+	for i := range pod.Spec.Volumes {
+		volume := &pod.Spec.Volumes[i]
+		if volume.PersistentVolumeClaim != nil && volume.Name == pod.Name {
+			claimName = volume.Name
+			break
+		}
+	}
```
enisoc, line 92: I don't think we need any of this. We already know the PVC name since it's the same as the Pod name. Also I don't think it's right to look at `volume.Name` since that's the name of the volume within the Pod, while we want the PVC name.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+
+	pvc := &v1.PersistentVolumeClaim{}
+	namespacedClaim := apitypes.NamespacedName{
+		Namespace: pod.Namespace,
+		Name:      claimName,
+	}
+	err := r.client.Get(ctx, namespacedClaim, pvc)
```
enisoc, line 99:
```suggestion
	pvcKey := client.ObjectKey{
		Namespace: pod.Namespace,
		Name:      pod.Name,
	}
	err := r.client.Get(ctx, pvcKey, pvc)
```

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+
+		for _, tabletKey := range poolTablets {
+			pod, ok := tabletPods[tabletKey]
+			if !ok {
+				continue
+			}
+
+			pvc, err := r.claimForTabletPod(ctx, pod)
+			if err != nil {
+				return resultBuilder.Error(err)
```
enisoc, line 45: If the error is NotFound, we should `continue` because if the PVC doesn't exist, it's going to be created with the correct size, so that's fine.

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -360,3 +362,45 @@ func tabletAvailableStatus(resultBuilder *results.Builder, pod *corev1.Pod, read
...
+
+func checkPVCFileSystemResizeCondition (pvc *corev1.PersistentVolumeClaim) bool {
+	for _, condition := range pvc.Status.Conditions {
+		if condition.Type != corev1.PersistentVolumeClaimFileSystemResizePending {
+			continue
+		}
+
+		if condition.Status == corev1.ConditionTrue {
+			return true
+		}
```
enisoc, line 402: We don't need to keep scanning after we find the `condition.Type` we're looking for.

```suggestion
		return condition.Status == corev1.ConditionTrue
```

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			// If we have reached this point in the loop, it indicates that there are disk size changes, so we
+			// set the variable anythingChanged to true. If we successfully complete this loop without bailing,
+			// we can be certain that we have disk size changes and that all required changes have been set.
+			anythingChanged = true
+
+			// If the pod does not have updates scheduled, bail out and wait until the scheduled annotation is applied.
+			if !rollout.Scheduled(pod) {
+				return resultBuilder.Result()
```
enisoc, line 60: Add an event saying we're waiting for the Pod to be scheduled for restart.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			// Finally, if the PVC's disk spec does not equal the new requested disk, bail out.
+			pvcDisk := pvc.Spec.Resources.Requests[v1.ResourceStorage]
+			if pvcDisk.String() != requestedDisk {
+				r.recorder.Eventf(vts, v1.EventTypeNormal, "RolloutCheckPaused", "Waiting for pvc %v to reflect updated disk.", pvc.Name)
```
enisoc, line 66:
```suggestion
				r.recorder.Eventf(vts, v1.EventTypeNormal, "PVCResizeWaiting", "Waiting for PVC %v spec to reflect desired disk size %v.", pvc.Name, requestedDiskQuantity.String())
```

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			pvc, err := r.claimForTabletPod(ctx, pod)
+			if err != nil {
+				return resultBuilder.Error(err)
+			}
+
+			// If the PVC does not have the FileSystemResizeCondition, bail out.
+			if !checkPVCFileSystemResizeCondition(pvc) {
+				return resultBuilder.Result()
```
enisoc, line 50: We don't know yet that there is a resize pending on this PVC. If we bail out for any PVC without the resize condition, we will be blocked if any PVCs are simply not being resized. We need to first skip (`continue`) any PVC whose Status size is already the same as the desired size.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+			// set the variable anythingChanged to true. If we successfully complete this loop without bailing,
+			// we can be certain that we have disk size changes and that all required changes have been set.
+			anythingChanged = true
+
+			// If the pod does not have updates scheduled, bail out and wait until the scheduled annotation is applied.
+			if !rollout.Scheduled(pod) {
+				return resultBuilder.Result()
+			}
+
+			// Finally, if the PVC's disk spec does not equal the new requested disk, bail out.
```
enisoc, line 63: We should check this before we check the resize condition. The condition means "I'm ready to resize", but it might be ready to resize to an old value, which is why we do this additional check.

In general, before we check whether some goal is reached, we should confirm that the goal itself is what we think it should be.

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -360,3 +362,45 @@ func tabletAvailableStatus(resultBuilder *results.Builder, pod *corev1.Pod, read
...
+	// If no PVC is configured for this tablet pod, bail out.
+	if tabletSpec.DataVolumePVCSpec == nil {
+		return
+	}
+
+	pvc := &corev1.PersistentVolumeClaim{}
+	namespacedClaim := apitypes.NamespacedName{
+		Namespace: pod.Namespace,
+		Name:      tabletSpec.DataVolumePVCName,
+	}
```
enisoc, line 376: The type `client.ObjectKey` is an alias of `apitypes.NamespacedName` so we can technically use either one. However, to make the codebase as a whole easier to work with, we picked one (`client.ObjectKey`) and defined a convention that we should stick with it.

```suggestion
	pvcKey := client.ObjectKey{
		Namespace: pod.Namespace,
		Name:      tabletSpec.DataVolumePVCName,
	}
```

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"vitess.io/vitess/go/vt/topo/topoproto"
+
+	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+	"planetscale.dev/vitess-operator/pkg/operator/rollout"
+)
+
+const (
+	pvcDiskSizeAnnotation = "planetscale.com/pvc-filesystem-resize"
```
enisoc, line 17:
```suggestion
	pvcFilesystemResizeAnnotation = "planetscale.com/pvc-filesystem-resize"
```

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -360,3 +362,45 @@ func tabletAvailableStatus(resultBuilder *results.Builder, pod *corev1.Pod, read
...
+		return
+	}
+
+	// If the matching PVC does not have the FileSystemResizePending condition, bail out.
+	if !checkPVCFileSystemResizeCondition(pvc) {
+		return
+	}
+
+	// If all checks pass, set the resize annotation.
+	requestedDiskQuantity := tabletSpec.DataVolumePVCSpec.Resources.Requests[corev1.ResourceStorage]
```
enisoc, line 390: We should also bail if the actual PVC's Spec size has not yet been updated to be equal to the desired size in DataVolumePVCSpec. As mentioned elsewhere, we should bail if DataVolumePVCSpec has a missing ResourceStorage entry as well.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
+	resultBuilder := &results.Builder{}
```
enisoc, line 21: Since the whole purpose of this function is now that we maybe will set cascade, we should check here whether `vts` is already cascading, and bail out since there's no point doing all that work below. This also ensures that we don't keep sending Update attempts to set the cascade annotation when the shard is already cascading.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+			anythingChanged = true
+
+			// If the pod does not have updates scheduled, bail out and wait until the scheduled annotation is applied.
+			if !rollout.Scheduled(pod) {
+				return resultBuilder.Result()
+			}
+
+			// Finally, if the PVC's disk spec does not equal the new requested disk, bail out.
+			pvcDisk := pvc.Spec.Resources.Requests[v1.ResourceStorage]
+			if pvcDisk.String() != requestedDisk {
```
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Oh! Yeah, I'm silly - Done.

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Done, out of curiosity, do we just want to use the more generic object? 
**Edit:** You preempted my question and answered it below, thanks :)

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	err := r.client.Get(ctx, namespacedClaim, pvc)
+	if err != nil {
+		return nil, err
+	}
+
+	return pvc, nil
+}
+
+func tabletsForPool(vts *planetscalev2.VitessShard, poolCell string, poolType string) ([]string, error) {
```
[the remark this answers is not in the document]
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+	tabletKeys := make([]string, 0, len(vts.Status.Tablets))
+	for key := range vts.Status.Tablets {
+		tabletKeys = append(tabletKeys, key)
+	}
```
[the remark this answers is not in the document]
avaidyanatha: Done!

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,129 @@
...
+
+	tabletsInCell := make([]string, 0, len(tabletKeys))
+	for _, tabletKey := range tabletKeys {
+		tablet := vts.Status.Tablets[tabletKey]
+		tabletAlias, err := topoproto.ParseTabletAlias(tabletKey)
+		if err != nil {
+			return tabletKeys, err
```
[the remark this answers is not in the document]
avaidyanatha: Oh, I thought that Go was going to get mad at me for returning nil instead of an array of strings. Done.

pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -105,6 +81,35 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
...
+
+	podList := &v1.PodList{}
+	listOpts := &client.ListOptions{
+		Namespace: vts.Namespace,
+		LabelSelector: apilabels.Set(map[string]string{
+			planetscalev2.ComponentLabel: planetscalev2.VttabletComponentName,
+			planetscalev2.ClusterLabel:   vts.Labels[planetscalev2.ClusterLabel],
+			planetscalev2.KeyspaceLabel:  vts.Labels[planetscalev2.KeyspaceLabel],
+			planetscalev2.ShardLabel:     vts.Spec.KeyRange.SafeName(),
+		}).AsSelector(),
```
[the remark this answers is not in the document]
avaidyanatha: Done!

pkg/controller/vitessshard/reconcile_tablets.go
avaidyanatha: Makes sense, done!

pkg/controller/vitessshard/reconcile_tablets.go
avaidyanatha: Sounds good, done. Going by the logic you gave me before, I moved the check for the `FileSystemResizeCondition` after checking that the PVC Spec size equals the desired size. Let me know if this logic doesn't hold here.

pkg/controller/vitessshard/reconcile_tablets.go
avaidyanatha: Done.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,141 @@
...
+		if tabletPool.DataVolumeClaimTemplate == nil {
+			continue
+		}
+
+		requestedDiskQuantity, ok := tabletPool.DataVolumeClaimTemplate.Resources.Requests[v1.ResourceStorage]
+		if !ok {
+			continue
+		}
+
+		poolTablets, err := tabletKeysForPool(vts, tabletPool.Cell, string(tabletPool.Type))
```
PrismaPhonic, line 46: Why not just have this function take the tablet pool type, rather than forcing the caller to use type casting?

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -360,3 +361,54 @@ func tabletAvailableStatus(resultBuilder *results.Builder, pod *corev1.Pod, read
 	resultBuilder.RequeueAfter(tabletAvailableTime - readyDuration)
 	return corev1.ConditionFalse
 }
+
+func (r *ReconcileVitessShard) checkPVCDiskUpdates(ctx context.Context, tabletSpec *vttablet.Spec, pod *corev1.Pod) {
```
PrismaPhonic, line 365: It's not obvious to me as a reader, based on this function name, that this functions purpose is either to bail, or mutate the tablet specs annotations. I would prefer a name that clearly communicates mutation is going to happen. Suggestion:
```suggestion
func (r *ReconcileVitessShard) updatePvcFilesystemResizeAnnotation(ctx context.Context, tabletSpec *vttablet.Spec, pod *corev1.Pod) {
```

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Yeah good point. Fixed it to take the tablet pool type.

pkg/controller/vitessshard/reconcile_tablets.go
avaidyanatha: Done, changed it to `UpdatePVCFilesystemResizeAnnotation`.

pkg/apis/planetscale/v2/vitessshard_methods.go
```diff
@@ -216,3 +217,13 @@ func (s *VitessShardStatus) DeepCopyConditions() map[VitessShardConditionType]*V
 
 	return out
 }
+
+func (s *VitessShardStatus) TabletAliases() []string {
```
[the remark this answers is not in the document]
enisoc: My IDE (VS Code) underlines go vet issues like spellcheck.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,141 @@
...
+			// If the PVC's disk spec does not equal the new requested disk, bail out.
+			pvcDisk := pvc.Spec.Resources.Requests[v1.ResourceStorage]
+			if pvcDisk.Value() != requestedDiskQuantity.Value() {
+				r.recorder.Eventf(vts, v1.EventTypeNormal, "PVCResizeWaiting", "Waiting for PVC %v spec to reflect desired disk size %v.", pvc.Name, requestedDiskQuantity.String())
+				return resultBuilder.Result()
+			}
+
+			// If the PVC does not have the FileSystemResizeCondition, bail out.
+			if !checkPVCFileSystemResizeCondition(pvc) {
+				r.recorder.Eventf(vts, v1.EventTypeNormal, "PVCResizeWaiting", "Waiting for pvc %v to reflect the filesystem resize condition.", pvc.Name)
```
enisoc, line 84:
```suggestion
				r.recorder.Eventf(vts, v1.EventTypeNormal, "PVCResizeWaiting", "Waiting for PVC %v to be ready for filesystem resize.", pvc.Name)
```

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -0,0 +1,141 @@
+package vitessshard
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
avaidyanatha: Updated!

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Done.

PR #84, opened 2020-05-05
Apply Annotations Instantly

PrismaPhonic:
We discovered that for all of our use cases, we always want annotations to apply instantly, so we decided to reverse the current behavior. This PR ensures that annotations that were previously held behind updates issued by an external actor, are always applied immediately.

### DIFF
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -156,6 +156,8 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vts.Labels, newShard.Labels)
 
+	// Remove old annotations that shouldn't be there that we injected previously.
+	// This must be done before we update vts.Spec.
 	updateVitessShardAnnotations(vts, newShard)
 
 	// For now, everything in Spec is safe to update.
@@ -168,12 +170,11 @@ func updateVitessShardInPlace(key client.ObjectKey, vts *planetscalev2.VitessSha
 	// For now, only disk size & annotations are safe to update in place.
 	update.ShardDiskSize(vts.Spec.TabletPools, newShard.Spec.TabletPools)
 
+	// Remove old annotations that shouldn't be there that we injected previously.
 	updateVitessShardAnnotations(vts, newShard)
 }
 
 func updateVitessShardAnnotations(vts *planetscalev2.VitessShard, newShard *planetscalev2.VitessShard) {
-	// Remove old annotations that shouldn't be there that we injected previously.
-	// This must be done before we update vts.Spec.
 	differentAnnotations := differentKeys(vts.Spec.Annotations, newShard.Spec.Annotations)
 	for _, annotation := range differentAnnotations {
 		delete(vts.Annotations, annotation)
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -182,6 +182,8 @@ func updateVitessShardAnnotations(vts *planetscalev2.VitessShard, newShard *plan
 
 	// Update annotations we set.
 	update.Annotations(&vts.Annotations, newShard.Annotations)
+
+	vts.Spec.Annotations = newShard.Spec.Annotations
 }
 
 // differentKeys returns keys from an older map instance that are no longer in a newer map instance.
--- a/pkg/controller/vitesskeyspace/reconcile_shards.go
+++ b/pkg/controller/vitesskeyspace/reconcile_shards.go
@@ -162,6 +162,15 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	vts.Spec = newShard.Spec
 }
 
+func updateVitessShardInPlace(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk *planetscalev2.VitessKeyspace, parentLabels map[string]string, shard *planetscalev2.VitessKeyspaceKeyRangeShard) {
+	newShard := newVitessShard(key, vtk, parentLabels, shard)
+
+	// For now, only disk size & annotations are safe to update in place.
+	update.ShardDiskSize(vts.Spec.TabletPools, newShard.Spec.TabletPools)
+
+	updateVitessShardAnnotations(vts, newShard)
+}
+
 func updateVitessShardAnnotations(vts *planetscalev2.VitessShard, newShard *planetscalev2.VitessShard) {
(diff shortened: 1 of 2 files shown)


pkg/operator/vttablet/pod.go
```diff
@@ -338,6 +333,18 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	obj.Spec.PriorityClassName = vttabletPriorityClassName
 }
 
+func updatePodAnnotations(obj *corev1.Pod, spec *Spec) {
```
avaidyanatha, line 336: This is fine, but for some reason I sorta like the idea of forgoing DRY here and distinctly expressing everything inside the strategy handler functions. Do we ever foresee annotations diverging between strategies after this?

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -156,6 +156,13 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vts.Labels, newShard.Labels)
 
+	updateVitessShardAnnotations(vts, newShard)
+
+	// For now, everything in Spec is safe to update.
+	vts.Spec = newShard.Spec
+}
+
+func updateVitessShardAnnotations(vts *planetscalev2.VitessShard, newShard *planetscalev2.VitessShard) {
```
avaidyanatha, line 165: Nit: move this below `updateVitessShardInPlace` since it's a helper function.

pkg/operator/vttablet/pod.go
```diff
@@ -61,6 +61,8 @@ func NewPod(key client.ObjectKey, spec *Spec) *corev1.Pod {
 func UpdatePodInPlace(obj *corev1.Pod, spec *Spec) {
```
avaidyanatha, line 61: What do you think about forgoing DRY here and not creating the `updatePodAnnotations` subroutine? I liked the explicitness of the strategy handlers before, but it's mostly up to opinion.

pkg/operator/vttablet/pod.go
```diff
@@ -61,6 +61,8 @@ func NewPod(key client.ObjectKey, spec *Spec) *corev1.Pod {
...
 	// Update labels and annotations, but ignore existing ones we don't set.
 	update.Labels(&obj.Labels, spec.Labels)
+
+	updatePodAnnotations(obj, spec)
```
enisoc, line 65: I didn't think we talked about changing the behavior of other annotations besides those set through VitessShardSpec.Annotations. The use case we have for VitessShardSpec.Annotations doesn't apply here, so we'd have to separately justify this change and verify that it's safe.

PrismaPhonic: Alright, that's a good point. I can revert the pod annotation changes.

pkg/operator/vttablet/pod.go
PrismaPhonic: It sounds like I'll be removing updatePodAnnotations entirely and reverting back because we only want to apply annotations instantly for non-pod annotations.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Good idea.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -165,16 +172,15 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
...
 }
 
 func updateVitessShardInPlace(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk *planetscalev2.VitessKeyspace, parentLabels map[string]string, shard *planetscalev2.VitessKeyspaceKeyRangeShard) {
 	newShard := newVitessShard(key, vtk, parentLabels, shard)
 
-	// For now, only disk size is safe to update in place.
+	// For now, only disk size & annotations are safe to update in place.
 	update.ShardDiskSize(vts.Spec.TabletPools, newShard.Spec.TabletPools)
+
+	updateVitessShardAnnotations(vts, newShard)
```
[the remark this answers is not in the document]
PrismaPhonic: Is this something we missed before? I only moved existing logic into a helper function that could be called by both `Update` and `UpdateInPlace`.

pkg/operator/vttablet/pod.go
PrismaPhonic: It's been removed entirely, so resolving this conversation.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -156,6 +156,13 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vts.Labels, newShard.Labels)
 
+	updateVitessShardAnnotations(vts, newShard)
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/operator/vttablet/pod.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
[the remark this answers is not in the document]
PrismaPhonic: Or was this taken care of by overwriting `vts.Spec` entirely with the new shard spec?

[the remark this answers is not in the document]
PrismaPhonic: Regardless I've updated the code to overwrite this in the helper. LMK if this is what you meant so I know we're on the same page. 
```
vts.Spec.Annotations = newShard.Spec.Annotations
```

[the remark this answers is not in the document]
enisoc: Yeah it used to be taken care of by overwriting all of Spec.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -158,13 +158,7 @@ func updateVitessShard(key client.ObjectKey, vts *planetscalev2.VitessShard, vtk
 
 	// Remove old annotations that shouldn't be there that we injected previously.
```
enisoc, line 159:
```suggestion
	// Add or remove annotations requested in vts.Spec.Annotations.
```

PrismaPhonic: Done.

PR #89, opened 2020-05-07
Allow custom database name override.

enisoc:
Add a field to let the user customize the name of the physical MySQL database that Vitess creates.

### DIFF
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -96,6 +96,22 @@ type VitessKeyspaceTemplate struct {
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([a-z0-9]*[a-z0-9])?$
 	Name string `json:"name"`
 
+	// DatabaseName is the name to use for the underlying, physical MySQL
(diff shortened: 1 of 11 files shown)


pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -96,6 +96,22 @@ type VitessKeyspaceTemplate struct {
...
+	// only the keyspace name as a logical database. However, you may want to
+	// set this to control the name used by clients that bypass Vitess and
+	// connect directly to the underlying MySQL, such as certain DBA tools.
+	//
+	// The default is to add a "vt_" prefix to the keyspace name since that has
+	// historically been the default in Vitess itself. However, it's often
+	// preferable to set this to be the same as the keyspace name to reduce
+	// confusion.
+	//
+	// Default: Add a "vt_" prefix to the keyspace name.
```
teejae, line 112: can you mention that the default is when the string is empty somehow? if it was a pointer, i don't think it would be a problem, but this isn't a pointer optional

enisoc: Done.

PR #91, opened 2020-05-15
Treat externalmaster as master-eligible during rollout.

enisoc:
During VitessShard cascading rollout, we restart the master in place if there are no other master-eligible tablets to reparent to anyway. We should do this for externalmaster tablets too, since the same
reasoning applies.

### DIFF
--- a/pkg/controller/vitessshard/reconcile_rollout.go
+++ b/pkg/controller/vitessshard/reconcile_rollout.go
@@ -3,7 +3,7 @@ package vitessshard
 import (
 	"context"
 
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	apilabels "k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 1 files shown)


pkg/controller/vitessshard/reconcile_rollout.go
```diff
@@ -56,35 +56,36 @@ func (r *ReconcileVitessShard) reconcileRollout(ctx context.Context, vts *planet
...
+		r.recorder.Eventf(vts, corev1.EventTypeNormal, "RollingRestartComplete", "Cascading rollout of tablets is complete.")
 		return resultBuilder.Result()
 	}
 
 	masterEligibleTablets := vts.Spec.MasterEligibleTabletCount()
 	deletePod := false
 	tabletType := pod.Labels[planetscalev2.TabletTypeLabel]
 	// These two conditions guarantee that the tablet is a lone master.
-	if masterEligibleTablets == 1 && tabletType == string(planetscalev2.ReplicaPoolType) {
+	if masterEligibleTablets < 2 &&
```
avaidyanatha, line 71: I assume the 0 case never happens or that we don't disambiguate it?

enisoc: Originally I had written it as `masterEligibleTablets == 1` because I thought 0 couldn't happen. After all, how could the total be 0 if we are looking at a master-eligible Pod right now?

I later realized, `masterEligibleTablets` tells us the desired number of tablets, whereas we might be looking at a Pod that hasn't been reconciled yet. So I decided we should handle the `0` case too.

PR #92, opened 2020-05-15
Switch to the new TabletExternallyReparented.

enisoc:
Upstream Vitess moved TER from TabletManagerClient to Wrangler. We stayed on the old one for a while because the new one requires that users have upgraded to a certain minimum Vitess version. It has been long enough now that I think we can assume that's the case.

### DIFF
--- a/pkg/controller/vitessshardreplication/reconcile_drain.go
+++ b/pkg/controller/vitessshardreplication/reconcile_drain.go
@@ -317,12 +317,8 @@ func (r *ReconcileVitessShard) handleExternalReparent(ctx context.Context, vts *
 	err := wr.TabletExternallyReparented(ctx, newMasterAlias)
 
 	if err == nil {
-		// TODO: Create a specialized function to take the place of TER, that sets the old master as spare, rather than
-		// replica.  Currently TER sets the old master as type replica, and it does so asyncronously.  This may cause
-		// a race where we are left with the old master being of type replica, even though we try to explicitely set it to
-		// be of type spare.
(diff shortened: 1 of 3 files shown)


pkg/controller/vitessshardreplication/reconcile_drain.go
```diff
@@ -313,11 +313,10 @@ func (r *ReconcileVitessShard) reconcileDrain(ctx context.Context, vts *planetsc
...
 }
 
-func (r *ReconcileVitessShard) handleExternalReparent(ctx context.Context, vts *planetscalev2.VitessShard, wr *wrangler.Wrangler, newMasterTablet *topodatapb.Tablet, oldMasterAlias *topodatapb.TabletAlias) error {
-	err := wr.TabletManagerClient().TabletExternallyReparented(ctx, newMasterTablet, "")
+func (r *ReconcileVitessShard) handleExternalReparent(ctx context.Context, vts *planetscalev2.VitessShard, wr *wrangler.Wrangler, newMasterAlias, oldMasterAlias *topodatapb.TabletAlias) error {
+	err := wr.TabletExternallyReparented(ctx, newMasterAlias)
 
 	if err == nil {
-
 		// TODO: Create a specialized function to take the place of TER, that sets the old master as spare, rather than
```
deepthi, line 320: You could use the `demote_master_type=SPARE` vttablet flag to fix this TODO.

enisoc: Ooh nice! Thanks!

PR #94, opened 2020-05-19
Add cluster level user toggle for choosing disk update strategy

avaidyanatha:
## Main Changes
- We want users to be able to choose between immediate and externally managed disk updates, as some may want finer control over when tablet pods and PVCs are restarted


pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -135,6 +135,23 @@ type VitessClusterUpdateStrategy struct {
...
+	//   volume claims but wait for an external tool to release them by adding
+	//   the 'rollout.planetscale.com/released' annotation.
+	// - Immediate: Release updates to all tablet pools and persistent volume
+	//   claims as soon as the VitessCluster spec is changed. Perform rolling
+	//   restart of one tablet Pod per shard at a time, with automatic planned
+	//   reparents whenever possible to avoid master downtime.
+	//
+	// Default: External
+	// +kubebuilder:validation:Enum=External,Immediate
+	DataVolumeClaimResize *DataVolumeClaimResizeType `json:"dataVolumeClaimResize,omitempty"`
```
PrismaPhonic, line 154: My feeling is that this shouldn't be separate from the `type` field. If we keep following this pattern, we'll have more toggles than a user knows what to do with. I think it should just all get pushed into `type`.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -91,6 +91,9 @@ type VitessShardSpec struct {
 
 	// TopologyReconciliation is inherited from the parent's VitessClusterSpec.
 	TopologyReconciliation *TopoReconcileConfig `json:"topologyReconciliation,omitempty"`
+
+	// UpdateStrategy is inherited from the parent's VitessClusterSpec.
+	UpdateStrategy *VitessClusterUpdateStrategy `json:"updateStrategy,omitempty"`
```
enisoc, line 96: We need to call `DefaultUpdateStrategy()` from `DefaultVitessShard()` now, so VitessShard controller sees the right, non-nil defaults before VitessKeyspace controller gets around to filling in the new field.

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -135,6 +135,23 @@ type VitessClusterUpdateStrategy struct {
...
+	//
+	// - External: Schedule disk size updates on tablet pools and persistent
+	//   volume claims but wait for an external tool to release them by adding
+	//   the 'rollout.planetscale.com/released' annotation.
+	// - Immediate: Release updates to all tablet pools and persistent volume
+	//   claims as soon as the VitessCluster spec is changed. Perform rolling
+	//   restart of one tablet Pod per shard at a time, with automatic planned
+	//   reparents whenever possible to avoid master downtime.
+	//
+	// Default: External
```
enisoc, line 152: Need to make this actually happen in `DefaultUpdateStrategy()`.

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -21,6 +21,11 @@ const (
 func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetscalev2.VitessShard) (reconcile.Result, error) {
 	resultBuilder := &results.Builder{}
 
+	// If the user has specified their disk resizes to be handled externally, wait for a manual rollout to apply changes.
+	if *vts.Spec.UpdateStrategy.DataVolumeClaimResize == planetscalev2.ExternalDataVolumeClaimResizeType {
```
enisoc, line 25: Let's bail if it's anything but Immediate. We don't know what other types may be added later, so we should be conservative in when we enable this.

Also, if the main UpdateStrategy.Type is Immediate, we should go ahead and do this without even checking DataVolumeClaimResize. That way setting the overall type to Immediate really makes everything immediate.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: Done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: I agree overall too, I was thinking about this... but the interest on this tech debt doesn't seem too bad for the time being, so I was thinking we approach this in a team meeting.

pkg/apis/planetscale/v2/vitessshard_types.go
avaidyanatha: Done!

pkg/controller/vitessshard/reconcile_disk.go
avaidyanatha: Sounds good, done.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: My sense is that this is the wrong place to put this if it's an override. Essentially the types are no different at all (it's either immediate or external). I would prefer it being it's own field and just using the existing struct in its entirety. so rather than on `updateStrategy` field, it could be `dataVolumeUpdateStrategy` field. To me this reads as an odd duplication when the it fundamentally represents the exact same two options.

PrismaPhonic: Or better yet, have the update strategy live on a child where it's the most relevant to data volumes.

enisoc: > being it's own field and just using the existing struct in its entirety

I was envisioning that the `VitessClusterUpdateStrategy` would eventually have more settings specific to the scope of a Vitess Cluster (like how to update keyspaces in particular), so I don't think we should share the whole struct. It might make sense to share the `VitessClusterUpdateStrategyType` type, but even then I feel like we may need values there that only make sense in certain scopes, like the scope of a whole VitessCluster.

> so rather than on updateStrategy field, it could be dataVolumeUpdateStrategy field

The trade-off there is that if you extend this pattern for other things we update (e.g. keyspaces), the top-level VitessCluster.Spec starts to feel polluted with similar fields (`*UpdateStrategy`) that should instead be grouped:

```yaml
# everything under this tree is assumed to be about update strategy,
# so field names can be cleaner and related concepts are grouped.
updateStrategy:
  # data volume claim update strategy
  dataVolumeClaim: blah
  # keyspace update strategy
  keyspace: blah
  # cell update strategy
  cell: blah
```

This is what we were going for with `dataVolumeClaimResize` but maybe we should generalize it to `dataVolumeClaim` and make it about any updates to the data volume PVC?

> have the update strategy live on a child where it's the most relevant to data volumes

If we were talking about overriding the strategy for a whole subtree of our hierarchy like a shard or a keyspace, then I'd agree with embedding another copy of the struct in the root of the subtree. However, the set of all data volume claims in the cluster is not a subtree; it's a bunch of leaves from all over the tree. What we want here is to affect all PVCs in the cluster, so VitessCluster spec is already the root of the subtree we want to affect.

I do agree overall that this API doesn't feel right, so it's worth spending a bit of time to see if we can come up with something better.

For context, the `updateStrategy` field was meant to mimic the pattern used by StatefulSet's `updateStrategy` and Deployment's `strategy` so users would find it familiar. In that pattern, the `updateStrategy` field is a struct with not only an overall `type` but also subfields to customize settings for the selected type:

```yaml
updateStrategy:
  type: RollingUpdate
  rollingUpdate:
    # when rolling update is enabled, customize how exactly to roll
    foo: bar
```

Perhaps we could double down on copying the k8s API and think of this new field we're adding as a customization of the External strategy type:

```yaml
updateStrategy:
  type: External
  external:
    # Even though our overall strategy type is External,
    # we have customizations to let certain things through.
    allowResourceChanges:
    - storage
    # later we would support these too:
    - cpu
    - memory
```

avaidyanatha: I'm a fan of copying the k8s API, I think it's a nice solution that incorporates Peter's suggestions and gets across the idea that choosing immediate on resources is more like a bypass that doesn't live independently of the `external` option.

pkg/controller/vitesscluster/reconcile_keyspaces.go
```diff
@@ -198,8 +198,10 @@ func updateVitessKeyspaceInPlace(key client.ObjectKey, vtk *planetscalev2.Vitess
 	// Update labels, but ignore existing ones we don't set.
 	update.Labels(&vtk.Labels, newKeyspace.Labels)
 
-	// Update disk size.
-	update.KeyspaceDiskSize(&vtk.Spec.VitessKeyspaceTemplate, &newKeyspace.Spec.VitessKeyspaceTemplate)
+	// Update disk size immediately if specified to.
```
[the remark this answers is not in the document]
avaidyanatha: Done!

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: I like this suggestion - I think it reads much better. I'm on board.

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -135,6 +135,10 @@ type VitessClusterUpdateStrategy struct {
 	// Default: External
 	// +kubebuilder:validation:Enum=External,Immediate
 	Type *VitessClusterUpdateStrategyType `json:"type,omitempty"`
+
+	// ExternalOptions can optionally be used to enable the user to customize their external update strategy
+	// to allow certain updates to pass through immediately without using an external tool.
+	ExternalOptions *ExternalVitessClusterUpdateStrategyOptions `json:"external,omitempty"`
```
enisoc, line 141: The Go field name needs to match the JSON field name. Did `make generate` not complain about this?

```suggestion
	// External can optionally be used to customize the external update strategy.
	External *ExternalVitessClusterUpdateStrategyOptions `json:"external,omitempty"`
```

pkg/apis/planetscale/v2/vitesscluster_methods.go
```diff
@@ -69,3 +69,13 @@ func (image *MysqldImage) Flavor() string {
 		return ""
 	}
 }
+
+func (externalOptions *ExternalVitessClusterUpdateStrategyOptions) Storage() bool {
```
enisoc, line 73: The options struct isn't all about resource changes, so we should add some context:

```suggestion
func (externalOptions *ExternalVitessClusterUpdateStrategyOptions) AllowStorageChanges() bool {
```

Actually maybe we should keep this general so we don't end up duplicating it with slightly different search terms:

```suggestion
func (externalOptions *ExternalVitessClusterUpdateStrategyOptions) ResourceChangesAllowed(resource corev1.ResourceName) bool {
```

pkg/apis/planetscale/v2/vitesscluster_methods.go
```diff
@@ -69,3 +69,13 @@ func (image *MysqldImage) Flavor() string {
...
+	for _, resource := range externalOptions.AllowResourceChanges {
+		if resource == "storage" {
+			return true
+			}
```
enisoc, line 77: Run gofmt/goimports?

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -148,6 +152,17 @@ const (
 	ImmediateVitessClusterUpdateStrategyType VitessClusterUpdateStrategyType = "Immediate"
 )
 
+type ExternalVitessClusterUpdateStrategyOptions struct {
+	// AllowResourceChanges can be used to allow certain resource updates to the DataVolumeClaimTemplate
+	// to be updated immediately without using an external tool.
```
enisoc, line 157: This field is meant to be about resources in general, since we plan to support cpu/mem later, so it isn't specific to the data volume.

```suggestion
	// AllowResourceChanges can be used to allow changes to certain resource
	// requests and limits to propagate immediately, bypassing the external rollout tool.
```

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -148,6 +152,17 @@ const (
...
+	//
+	// Supported options:
+	// - storage
+	//
+	// Default: Update no fields immediately.
```
enisoc, line 162:
```suggestion
	// Default: All resource changes wait to be released by the external rollout tool.
```

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -148,6 +152,17 @@ const (
...
+	AllowResourceChanges []string
```
enisoc, line 163: Let's use the official "enum" type for resources since we plan to match this up against the keys in a `corev1.ResourceList`. Also, all API fields need JSON tags.

```suggestion
	AllowResourceChanges []corev1.ResourceName `json:"allowResourceChanges,omitempty"`
```

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -148,6 +152,17 @@ const (
 	ImmediateVitessClusterUpdateStrategyType VitessClusterUpdateStrategyType = "Immediate"
 )
 
+type ExternalVitessClusterUpdateStrategyOptions struct {
```
enisoc, line 155: We should add defaulting for this new sub-struct in `DefaultUpdateStrategy()`, whenever the `type` is `external`, so we can safely assume this is non-nil after checking `type`.

pkg/controller/vitesscluster/reconcile_keyspaces.go
```diff
@@ -198,9 +198,18 @@ func updateVitessKeyspaceInPlace(key client.ObjectKey, vtk *planetscalev2.Vitess
...
 	update.Labels(&vtk.Labels, newKeyspace.Labels)
 
-	// Update disk size.
-	update.KeyspaceDiskSize(&vtk.Spec.VitessKeyspaceTemplate, &newKeyspace.Spec.VitessKeyspaceTemplate)
+	// Switching update strategies should always take effect immediately.
+	vtk.Spec.UpdateStrategy = newKeyspace.Spec.UpdateStrategy
+
+	// Update disk size immediately if specified to.
+	if *vtk.Spec.UpdateStrategy.Type == planetscalev2.ExternalVitessClusterUpdateStrategyType {
+		if vtk.Spec.UpdateStrategy.ExternalOptions != nil {
```
enisoc, line 206: We can get rid of this after adding defaulting.

pkg/apis/planetscale/v2/vitesscluster_methods.go
avaidyanatha: I had tons of issues with naming in these API changes so thanks for offering suggestions; I was very stuck. Also, changing this method to take in any resource is solid, done.

pkg/apis/planetscale/v2/vitesscluster_methods.go
avaidyanatha: Apologies, apparently I only enabled my gofmt save hook on a different project. Done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: `make generate` did not seem to complain, but I've changed it anyway. Done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: Yep, done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: Done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: Done.

pkg/apis/planetscale/v2/vitesscluster_types.go
avaidyanatha: Done.

pkg/controller/vitesscluster/reconcile_keyspaces.go
avaidyanatha: Makes sense, done!

Issue #85, opened 2020-05-05
Add setting to disable replication repair


enisoc:
When running something like Orchestrator alongside Vitess, it's best to disable the replication repair feature of Vitess Operator so only one agent tries to fix the master. We would need to add a new field under [replication](https://vitess-operator.planetscale.dev/api/#planetscale.com/v2.VitessReplicationSpec) to support this, since replication repair is currently always on.

luisfmcalado:
I was able to test without the orchestrator that it keeps the master as read-only when the repair is disabled. I'm having some issues with Vitess + orchestrator configuration so I wasn't able to test if the race condition issue is fixed.

luisfmcalado:
I've tested with vitess + orchestrator (mysql flavor) and a new master was elected without any issues.

PR #95, opened 2020-05-23
Add option to disable replication repair

luisfmcalado:
When using the operator with orchestrator the repair replication should be disabled to avoid any race condition with multiple agents fixing the same master.
The following PR adds the replication option for disabling the operator replication repair.

```yaml
replication:
   recoverRestartedMaster: false
```

Closes: #85

### DIFF
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -152,6 +152,12 @@ type VitessReplicationSpec struct {
 	//
 	// Default: true.
 	InitializeBackup *bool `json:"initializeBackup,omitempty"`
+
+	// ReplicationRepair specifies whether the replication repair should be executed
+	// when replica tablets have broken replication.
+	//
+	// Default: true.
+	ReplicationRepair *bool `json:"replicationRepair,omitempty"`
 }
 
 // VitessShardTabletPool defines a pool of tablets with a similar purpose.
--- a/pkg/apis/planetscale/v2/vitessshard_types.go
+++ b/pkg/apis/planetscale/v2/vitessshard_types.go
@@ -153,11 +153,12 @@ type VitessReplicationSpec struct {
 	// Default: true.
 	InitializeBackup *bool `json:"initializeBackup,omitempty"`
 
-	// ReplicationRepair specifies whether the replication repair should be executed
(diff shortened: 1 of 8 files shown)


pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -152,6 +152,12 @@ type VitessReplicationSpec struct {
 	//
 	// Default: true.
 	InitializeBackup *bool `json:"initializeBackup,omitempty"`
+
+	// ReplicationRepair specifies whether the replication repair should be executed
+	// when replica tablets have broken replication.
+	//
+	// Default: true.
+	ReplicationRepair *bool `json:"replicationRepair,omitempty"`
```
enisoc, line 160: So far, we've tried to keep the field names in VitessReplicationSpec as verb phrases, just for consistency. I'd also like to be more specific about what behavior this controls. Here's my suggestion, based on what the operator actually does when this is enabled:

```suggestion
	// RecoverRestartedMaster specifies whether the operator attempts to repair
	// replication when the master MySQL restarts in-place (due to a crash) or
	// its Pod gets deleted and recreated, causing the Pod IP to change.
	//
	// Default: true.
	RecoverRestartedMaster *bool `json:"recoverRestartedMaster,omitempty"`
```

pkg/controller/vitessshardreplication/repair_replication.go
```diff
@@ -81,6 +81,12 @@ func (r *ReconcileVitessShard) repairReplication(ctx context.Context, vts *plane
 
 	// If using external datastore then replication is handled for us.
 	if vts.Spec.UsingExternalDatastore() {
+	 	return resultBuilder.Result()
+	}
+
+	// If we have configured the operator to ignore replication repair, bail.
+	if !*vts.Spec.Replication.ReplicationRepair {
+		log.Warningf("skip replication repair for master tablet %v", vts.Status.MasterAlias)
```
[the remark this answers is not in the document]
luisfmcalado: yes, I will remove the log statement.

luisfmcalado:
@enisoc don't know what's the issue with netlify checks

pkg/apis/planetscale/v2/vitessshard_types.go
deepthi: +1

PR #102, opened 2020-06-05
When static auth is enabled, enable CLI flag to reload it periodically, and stop bouncing vtgate when the static auth secret is changed

mcronce:


pkg/apis/planetscale/v2/vitesscell_methods.go
```diff
@@ -25,10 +25,6 @@ import (
 func (s *VitessCellGatewaySpec) SecretNames() sets.String {
```
enisoc, line 25: Originally we were thinking this would return all secrets, but now we're changing the meaning to only those secrets that should trigger rolling restart. Do you mind renaming the method to something like `ReloadSecretNames`, removing the TODO, and changing the comment to say `ReloadSecretNames returns the names of gateway Secrets that have to be forcefully reloaded by restarting all vtgates.`?

mcronce: Done

PR #105, opened 2020-06-29
Bumping pinned vitess version

dkhenry:
Bumping Pinned Vitess Verison


pkg/operator/vitesstopo/prune_cells.go
```diff
@@ -82,7 +82,7 @@ func DeleteCells(ctx context.Context, ts *topo.Server, recorder record.EventReco
 
 	for _, cellName := range cellNames {
 		// topo.NoNode is the error type returned if we can't find the cell when deleting. This ensures that this operation is idempotent.
-		if err := ts.DeleteCellInfo(ctx, cellName); err != nil && !topo.IsErrType(err, topo.NoNode) {
+		if err := ts.DeleteCellInfo(ctx, cellName, true); err != nil && !topo.IsErrType(err, topo.NoNode) {
```
deepthi, line 85:
```suggestion
		if err := ts.DeleteCellInfo(ctx, cellName, true /*force*/); err != nil && !topo.IsErrType(err, topo.NoNode) {
```

pkg/controller/vitessshardreplication/repair_replication.go
```diff
@@ -289,8 +289,7 @@ func (r *ReconcileVitessShard) repairReplicationLocked(ctx context.Context, vts
 			if err != nil {
 				return
 			}
-			if status.MasterHost == topoproto.MysqlHostname(masterTabletInfo.Tablet) &&
-				status.MasterPort == topoproto.MysqlPort(masterTabletInfo.Tablet) {
+			if netutil.JoinHostPort(status.MasterHost, status.MasterPort) != topoproto.MysqlAddr(masterTabletInfo.Tablet) {
```
enisoc, line 292: This inverted the logic.

pkg/operator/vitesstopo/prune_cells.go
enisoc: We should leave this as `false` because that was the behavior before the new argument was added.

PR #109, opened 2020-07-17
Allow users to override certain Service fields.

enisoc:
This starts to allow some customization of the Service objects that the operator manages.


pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -390,6 +398,23 @@ type VitessDashboardSpec struct {
...
+
+	// Service can optionally be used to customize the vtctld Service.
+	Service *ServiceOverrides `json:"service,omitempty"`
+}
+
+// ServiceOverrides allows customization of an arbitrary Service object.
+type ServiceOverrides struct {
+	// Annotations specifies extra annotations to add to the Service object.
+	// Annotations added in this way will NOT be automatically removed from the
+	// Service object if they are removed here.
```
teejae, line 410: i presume we always only want additive annotations right?

PR #115, opened 2020-08-08
Report status for resharding workflows

avaidyanatha:
## Main Changes
- In order for the operator to act on in-progress resharding processes, we must retrieve the information about the current workflows from the tablet manager client. These new fields on `VitessKeyspaceStatus` indicate whether or not resharding is currently happening and stores the list of current workflows in progress.


pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -198,3 +220,18 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
 	reconcileCount.WithLabelValues(vtk.Labels[planetscalev2.ClusterLabel], vtk.Spec.Name, metrics.Result(err)).Inc()
 	return result, err
 }
+
+// newWrangler initializes a new Vitess Wrangler that gives us access to information
+// about resharding workflows.
+func newWrangler(ctx context.Context, vtk planetscalev2.VitessKeyspace) (*wrangler.Wrangler, error) {
```
PrismaPhonic, line 226: If we don't pass a pointer for the vtk, it will copy it, and it's a large object: 

```suggestion
func newWrangler(ctx context.Context, vtk *planetscalev2.VitessKeyspace) (*wrangler.Wrangler, error) {
```

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -179,6 +184,23 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
 	topoResult, err := r.reconcileTopology(ctx, vtk)
 	resultBuilder.Merge(topoResult, err)
 
+	wr, err := newWrangler(ctx, *vtk)
+	if err != nil {
+		r.recorder.Eventf(vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to retrieve resharding information: %v", err)
+	}
+	if wr != nil {
```
PrismaPhonic, line 191: Should we hit this? I think we should guarantee that if err == nil that wr will be valid.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -179,6 +184,23 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
+		workflows, err := wr.ListAllWorkflows(ctx, vtk.Name)
+		if err != nil {
+			r.recorder.Eventf(vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to list active resharding workflows: %v", err)
+		}
+		if len(workflows) != 0 {
+			vtk.Status.ReshardingInProgress = true
+			vtk.Status.ActiveWorkflows = workflows
+		} else {
+			vtk.Status.ReshardingInProgress = false
+		}
```
PrismaPhonic, line 201: Cleanliness suggestion:
```suggestion
		vtk.Status.ReshardingInProgress = len(workflows) != 0
		vtk.Status.ActiveWorkflows = workflows
```

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
PrismaPhonic: Also, I think it would be better to scope the inputs to exactly what we need rather than passing the entire vtk, so scratch my last comment. 
```suggestion
func newWrangler(ctx context.Context, lockserverSpec planetscalev2.LockserverSpec) (*wrangler.Wrangler, error) {
```

avaidyanatha: I think you mean `planetscalev2.VitessLockserverParams` right? But that makes sense - done.

PrismaPhonic: Yep

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
avaidyanatha: Was an artifact of a previous version of the code when it lived in a different file, it's gone now!

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
avaidyanatha: Nice suggestion, done!

PrismaPhonic:
@enisoc, I took this over because @avaidyanatha has too much other work on his plate right now. I've tested this locally and it works for `MoveTables`. There's a bug in `VExec` that makes Workflow commands not work for horizontal resharding that @rohit-nayak-ps is working on a fix for. Once that fix is in place the `Reshard` command should also trigger these status updates.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -198,3 +216,18 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
 	reconcileCount.WithLabelValues(vtk.Labels[planetscalev2.ClusterLabel], vtk.Spec.Name, metrics.Result(err)).Inc()
 	return result, err
 }
+
+// newWrangler initializes a new Vitess Wrangler that gives us access to information
+// about resharding workflows.
+func newWrangler(ctx context.Context, lockserverSpec planetscalev2.VitessLockserverParams) (*wrangler.Wrangler, error) {
+	// We need to initialize for the first time if we got here.
+	ts, err := toposerver.Open(ctx, lockserverSpec)
```
enisoc, line 224: We need to ensure this gets closed somehow, or else we'll leak ref counts for this connection.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -179,6 +184,19 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
 	topoResult, err := r.reconcileTopology(ctx, vtk)
 	resultBuilder.Merge(topoResult, err)
 
+	wr, err := newWrangler(ctx, vtk.Spec.GlobalLockserver)
```
enisoc, line 187: It seems like you continue on even if we fail to make a wrangler. It makes sense that we don't abort the whole Reconcile just because we can't get resharding status, but we should avoid trying to use the nonexistent  wrangler. It'll probably be clearer to fix this and the Close() problem if we make a subroutine like `r.reconcileResharding`. It would also be nice to put that and all resharding-related code in a new file, `reconcile_resharding.go`.

pkg/apis/planetscale/v2/defaults.go
```diff
@@ -91,7 +91,7 @@ const (
 	// DefaultMysqlPortName is the name for the MySQL port.
 	DefaultMysqlPortName = "mysql"
 
-	defaultVitessLiteImage = "vitess/lite:v7.0.0"
+	defaultVitessLiteImage = "vitess/lite:latest"
```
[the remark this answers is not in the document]
PrismaPhonic: I had planned to merge this into a dev branch until we are all done, as we are expecting more upstream changes.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -179,6 +184,19 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
+	if err != nil {
+		r.recorder.Eventf(vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to retrieve resharding information: %v", err)
+	}
+
+	workflows, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
```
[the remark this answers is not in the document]
PrismaPhonic: They are sorted, yes.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -291,6 +291,10 @@ type VitessKeyspaceStatus struct {
 	// If Idle is True, the keyspace is not deployed in any cells, so it should
 	// be safe to turn down the keyspace.
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
+	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
+	ReshardingInProgress bool `json:"reshardingInProgress,omitempty"`
+	// ActiveWorkflows is a list of the current VReplication workflows in progress.
+	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
```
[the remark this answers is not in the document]
PrismaPhonic: From an OSS perspective, it's useful to know about active workflows so you can manually check up on their current status via `Workflow Show`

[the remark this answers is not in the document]
PrismaPhonic: In reference to your question about which shards are serving traffic, that can't be known without drilling down into an active workflow.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -179,6 +184,19 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
+	if err != nil {
+		r.recorder.Eventf(vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to list active resharding workflows: %v", err)
+		workflows = make([]string, 0)
+	}
+	vtk.Status.ReshardingInProgress = len(workflows) != 0
```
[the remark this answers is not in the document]
PrismaPhonic: That's a good point - but if that's the case I can't think of a way to know for sure if resharding is in progress without storing state about workflows we intend to execute. From an OSS perspestive, it might not even be possible at all.

PR #116, opened 2020-08-10
Backup Annotiations

PrismaPhonic:
This PR adds support to pass annotations down to both backup pods and the backup sub-controller. A use case for this feature is to support IAM access.


pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -72,6 +72,9 @@ type VitessBackupLocation struct {
 	// a shared host path for local testing.
 	// +kubebuilder:validation:EmbeddedResource
 	Volume *corev1.VolumeSource `json:"volume,omitempty"`
+	// Annotations can optionally be used to attach custom annotations to Pods
+	// created for this component.
```
enisoc, line 76:
```suggestion
	// that need access to this backup storage location.
```

pkg/controller/vitessbackupstorage/reconcile_subcontroller.go
```diff
@@ -77,14 +77,16 @@ func (r *ReconcileVitessBackupStorage) reconcileSubcontroller(ctx context.Contex
...
-			return &corev1.Pod{
+			pod := &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: key.Namespace,
 					Name:      key.Name,
 					Labels:    labels,
 				},
 				Spec: *spec,
 			}
+			update.Annotations(&pod.Annotations, vbs.Spec.Location.Annotations)
```
enisoc, line 88: We should also add the annotations in UpdateInPlace.

pkg/controller/vitessshard/reconcile_backup_job.go
```diff
@@ -219,31 +220,35 @@ func vtbackupSpec(key client.ObjectKey, vts *planetscalev2.VitessShard, parentLa
...
+		DatabaseName:             vts.Spec.DatabaseName,
+		DatabaseInitScriptSecret: vts.Spec.DatabaseInitScriptSecret,
+		BackupLocation:           backupLocation,
+		BackupEngine:             vts.Spec.BackupEngine,
+		InitContainers:           pool.InitContainers,
+		SidecarContainers:        pool.SidecarContainers,
+		ExtraEnv:                 pool.ExtraEnv,
+		Annotations:              vts.Spec.Annotations,
+	}
+	update.Annotations(&tabletSpec.Annotations, backupLocation.Annotations)
```
enisoc, line 243: We should also do this for vttablet Pods, since they also access backup storage.

pkg/controller/vitessshard/reconcile_backup_job.go
```diff
@@ -219,31 +220,35 @@ func vtbackupSpec(key client.ObjectKey, vts *planetscalev2.VitessShard, parentLa
...
+		DataVolumePVCSpec:        pool.DataVolumeClaimTemplate,
+		KeyspaceName:             keyspaceName,
+		DatabaseName:             vts.Spec.DatabaseName,
+		DatabaseInitScriptSecret: vts.Spec.DatabaseInitScriptSecret,
+		BackupLocation:           backupLocation,
+		BackupEngine:             vts.Spec.BackupEngine,
+		InitContainers:           pool.InitContainers,
+		SidecarContainers:        pool.SidecarContainers,
+		ExtraEnv:                 pool.ExtraEnv,
+		Annotations:              vts.Spec.Annotations,
```
enisoc, line 241: The field `vts.Spec.Annotations` is for annotations on the VitessShard object. I think what we want is `pool.Annotations`?

Also, did you check that this field is respected for vtbackup Pods even though it wasn't being filled in before?

pkg/apis/planetscale/v2/vitessbackupstorage_types.go
PrismaPhonic: Done.

pkg/controller/vitessbackupstorage/reconcile_subcontroller.go
PrismaPhonic: Done.

pkg/controller/vitessshard/reconcile_backup_job.go
PrismaPhonic: Done.

pkg/controller/vitessshard/reconcile_backup_job.go
PrismaPhonic: Done. Also, I'm not sure what you mean by `respected`. Would you mind re-phrasing that question?

enisoc: The code that builds vtbackup Pods is separate from the code that builds vttablet Pods. Only some of these vttablet spec fields are relevant to vtbackup Pods, so by default a field added here will have no effect on vtbackup Pods unless we go fix the vtbackup Pod code to look at the new field.

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -306,7 +306,7 @@ func vttabletSpecs(vts *planetscalev2.VitessShard, parentLabels map[string]strin
 				drain.SupportedAnnotation: "ensure that the tablet is not a master",
 			}
 			update.Annotations(&annotations, pool.Annotations)
-
+			update.Annotations(&annotations, backupLocation.Annotations)
```
enisoc, line 309: I think we need to watch out for the case when `backupLocation` is `nil`. Unlike vtbackup, we still reconcile vttablets when backups are not configured.

PR #118, opened 2020-08-13
Support keyspace name with underscore

PinZhang:
According to https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions,
underscore is not allowed in the name of CRD resource, we replace the underscore with hypen as a workaround.


enisoc:
We support underscores now after #160.

Issue #124, opened 2020-08-28
Resharding Workflow Status Updates


PrismaPhonic:
We need the following status updates pulled from the new `Workflow Show` command:

1.  Status indicating which shards are actively serving (MasterIsServing field from Workflow Show return)

- example: `servingShards: -80,80-`

2. Status showing whether we are `Copying` or `Running` for a given workflow.

- example: `workflowState: Copying`

3. Status boolean field to indicate whether we have gone above or below an acceptable max vreplication lag to allow for switching writes. As of now that acceptable value is 10s. We could allow configuring this value in the future, but for v1, let's just stick to this hard coded value.

- example: `replicationLagUnsafe: true`

## Blockers
Item 3 is currently blocked until this PR is merged: https://github.com/vitessio/vitess/pull/6635

PR #126, opened 2020-09-02
Status Updates from Workflow Show

PrismaPhonic:
This PR provides status updates pulled from `Workflow Show` as detailed in this issue: https://github.com/planetscale/vitess-operator/issues/124

### DIFF
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -294,8 +294,33 @@ type VitessKeyspaceStatus struct {
 	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
 	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
 	// ActiveWorkflows is a list of the current VReplication workflows in progress.
-	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
+	ActiveWorkflows []WorkflowStatus `json:"activeWorkflows,omitempty"`
+	// ServingShards is a list of shards that are currently serving writes.
+	ServingShards []string `json:"servingShards,omitempty"`
 }
+
+// WorkflowStatus defines some of the workflow related status information.
+type WorkflowStatus struct {
+	// Workflow represents the name of the workflow relevant to the related replication statuses.
+	Workflow string `json:"workflow"`
+	// State is either 'Running', 'Copying', or 'Lagging'.
+	State WorkflowState `json:"state"`
+}
+
+// WorkflowState represents the current state for the given Workflow.
+type WorkflowState string
+
+const (
+	// RunningState indicates that the workflow is currently in the Running state. This state
+	// indicates that vreplication is ongoing, but we have moved passed the copying phase.
+	RunningState WorkflowState = "Running"
+	// CopyingState indicates that the workflow is currently in the Copying state.
+	CopyingState WorkflowState = "Copying"
+	// Lagging indicates that the workflow currently has vreplication lag exceeding 10 seconds.
+	LaggingState WorkflowState = "Lagging"
+	// ErrorState indicates that the workflow is currently experiencing some kind of error.
+	ErrorState WorkflowState = "Error"
+)
 
 // NewVitessKeyspaceStatus creates a new status object with default values.
 func NewVitessKeyspaceStatus() VitessKeyspaceStatus {
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -2,6 +2,8 @@ package vitesskeyspace
 
 import (
 	"context"
+	"sort"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -20,18 +22,53 @@ func (r *ReconcileVitessKeyspace) reconcileResharding(ctx context.Context, vtk *
 	}
 	defer ts.Close()
 
-	workflows, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
+	workflowList, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
 	if err != nil {
-		workflows = make([]string, 0)
+		workflowList = make([]string, 0)
 	}
 
-	vtk.Status.ActiveWorkflows = workflows
-	// TODO: Address when ConditionUnknown is applicable.
-	if len(workflows) != 0 {
-		vtk.Status.ReshardingInProgress = corev1.ConditionTrue
-	} else {
-		vtk.Status.ReshardingInProgress = corev1.ConditionFalse
+	reshardingInProgress := corev1.ConditionUnknown
+	workflows := make([]planetscalev2.WorkflowStatus, 0, len(workflowList))
+	for _, workflowName := range workflowList {
+		workflow, err := wr.ShowWorkflow(ctx, workflowName, vtk.Spec.Name)
+		if err != nil {
+			return err
+		}
+		workflowStatus := planetscalev2.WorkflowStatus{
+			Workflow: workflow.Workflow,
+		}
+
+		for name, status := range workflow.ShardStatuses {
+			if status.MasterIsServing {
+				shard := strings.Split(name, "/")[0]
+				vtk.Status.ServingShards = append(vtk.Status.ServingShards, shard)
+			}
+			for _, vReplRow := range status.MasterReplicationStatuses {
+				if vReplRow.State == "Error" {
+					workflowStatus.State = planetscalev2.ErrorState
+					break
+				}
+				if vReplRow.State == "Copying" {
+					workflowStatus.State = planetscalev2.CopyingState
+				}
+				if vReplRow.State == "Lagging" && workflowStatus.State != planetscalev2.CopyingState {
+					workflowStatus.State = planetscalev2.LaggingState
+				}
+				if vReplRow.State == "Running" && workflowStatus.State == "" {
+					workflowStatus.State = planetscalev2.RunningState
+				}
+			}
+		}
+		workflows = append(workflows, workflowStatus)
+
+		if workflow.SourceLocation.Keyspace == workflow.TargetLocation.Keyspace {
+			reshardingInProgress = corev1.ConditionTrue
+		} else if reshardingInProgress != corev1.ConditionTrue {
+			reshardingInProgress = corev1.ConditionFalse
+		}
 	}
+	vtk.Status.ReshardingInProgress = reshardingInProgress
+	sort.Strings(vtk.Status.ServingShards)
 
 	return nil
 }
--- a/go.mod
+++ b/go.mod
@@ -18,11 +18,11 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a // indirect
 	k8s.io/kubernetes v1.16.2
-	k8s.io/utils v0.0.0-20191010214722-8d271d903fe4
+	k8s.io/utils v0.0.0-20191114184206-e782cd3c129f
 	sigs.k8s.io/controller-runtime v0.4.0
 	sigs.k8s.io/controller-tools v0.2.4
 	sigs.k8s.io/yaml v1.1.0
-	vitess.io/vitess v0.0.0-20200810101014-2130a0e75923
+	vitess.io/vitess v0.0.0-20200901235543-6073f0f6ba43
 )
 
 // ****************************
--- a/go.sum
+++ b/go.sum
@@ -176,6 +176,7 @@ github.com/cockroachdb/cockroach-go v0.0.0-20181001143604-e0a95dfd547c/go.mod h1
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd h1:qMd81Ts1T2OTKmB4acZcyKaMtRnY5Y44NuXGX2GFJ1w=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
 github.com/codegangsta/cli v1.20.0/go.mod h1:/qJNoX69yVSKu5o4jLyXAENLRyk1uhi7zkbQ3slBdOA=
+github.com/codegangsta/inject v0.0.0-20150114235600-33e0aa1cb7c0/go.mod h1:4Zcjuz89kmFXt9morQgcfYZAYZ5n8WHjt81YYWIwtTM=
 github.com/codegangsta/negroni v1.0.0/go.mod h1:v0y3T5G7Y1UlFfyxFn/QLRU4a2EuNau2iZY63YTKWo0=
 github.com/container-storage-interface/spec v1.1.0/go.mod h1:6URME8mwIBbpVyZV93Ce5St17xBiQJQY67NDsuohiy4=
 github.com/containerd/console v0.0.0-20170925154832-84eeaae905fa/go.mod h1:Tj/on1eG8kiEhd0+fhSDzsPAFESxzBBvdyEgyryXffw=
@@ -214,6 +215,7 @@ github.com/coreos/prometheus-operator v0.34.0/go.mod h1:Li6rMllG/hYIyXfMuvUwhyC+
 github.com/coreos/rkt v1.30.0/go.mod h1:O634mlH6U7qk87poQifK6M2rsFNt+FyUTWNMnP1hF1U=
 github.com/corpix/uarand v0.1.1/go.mod h1:SFKZvkcRoLqVRFZ4u25xPmp6m9ktANfbpXZ7SJ0/FNU=
 github.com/cpuguy83/go-md2man v1.0.10/go.mod h1:SmD6nW6nTyfqj6ABTjUi3V3JVMnlJmwcJI5acqYI6dE=
+github.com/cyberdelia/go-metrics-graphite v0.0.0-20161219230853-39f87cc3b432/go.mod h1:xwIwAxMvYnVrGJPe2FKx5prTrnAjGOD8zvDOnxnrrkM=
 github.com/cyphar/filepath-securejoin v0.2.2 h1:jCwT2GTP+PY5nBz3c/YL5PAIbusElVrPujOBSCj8xRg=
 github.com/cyphar/filepath-securejoin v0.2.2/go.mod h1:FpkQEhXnPnOthhzymB7CGsFk2G9VLXONKD9G7QGMM+4=
 github.com/cznic/b v0.0.0-20180115125044-35e9bbe41f07/go.mod h1:URriBxXwVq5ijiJ12C7iIZqlA69nTlI+LgI6/pwftG8=
@@ -319,6 +321,7 @@ github.com/go-logr/logr v0.1.0/go.mod h1:ixOQHD9gLJUVQQ2ZOR7zLEifBX6tGkNJF4QyIY7
 github.com/go-logr/zapr v0.1.0/go.mod h1:tabnROwaDl0UNxkVeFRbY8bwB37GwRv0P8lg6aAiEnk=
 github.com/go-logr/zapr v0.1.1 h1:qXBXPDdNncunGs7XeEpsJt8wCjYBygluzfdLO0G5baE=
 github.com/go-logr/zapr v0.1.1/go.mod h1:tabnROwaDl0UNxkVeFRbY8bwB37GwRv0P8lg6aAiEnk=
+github.com/go-martini/martini v0.0.0-20170121215854-22fa46961aab/go.mod h1:/P9AEU963A2AYjv4d1V5eVL1CQbEJq6aCNHDDjibzu8=
 github.com/go-openapi/analysis v0.0.0-20180825180245-b006789cd277/go.mod h1:k70tL6pCuVxPJOHXQ+wIac1FUrvNkHolPie/cLEU6hI=
 github.com/go-openapi/analysis v0.17.0/go.mod h1:IowGgpVeD0vNm45So8nr+IcQ3pxVtpRoBWb8PVZO0ik=
 github.com/go-openapi/analysis v0.18.0/go.mod h1:IowGgpVeD0vNm45So8nr+IcQ3pxVtpRoBWb8PVZO0ik=
@@ -557,6 +560,7 @@ github.com/heketi/tests v0.0.0-20151005000721-f3775cbcefd6/go.mod h1:xGMAM8JLi7U
 github.com/heketi/utils v0.0.0-20170317161834-435bc5bdfa64/go.mod h1:RYlF4ghFZPPmk2TC5REt5OFwvfb6lzxFWrTWB+qs28s=
 github.com/helm/helm-2to3 v0.2.0 h1:1wgpvw93CHlO+mC6AMrfc5xboYyk1MkoI3ku7eI5uvE=
 github.com/helm/helm-2to3 v0.2.0/go.mod h1:jQUVAWB0bM7zNIqKPIfHFzuFSK0kHYovJrjO+hqcvRk=
+github.com/howeyc/gopass v0.0.0-20190910152052-7cb4b85ec19c/go.mod h1:lADxMC39cJJqL93Duh1xhAs4I2Zs8mKS89XWXFGp9cs=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
 github.com/huandu/xstrings v1.2.0 h1:yPeWdRnmynF7p+lLYz0H2tthW9lqhMJrQV/U7yy4wX0=
@@ -656,6 +660,9 @@ github.com/maorfr/helm-plugin-utils v0.0.0-20181205064038-588190cb5e3b/go.mod h1
 github.com/markbates/inflect v1.0.4 h1:5fh1gzTFhfae06u3hzHYO9xe3l3v3nW5Pwt3naLTP5g=
 github.com/markbates/inflect v1.0.4/go.mod h1:1fR9+pO2KHEO9ZRtto13gDwwZaAKstQzferVeWqbgNs=
 github.com/marten-seemann/qtls v0.2.3/go.mod h1:xzjG7avBwGGbdZ8dTGxlBnLArsVKLvwmjgmPuiQEcYk=
+github.com/martini-contrib/auth v0.0.0-20150219114609-fa62c19b7ae8/go.mod h1:ahTFgV/NtzY/CALneRrC67m1dis5arHTQDfyIhKk69E=
+github.com/martini-contrib/gzip v0.0.0-20151124214156-6c035326b43f/go.mod h1:jhUB0rZB2TPWqy0yGugKRRictO591eSO7If7O4MfCaA=
+github.com/martini-contrib/render v0.0.0-20150707142108-ec18f8345a11/go.mod h1:Ah2dBMoxZEqk118as2T4u4fjfXarE0pPnMJaArZQZsI=
 github.com/martinlindhe/base36 v1.0.0 h1:eYsumTah144C0A8P1T/AVSUk5ZoLnhfYFM3OGQxB52A=
 github.com/martinlindhe/base36 v1.0.0/go.mod h1:+AtEs8xrBpCeYgSLoY/aJ6Wf37jtBuR0s35750M27+8=
 github.com/mattbaird/jsonpatch v0.0.0-20171005235357-81af80346b1a h1:+J2gw7Bw77w/fbK7wnNJJDKmw1IbWft2Ul5BzrG1Qm8=
@@ -687,6 +694,7 @@ github.com/mattn/go-shellwords v1.0.5/go.mod h1:3xCvwCdWdlDJUrvuMn7Wuy9eWs4pE8vq
 github.com/mattn/go-sqlite3 v1.9.0/go.mod h1:FPy6KqzDD04eiIsT53CuJW3U88zkxoIYsOqkbpncsNc=
 github.com/mattn/go-sqlite3 v1.10.0 h1:jbhqpg7tQe4SupckyijYiy0mJJ/pRyHvXf7JdWK860o=
 github.com/mattn/go-sqlite3 v1.10.0/go.mod h1:FPy6KqzDD04eiIsT53CuJW3U88zkxoIYsOqkbpncsNc=
+github.com/mattn/go-sqlite3 v1.14.0/go.mod h1:JIl7NbARA7phWnGvh0LKTyg7S9BA+6gx71ShQilpsus=
 github.com/matttproud/golang_protobuf_extensions v1.0.1 h1:4hp9jkHxhMHkqkrB3Ix0jegS5sx/RkqARlsWZ6pIwiU=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
 github.com/maxbrunsfeld/counterfeiter/v6 v6.2.1/go.mod h1:F9YacGpnZbLQMzuPI0rR6op21YvNu/RjL705LJJpM3k=
@@ -736,6 +744,7 @@ github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742/go.mod h1:bx2lN
 github.com/modern-go/reflect2 v1.0.1 h1:9f412s+6RmYXLWZSEzVVgPGK7C2PphHj5RJrvfx9AWI=
 github.com/modern-go/reflect2 v1.0.1/go.mod h1:bx2lNnkwVCuqBIxFjflWJWanXIb3RllmbCylyMrvgv0=
 github.com/mohae/deepcopy v0.0.0-20170603005431-491d3605edfb/go.mod h1:TaXosZuwdSHYgviHp1DAtfrULt5eUgsSMsZf+YrPgl8=
+github.com/montanaflynn/stats v0.6.3/go.mod h1:wL8QJuTMNUDYhXwkmfOly8iTdp5TEcJFWZD2D7SIkUc=
 github.com/morikuni/aec v0.0.0-20170113033406-39771216ff4c h1:nXxl5PrvVm2L/wCy8dQu6DMTwH4oIuGN8GJDAlqDdVE=
 github.com/morikuni/aec v0.0.0-20170113033406-39771216ff4c/go.mod h1:BbKIizmSmc5MMPqRYbxO4ZU0S0+P200+tUnFx7PXmsc=
 github.com/mrunalp/fileutils v0.0.0-20160930181131-4ee1cc9a8058/go.mod h1:x8F1gnqOkIEiO4rqoeEEEqQbo7HjGMTvyoq3gej4iT0=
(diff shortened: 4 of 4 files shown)


pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
 	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
 	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
 	// ActiveWorkflows is a list of the current VReplication workflows in progress.
-	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
+	ActiveWorkflows []WorkflowStatus `json:"activeWorkflows,omitempty"`
+	// ServingShards is a list of shards that are currently serving writes.
+	ServingShards []string `json:"servingShards,omitempty"`
```
enisoc, line 299: We're only going to fetch this info if resharding is active, right? If so, I think we should put it inside a `resharding:` struct so everything is grouped. Then we have a place to add other info like `sourceShards` and `targetShards` so we can tell which direction resharding is going.

As discussed below, I think we should also add a generic `conditions:` list like we did for VitessShardStatus, and move things like `ReshardingInProgress` there with full reason/message/lastTransitionTime. We can copy and/or refactor the code you wrote to manage those fields for VitessShardCondition.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +15,69 @@ import (
...
 	if err != nil {
 		return err
 	}
 	defer ts.Close()
 
-	workflows, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
+	workflowList, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
 	if err != nil {
-		workflows = make([]string, 0)
+		workflowList = make([]string, 0)
```
enisoc, line 31: We should bail out of `reconcileResharding` (with success but requeue, similar to topo failures) if we failed to list workflows, instead of claiming resharding is inactive.

We can borrow the code from VitessShard where we register status conditions in advance, so that they will be set to Unknown if we never give them an updated value.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +15,69 @@ import (
...
 
-	vtk.Status.ActiveWorkflows = workflows
-	// TODO: Address when ConditionUnknown is applicable.
-	if len(workflows) != 0 {
-		vtk.Status.ReshardingInProgress = corev1.ConditionTrue
-	} else {
-		vtk.Status.ReshardingInProgress = corev1.ConditionFalse
+	reshardingInProgress := corev1.ConditionUnknown
+	workflows := make([]planetscalev2.WorkflowStatus, 0, len(workflowList))
+	for _, workflowName := range workflowList {
```
enisoc, line 36: We should skip over workflows that are not resharding workflows (source/target keyspace is the same, but source/target shard sets are not exactly the same).

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +15,69 @@ import (
...
+		if err != nil {
+			return err
+		}
+		workflowStatus := planetscalev2.WorkflowStatus{
+			Workflow: workflow.Workflow,
+		}
+
+		for name, status := range workflow.ShardStatuses {
+			if status.MasterIsServing {
+				shard := strings.Split(name, "/")[0]
```
enisoc, line 47: What's the format of the key in the ShardStatuses map? Is that format guaranteed to stay the same forever? Is there a function in Vitess to parse it for us so we don't have to duplicate the logic in the operator?

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
...
 }
 
+// WorkflowStatus defines some of the workflow related status information.
+type WorkflowStatus struct {
+	// Workflow represents the name of the workflow relevant to the related replication statuses.
+	Workflow string `json:"workflow"`
+	// State is either 'Running', 'Copying', or 'Lagging'.
```
enisoc, line 306: or `Error`

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
...
+// WorkflowState represents the current state for the given Workflow.
+type WorkflowState string
+
+const (
+	// RunningState indicates that the workflow is currently in the Running state. This state
+	// indicates that vreplication is ongoing, but we have moved passed the copying phase.
+	RunningState WorkflowState = "Running"
+	// CopyingState indicates that the workflow is currently in the Copying state.
+	CopyingState WorkflowState = "Copying"
+	// Lagging indicates that the workflow currently has vreplication lag exceeding 10 seconds.
```
enisoc, line 321: What's the difference between this `Lagging` state and the new transaction_timestamp stuff you added recently? From the Vitess code, it looks like the Lagging state only takes time_updated into account?

We should think about whether we need to expose both types of lag (time_updated vs. transaction_timestamp) in our status API. It seems confusing so it might be better to hide the complexity somehow.

For example, consumers currently need to know that Lagging is really a special case of Running (so a UI should still show that resharding is running, for example), and also that Lagging is a transient state that might revert to Running at any moment. Maybe we should just look at transaction_timestamp to report lag orthogonally to state, and hide this not-very-useful Lagging state by reporting Running at our status API level?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +15,69 @@ import (
...
+				vtk.Status.ServingShards = append(vtk.Status.ServingShards, shard)
+			}
+			for _, vReplRow := range status.MasterReplicationStatuses {
+				if vReplRow.State == "Error" {
```
enisoc, line 51: We should be able to extract the error message, right? We could use that in the `message` field of a status condition for resharding, e.g. to explain why the `ReshardingInSync` condition is `False`.

We'd have to be careful though to avoid spurious updates. Are the entries in MasterReplicationStatuses always in the same order? We also need to decide if we show all errors, or only the first row with an error.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +15,69 @@ import (
...
+	for _, workflowName := range workflowList {
+		workflow, err := wr.ShowWorkflow(ctx, workflowName, vtk.Spec.Name)
+		if err != nil {
+			return err
+		}
+		workflowStatus := planetscalev2.WorkflowStatus{
+			Workflow: workflow.Workflow,
+		}
+
+		for name, status := range workflow.ShardStatuses {
```
enisoc, line 45: Could you add a comment above this loop explaining the logic for aggregating status across all the vreplication rows for all the shards?

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
 	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
 	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
 	// ActiveWorkflows is a list of the current VReplication workflows in progress.
-	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
+	ActiveWorkflows []WorkflowStatus `json:"activeWorkflows,omitempty"`
```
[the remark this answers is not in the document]
PrismaPhonic: I had thought about this while building this out - and you're right, in an ideal world. However, it is possible to have more than one at a time. What do we do if that's the state we're in?

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: The format is `shard/tablet-alias`. I don't know if it will stay the same forever. There is not a function in Vitess to parse it for us.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: These states mimick the four states that we will see in the vrepl row State results. @rohit-nayak-ps designed it such that the `State` field gets updated based on a set of criteria. `Lagging` already existed as a given state, but isn't as accurate as our new `MaxVReplicationLag` field. Yes, it only takes time_updated into account. 

This is mostly so we can translate from these known states. I guess another option is we can remove "Lagging" entirely and skip over it if we find it in a `ReplicationStatusResult`

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Yes, the error message will be in the `Message` field of the `ReplicationStatus` (which really represents a vrepl row). The same one that we are looking at the State field on. 

I think for simplicity's sake we can just show the first error for now. We can always expand that in the future to include all errors.

PrismaPhonic: To answer your question on ordering - my guess is no. It's the result of a SQL query with no `ORDER BY`, so I don't think order can be guaranteed.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Sure thing.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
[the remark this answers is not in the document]
enisoc: Would it be a legitimate use of vreplication to have more than one resharding workflow active at once? Or is it considered an error to be in that state?

pkg/controller/vitesskeyspace/reconcile_resharding.go
enisoc: Can there ever be more than one entry in the map for the same shard, but different tablet aliases? If there can only be one per shard, is it too late to change the format upstream to be just the shard name and put the tablet alias inside the struct? Or are we the first ones to try to use this struct?

pkg/controller/vitesskeyspace/reconcile_resharding.go
enisoc: Showing the first error sounds good for now. We just need to make sure to iterate in deterministic order so we don't pick different "first errors" for the same actual data set.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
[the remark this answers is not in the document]
PrismaPhonic: That's a good question and one I'm not a voice of authority on to answer. @rohit-nayak-ps ?

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: No, there can't be, because we are only iterating over master tablets and we only have 1 master per shard. I would be happy to change the format upstream but I think this did get included in 7.0 and wasn't even identified as experimental so we might need to think about backwards compatibility. I'm not sure if anyone is depending on this yet. As of that release, this only effected what got logged though, so I can't imagine there's a strong dependency from a user.

enisoc: Does the code know how to ignore false masters?

PrismaPhonic: It just gets masters from topo and trusts that as source of truth.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: It sounds like for that to be true we have to collect the errors regardless, sort them and then return the first?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +31,73 @@ import (
...
 )
 
-func (r *ReconcileVitessKeyspace) reconcileResharding(ctx context.Context, vtk *planetscalev2.VitessKeyspace) error {
-	wr, ts, err := newWrangler(ctx, vtk.Spec.GlobalLockserver)
+const (
+	maxSafeVReplicationLag = 10
+)
+
+func (r *reconcileHandler) reconcileResharding(ctx context.Context) error {
+	wr, ts, err := newWrangler(ctx, r.vtk.Spec.GlobalLockserver)
```
avaidyanatha, line 39: Do we still need this if we initialize a wrangler with `tsInit()` that lives on the `reconcileHandler` object? Or is that for a different purpose?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +31,73 @@ import (
...
+				if vReplRow.State == "Running" && workflowStatus.State == "" {
+					workflowStatus.State = planetscalev2.RunningState
+				}
+			}
+		}
+		workflows = append(workflows, workflowStatus)
+
+		if workflow.SourceLocation.Keyspace == workflow.TargetLocation.Keyspace {
+			reshardingInProgress = corev1.ConditionTrue
+		} else if reshardingInProgress != corev1.ConditionTrue {
```
avaidyanatha: So for my knowledge, if we've reached this point, this is guaranteed to not be a `ConditionUnknown`?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -13,25 +31,73 @@ import (
...
+			reshardingInProgress = corev1.ConditionFalse
+		}
+
+		// If MaxVReplicationLag ever exceeds max safe value, we need to set UnsafeVReplicationLag to true.
```
avaidyanatha, line 95: +1 to @enisoc's comment about making the variable's `true` state the positive one, so the logic would change here.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: I've done part 2 now of what you mentioned. What do you think we should do if we end up getting back more than one active resharding workflow?

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Still working on it - but already removed in the code that's local. Will push up soon.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Yeah, or it's already false and we don't mind overwriting it.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
...
+	UnsafeVReplicationLag bool `json:"unsafeVReplicationLag,omitempty"`
+}
+
+// WorkflowState represents the current state for the given Workflow.
+type WorkflowState string
+
+const (
+	// RunningState indicates that the workflow is currently in the Running state. This state
+	// indicates that vreplication is ongoing, but we have moved passed the copying phase.
+	RunningState WorkflowState = "Running"
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
...
+	// State is either 'Running', 'Copying', or 'Lagging'.
+	State WorkflowState `json:"state"`
+	// UnsafeVReplicationLag indicates if VReplication lag is currently exceeding 10s for the given workflow
+	UnsafeVReplicationLag bool `json:"unsafeVReplicationLag,omitempty"`
+}
+
+// WorkflowState represents the current state for the given Workflow.
+type WorkflowState string
+
+const (
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -294,9 +294,36 @@ type VitessKeyspaceStatus struct {
...
 }
 
+// WorkflowStatus defines some of the workflow related status information.
+type WorkflowStatus struct {
+	// Workflow represents the name of the workflow relevant to the related replication statuses.
+	Workflow string `json:"workflow"`
+	// State is either 'Running', 'Copying', or 'Lagging'.
+	State WorkflowState `json:"state"`
+	// UnsafeVReplicationLag indicates if VReplication lag is currently exceeding 10s for the given workflow
+	UnsafeVReplicationLag bool `json:"unsafeVReplicationLag,omitempty"`
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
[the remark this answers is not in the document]
rohit-nayak-ps: At this moment, it is possible to have more than one resharding workflow to be active. Will discuss with Sugu if there is a reasonable use case for allowing such a scenario, but currently there is no check, on a reshard, to validate if there is another resharding workflow in progress.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -152,60 +152,91 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
 	// Check resharding status and report back.
-	if err := r.reconcileResharding(ctx, vtk); err != nil {
-		r.recorder.Eventf(vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to retrieve resharding information: %v", err)
+	if err := handler.reconcileResharding(ctx); err != nil {
+		handler.recorder.Eventf(handler.vtk, corev1.EventTypeWarning, "StatusUpdateWarning", "failed to retrieve resharding information: %v", err)
 		resultBuilder.Error(err)
 	}
 
-	// Update status if needed.
-	vtk.Status.ObservedGeneration = vtk.Generation
```
enisoc, line 194: We still need to do this somewhere.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -152,60 +152,91 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
+			return nil, nil
+		}
+		// Error reading the object - return error so caller re-queues.
+		return nil, err
+	}
+	planetscalev2.DefaultVitessKeyspace(vtk)
+
+	oldStatus := vtk.Status.DeepCopy()
+	vtk.Status = planetscalev2.NewVitessKeyspaceStatus()
+	vtk.Status.Conditions = oldStatus.Conditions
```
enisoc, line 219:
```suggestion
	vtk.Status.Conditions = oldStatus.DeepCopyConditions()
```

pkg/controller/vitesskeyspace/reconcile_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	wr *wrangler.Wrangler
+	// This field holds a tablet manager client internally for closing upon collection of reconcileHandler.
+	// Please don't try to access until you have run tsInit().
+	tmc tmclient.TabletManagerClient
+}
+
+// tsInit will initialize a toposerver connection, as well as a
+// tablet manager client and wrangler for subroutine use.
+func (r *reconcileHandler) tsInit(ctx context.Context) error {
+	if r == nil || r.ts != nil {
```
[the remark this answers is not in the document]
PrismaPhonic: It would be a bug, however the second check is still valid (checking for r.ts being set already)

pkg/controller/vitesskeyspace/reconcile_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	wr := wrangler.New(logutil.NewConsoleLogger(), r.ts.Server, r.tmc)
+	r.wr = wr
+
+	return nil
+}
+
+// close should be called in a defer upon construction of a reconcileHandler to
+// defer the closing of underlying topo if we successfully created one.
+func (r *reconcileHandler) close() {
+	if r == nil {
```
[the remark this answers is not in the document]
PrismaPhonic: My thought more here was that if receiver is nil then there's nothing to close. It's true though that r should never really be nil. a defer close call should be done after constructing a reconcileHandler.

pkg/controller/vitesskeyspace/reconcile_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+}
+
+func (r *reconcileHandler) updateStatus(ctx context.Context) error {
+	// Before updating status, set conditions we haven't touched to unknown.
+	for condition := range r.untouchedConditions {
+		r.setConditionStatus(condition, v1.ConditionUnknown, "ReconcileFailed", "Failed to determine status of the condition.")
+	}
+
+	if !equality.Semantic.DeepEqual(&r.vtk.Status, &r.oldStatus) {
```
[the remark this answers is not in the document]
PrismaPhonic: This logic is copied from the main Reconcile loop, and is the same for every controller. So does every controller Reconcile just update on every single loop regardless?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -1,52 +1,111 @@
...
-	if err != nil {
-		workflows = make([]string, 0)
+	reshardingInProgress := corev1.ConditionUnknown
+	var workflowStatus *planetscalev2.ReshardingStatus
+	for _, workflowName := range workflowList {
+		workflow, err := r.wr.ShowWorkflow(ctx, workflowName, r.vtk.Spec.Name)
+		if err != nil {
+			return err
+		}
+		if workflow.SourceLocation.Keyspace != workflow.TargetLocation.Keyspace {
```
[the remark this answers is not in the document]
PrismaPhonic: We could also check for this, but I didn't realize that's a valid vreplication use case, so I didn't bother.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -1,52 +1,111 @@
...
+		reshardingInProgress = corev1.ConditionTrue
+
+		if workflowStatus != nil {
+			return fmt.Errorf("found more than one active resharding workflow")
+		}
+
+		workflowStatus = &planetscalev2.ReshardingStatus{
+			Workflow:     workflow.Workflow,
+			State:        planetscalev2.WorkflowUnknown,
+			SourceShards: workflow.SourceLocation.Shards,
```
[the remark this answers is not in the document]
PrismaPhonic: I wrote the code for generating the shards lists for `SourceLocation` and `TargetLocation`. They do get sorted.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -1,52 +1,111 @@
...
+		for name, status := range workflow.ShardStatuses {
+			if status.MasterIsServing {
+				shard := strings.Split(name, "/")[0]
+				workflowStatus.ServingShards = append(workflowStatus.ServingShards, shard)
+			}
+			for _, vReplRow := range status.MasterReplicationStatuses {
+				if vReplRow.State == "Error" {
+					workflowStatus.State = planetscalev2.WorkflowError
+					r.vtk.Status.SetConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "VReplicationError", fmt.Sprintf("Encountered a VReplication Error: %v", vReplRow.Message))
+					break
```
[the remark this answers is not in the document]
PrismaPhonic: ahhh good catch. I still want to sum up serving shards so we definitely don't want to break out of the outer loop yet. I thought I would get clever and avoid more complicated conditionals. I'll add one additional check immediately to see if workflowStatus.State is already set to WorkflowError and break on that.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
```diff
@@ -152,60 +152,91 @@ func (r *ReconcileVitessKeyspace) Reconcile(request reconcile.Request) (reconcil
...
+	defer func() {
+		err := handler.updateStatus(ctx)
+		if err != nil {
+			finalResult, finalErr = resultBuilder.Error(err)
+		}
+	}()
 
-	// Idle means the keyspace is not deployed in any cells (there are no tablet pools defined).
-	vtk.Status.Idle = k8s.ConditionStatus(len(vtk.Spec.CellNames()) == 0)
+	err = handler.tsInit(ctx)
```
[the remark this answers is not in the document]
PrismaPhonic: That's a fair point. I'll move it inside both `reconcileX()` functions that need it. It should be safe to call more than once because of the != nil check bail in tsInit.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -1,52 +1,111 @@
...
 	if err != nil {
 		return err
 	}
-	defer ts.Close()
 
-	workflows, err := wr.ListAllWorkflows(ctx, vtk.Spec.Name)
-	if err != nil {
-		workflows = make([]string, 0)
+	reshardingInProgress := corev1.ConditionUnknown
+	var workflowStatus *planetscalev2.ReshardingStatus
```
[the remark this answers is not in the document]
PrismaPhonic: ahh, I see - so basically start with a simple filter loop. That's a good idea.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -293,10 +293,42 @@ type VitessKeyspaceStatus struct {
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
 	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
 	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
-	// ActiveWorkflows is a list of the current VReplication workflows in progress.
-	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
+	// ReshardingStatus is used to provide information about an ongoing resharding operation.
+	ReshardingStatus ReshardingStatus `json:"reshardingStatus,omitempty"`
```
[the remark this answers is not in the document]
PrismaPhonic: Based on your comment update suggested text, doesn't that mean that the `ReshardingActive` subfield is now pointless? If the resharding field is only present if that field is `True` then we would never see other conditions regardless.

pkg/controller/vitesskeyspace/reconcile_handler.go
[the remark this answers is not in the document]
PrismaPhonic: Maybe r somehow got dropped from memory entirely before close() could get called. Is that possible since close only gets called from a defer? There would be no references left to `r` for the gc to worry about? Or do ref counts still consider refs inside defer calls?

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -1,52 +1,111 @@
...
+		if workflow.SourceLocation.Keyspace != workflow.TargetLocation.Keyspace {
+			if reshardingInProgress == corev1.ConditionUnknown {
+				reshardingInProgress = corev1.ConditionFalse
+			}
+			continue
+		}
+		reshardingInProgress = corev1.ConditionTrue
+
+		if workflowStatus != nil {
+			return fmt.Errorf("found more than one active resharding workflow")
```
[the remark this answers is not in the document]
PrismaPhonic: In this case we can determine that resharding is definitely active, however we are in some error state where we can't determine which resharding workflow to highlight, when considering whether we are "in sync".  So Active == True, but In Sync == Unknown, with this specific information. 

Would you agree?

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -293,10 +293,42 @@ type VitessKeyspaceStatus struct {
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
 	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
 	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_handler.go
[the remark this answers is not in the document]
PrismaPhonic: Removed the receiver nil check.

pkg/controller/vitesskeyspace/reconcile_handler.go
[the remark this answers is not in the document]
PrismaPhonic: Tentatively removed but still curious about this question ^

pkg/controller/vitesskeyspace/reconcile_resharding.go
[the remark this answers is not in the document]
PrismaPhonic: Executed on that assumption. LMK if it looks good to you.

pkg/controller/vitesskeyspace/reconcile_resharding.go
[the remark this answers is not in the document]
PrismaPhonic: Added the check.

pkg/controller/vitesskeyspace/reconcile_resharding.go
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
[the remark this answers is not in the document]
PrismaPhonic: Added logic to skip entirely if `workflowStatus.State` is already set to `WorkflowError`.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -293,10 +293,42 @@ type VitessKeyspaceStatus struct {
...
+	// ReshardingStatus is used to provide information about an ongoing resharding operation.
+	ReshardingStatus ReshardingStatus `json:"reshardingStatus,omitempty"`
+	// Conditions is a map of all VitessKeyspace specific conditions we want to set and monitor.
+	// It's ok for multiple controllers to add conditions here, and those conditions will be preserved.
+	Conditions map[VitessKeyspaceConditionType]VitessKeyspaceCondition `json:"conditions,omitempty"`
 }
 
+// ReshardingStatus defines some of the workflow related status information.
+type ReshardingStatus struct {
+	// Workflow represents the name of the workflow relevant to the related replication statuses.
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -344,6 +376,36 @@ func NewVitessKeyspaceShardStatus(spec *VitessKeyspaceKeyRangeShard) VitessKeysp
...
+	Message string `json:"message,omitempty"`
+}
+
+// VitessKeyspaceConditionType is a valid value for the key of a VitessKeyspaceCondition map where the key is a
+// VitessKeyspaceConditionType and the value is a VitessKeyspaceCondition.
+type VitessKeyspaceConditionType string
+
+// These are valid conditions of VitessKeyspace.
+const (
+	// VitessKeyspaceReshardingActive indicates that for the given keyspace, there is currently an ongoing resharding operation.
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -344,6 +376,36 @@ func NewVitessKeyspaceShardStatus(spec *VitessKeyspaceKeyRangeShard) VitessKeysp
...
+	VitessKeyspaceReshardingActive VitessKeyspaceConditionType = "ReshardingActive"
+	// VitessKeyspaceReshardingActive indicates that for the given keyspace, there is currently an ongoing resharding operation,
+	// and for that operation we are past the copying phase, and replication lag is below 10 seconds.
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/vitesskeyspace_controller.go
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Decided to go with the idea of summing up errors, sorting and returning the first of a sorted list because it seemed like a much simpler idea.

pkg/controller/vitesskeyspace/reconcile_handler.go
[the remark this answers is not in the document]
enisoc: > My thought more here was that if receiver is nil then there's nothing to close. It's true though that r should never really be nil.

In general, I think silently tolerating a nil receiver should be the exception rather than the rule. For most types, we should assume by default that it's a bug to try to call methods on a nil receiver, and letting such calls panic is the best way to catch those bugs.

The exception comes when calling a method on a nil receiver is a useful thing to do on purpose. Some examples could be if the receiver is a map or slice type (since it's conventional to treat nil the same as an empty container), or if the method does some conversion and the correct conversion of a nil value is another nil value.

In this case, if we have a mismatched close() call, I think we'd consider that a mistake and we'd want to be notified about the bug.

> Maybe r somehow got dropped from memory entirely before close() could get called. Is that possible since close only gets called from a defer? There would be no references left to r for the gc to worry about? Or do ref counts still consider refs inside defer calls?

The `defer` statement saves a local copy of any values computed in the defer expression, so they won't be garbage-collected. You can think of `defer r.f(a+g(), b)` as being equivalent to this:

```go
savedFunc := r.f
savedArg1 := a+g()
savedArg2 := b
defer savedFunc(savedArg1, savedArg2)
```

In addition, deferred functions execute before the surrounding function actually returns, so references in function-local variables are still alive:

```go
r := newThing()
defer func() {
    // r is still in scope here so it can't be GC'd
  }()
```

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -291,12 +291,42 @@ type VitessKeyspaceStatus struct {
...
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
-	// ReshardingInProgress indicates if there is a VReplication workflow in progress.
-	ReshardingInProgress corev1.ConditionStatus `json:"reshardingInProgress,omitempty"`
-	// ActiveWorkflows is a list of the current VReplication workflows in progress.
-	ActiveWorkflows []string `json:"activeWorkflows,omitempty"`
+	// ReshardingStatus is used to provide information about an ongoing resharding operation.
+	ReshardingStatus ReshardingStatus `json:"reshardingStatus,omitempty"`
+	// Conditions is a map of all VitessKeyspace specific conditions we want to set and monitor.
+	// It's ok for multiple controllers to add conditions here, and those conditions will be preserved.
+	Conditions map[VitessKeyspaceConditionType]VitessKeyspaceCondition `json:"conditions,omitempty"`
```
enisoc, line 298: Let's add a `FIXME: Convert this to a list for kubectl wait compatibility before merging the feature branch to master.`

pkg/controller/vitesskeyspace/reconcile_resharding.go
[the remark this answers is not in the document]
enisoc: > So Active == True, but In Sync == Unknown, with this specific information.

I might even say `Active = Unknown` as well and explain in the message that we don't know what to do if more than one resharding workflow claims to be active. The other subfields of `resharding` don't make sense if we don't know which workflow to look at, so we should probably leave `resharding` nil.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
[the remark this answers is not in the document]
PrismaPhonic: That's a good point. Ok, I'll update the doc comment text.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Done.

PR #127, opened 2020-09-03
Elaborate on docs for backup location name field.

enisoc:
The required matching between `spec.backup.locations[]` and `backupLocationName` was not well documented.


PrismaPhonic (approved):
LGTM

PR #128, opened 2020-09-04
Update Gateway Implementation

PrismaPhonic:
This PR updates the gateway implementation. Without this we cannot support SET statements that are supported in v7, even though `vtop` has already been updated to use v7.


mcronce (approved):

enisoc:
I believe @deepthi recommended that we upgrade staging to a newer Vitess build before we switch the gateway because they fixed a known issue in our current build.

PrismaPhonic:
We have not. I can bump to 7.0.1 first.

PR #129, opened 2020-09-04
Add a field to propagate annotations to VitessKeyspace objects.

enisoc:
This is copied from the existing field to add annotations to VitessShard objects.


PrismaPhonic (approved):
LGTM

PR #130, opened 2020-09-08
orchestrator: initial support

sougou:
This is an experimental support for Orchestrator. The deployment principles are similar to that of vtctld. There are some differences:
* API is marked as experimental.
* Default replica count is 0. If Orchestrator is unspecified, it will not be launched.
* It expects an additional secret: a config file with orchestrator and replication credentials among other things.
* It's highly recommended to set `disable_active_reparents=true` and `enable_semi_sync=false` on the tablets. Right now, the user has to set them manually. This part will be enforced after we finalize the API.

I have a corresponding PR on the vitess branch that demonstrates how this works. I'll submit that later.

There is a high likelihood that we'll quickly move away from this. One possibility is to go per-shard, in which case this can become a property in the shard object.

There are no tests because this is only for people to play with. We'll need to write tests once we finalize the API.


pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -102,7 +106,7 @@ type VitessClusterSpec struct {
 	Keyspaces []VitessKeyspaceTemplate `json:"keyspaces,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
 
 	// ExtraVitessFlags can optionally be used to pass flags to all Vitess components.
-	// WARNING: Any flags passed here must be flags that can be accepted by vtgate, vtctld and vttablet.
+	// WARNING: Any flags passed here must be flags that can be accepted by vtgate, vtctld, orchestrator and vttablet.
```
PrismaPhonic, line 109: Is this actually true? If so this seems like a pretty big downside to this integration right now. Have we verified which flags orchestrator does not accept that vtgate, vtctld and vttablet does accept?

pkg/controller/vitesscluster/reconcile_orchestrator.go
```diff
@@ -0,0 +1,183 @@
...
+	"planetscale.dev/vitess-operator/pkg/operator/results"
+)
+
+func (r *ReconcileVitessCluster) reconcileOrchestrator(ctx context.Context, vt *planetscalev2.VitessCluster) (reconcile.Result, error) {
+	key := client.ObjectKey{Namespace: vt.Namespace, Name: orchestrator.ServiceName(vt.Name)}
+	labels := map[string]string{
+		planetscalev2.ClusterLabel:   vt.Name,
+		planetscalev2.ComponentLabel: planetscalev2.OrcComponentName,
+	}
+	resultBuilder := results.Builder{}
```
PrismaPhonic, line 44: nit: we try to make this the first line of any reconcile subroutine.

pkg/controller/vitesscluster/reconcile_orchestrator.go
```diff
@@ -0,0 +1,183 @@
...
+			// We'll say orchestrator is Available overall if any of the Deployments is available.
+			// The important thing is that somebody will answer when a client hits the Service.
+			if available := conditions.Deployment(curObj.Status.Conditions, appsv1.DeploymentAvailable); available != nil {
+				// Update the overall status if either we found one that's True, or we previously knew nothing at all (Unknown).
+				if available.Status == corev1.ConditionTrue || vt.Status.VitessOrchestrator.Available == corev1.ConditionUnknown {
+					vt.Status.VitessOrchestrator.Available = available.Status
+				}
+			}
+
+			// TODO(enisoc): Aggregate other important parts of status besides conditions.
```
PrismaPhonic, line 115: Copy/paste left over?

PrismaPhonic (commented):
Most of this seems pretty boilerplate. I assume you tested this out?

Also, I had thought that vorchestrator was going to get integrated directly in with vtctld. Did I misunderstand that? This seems to deploy orchestrator as a standalone deployment?

Also, I see that you said enable_semi_sync should be false. What's the reasoning for recommending that it's false? If this is an important recommendation, it should be documented on that flag ideally.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: I would think that if Orchestrator is independent of Vitess (is this orchestrator support or vorchestrator support?) that these flags should NOT be passed to orchestrator.

sougou: This should generally be true. This version of Orchestrator is a vitess component. It shares (and requires) the same flags the other components do.

pkg/controller/vitesscluster/reconcile_orchestrator.go
sougou: Also fixed vtctld (source of copy-pasta).

pkg/controller/vitesscluster/reconcile_orchestrator.go
sougou: I've left it here because I assume the same TODO applies here also. Orchestrator shares the same concerns that vtctld has.

PrismaPhonic: That's a fair point. Ok, I'll resolve this comment.

pkg/controller/vitesscluster/reconcile_orchestrator.go
PrismaPhonic: Thanks.

pkg/apis/planetscale/v2/vitesscluster_types.go
PrismaPhonic: Ahh, I see. So this is vorchestrator? If it's a Vitess component style of orchestrator, then we might want to clear this language up some. To me as a reader, it sounds like I need to go reference the main orchestrator code base to find what flags it accepts. 

In other words, this reads as if it is a breaking change, when it's not.

sougou:
> Most of this seems pretty boilerplate. I assume you tested this out?

Yes. I only tested manually.

> Also, I had thought that vorchestrator was going to get integrated directly in with vtctld. Did I misunderstand that? This seems to deploy orchestrator as a standalone deployment?

This is only a POC. This will allow people to play with it and give early feedback. In the endgame, there may be no separate Orchestrator component.

> Also, I see that you said enable_semi_sync should be false. What's the reasoning for recommending that it's false? If this is an important recommendation, it should be documented on that flag ideally.

This is actually very important. It's going to be so important that the flag value will be either overridden by the operator, or we may go one step further and ignore it in vttablet itself. For this reason, I'm refraining from changing documentation of other components until we finalize how we're going to enforce these.

shlomi-noach (commented):
Just signalling that I've viewed this PR; I don't have the k8s/operator background to give it a technical review.

enisoc (approved):

PR #131, opened 2020-09-08
Set resharding status when we build and compute it

avaidyanatha:
## Main changes
- Self explanatory, fixing small bug from #126

### DIFF
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -133,6 +133,7 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 	}
 
 	sort.Strings(workflowStatus.ServingShards)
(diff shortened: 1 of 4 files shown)


pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -133,6 +133,7 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 	}
 
 	sort.Strings(workflowStatus.ServingShards)
+	r.vtk.Status.Resharding = *workflowStatus
```
PrismaPhonic, line 136: Rather than deref, let's just make the field type a pointer type to avoid copy semantics.

avaidyanatha: Done.

PrismaPhonic (approved):
Lgtm

PR #132, opened 2020-09-10
Add/remove partitionings in place.

enisoc:
Always immediately add or remove partitionings to stay in sync with the desired set of partitionings, even if changes to existing partitionings are held back by the update strategy. The deployment of new things or the removal of unwanted things should not need to wait for an external rollout.


enisoc:
Since this is just going into a feature dev branch, I'm going to go ahead and merge so I can continue testing the integration with the rest of the feature.

PR #135, opened 2020-09-11
Add customization flags for better OpenShift support

enisoc:
This makes various defaults that were previously hard-coded configurable via command-line flags. The default defaults remain the same as they were before if the new flags are left unset.

These particular values are ones that we've found may need to be overridden when running in OpenShift as opposed to pure Kubernetes.

### DIFF
--- a/pkg/apis/planetscale/v2/defaults.go
+++ b/pkg/apis/planetscale/v2/defaults.go
@@ -134,12 +134,14 @@ var (
 
 	// DefaultVitessRunAsUser is the UID to use by default for Pods that run
 	// Vitess components. This value can be configured at operator startup time
-	// with the --default_vitess_run_as_user flag.
+	// with the --default_vitess_run_as_user flag. A value less than 0 means
+	// don't set RunAsUser at all.
 	DefaultVitessRunAsUser int64 = 999
 
 	// DefaultVitessFSGroup is the GID to use by default for Pods that run
 	// Vitess components. This value can be configured at operator startup time
-	// with the --default_vitess_fs_group flag.
+	// with the --default_vitess_fs_group flag. A value less than 0 means don't
+	// set FSGroup at all.
 	DefaultVitessFSGroup int64 = 999
 
 	// DefaultEtcdServiceAccount is the name of the ServiceAccount to use by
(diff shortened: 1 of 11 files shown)


pkg/operator/etcd/pod.go
```diff
@@ -226,6 +235,17 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	obj.Spec.Subdomain = PeerServiceName(spec.LockserverName)
 	obj.Spec.ImagePullSecrets = spec.ImagePullSecrets
 
+	if planetscalev2.DefaultEtcdFSGroup > 0 {
+		if obj.Spec.SecurityContext == nil {
+			obj.Spec.SecurityContext = &corev1.PodSecurityContext{}
+		}
+		obj.Spec.SecurityContext.FSGroup = pointer.Int64Ptr(planetscalev2.DefaultEtcdFSGroup)
```
PrismaPhonic, line 242: Is zero not a valid security context? Why not set `obj.Spec.SecurityContext.FSGroup` to a pointer of `DefaultEtcdFSGroup` no matter what?

PrismaPhonic (commented):
LGTM for the most part. Have one open question.

pkg/apis/planetscale/v2/defaults.go
```diff
@@ -131,3 +120,45 @@ var defaultVitessImages = &VitessImages{
 
 	MysqldExporter: "prom/mysqld-exporter:v0.11.0",
 }
+
+var (
```
avaidyanatha, line 124: I didn't know we assigned a priority class, do we evict non-Vitess related pods on scheduling failure?

avaidyanatha (approved):
Looks good.

pkg/apis/planetscale/v2/defaults.go
enisoc: Yeah since the beginning we've been including a PriorityClass with the operator distro with a default priority value. Depending on the priority of other Pods in your system, this could cause Vitess Pods to preempt them. That'll still be the case by default, but at least with these new flags you can disable it.

pkg/operator/etcd/pod.go
enisoc: Since `FSGroup` is a pointer, there's a difference between 0 and nil, and I want the default default to remain nil as it was before.

I didn't see a way to tell the flag parser to leave the value nil by default, but maybe I should avoid using 0 to mean nil since 0 is a valid value. I'll change it to treat 0 as valid and `<0` as nil. I'll do the same for `RunAsUser`.

PrismaPhonic: Ok great. That was my thought - that a user could specify 0, so it should be treated as a valid value rather than assuming that means unset.

PrismaPhonic (approved):
LGTM

PR #136, opened 2020-09-16
Vtop: Refactor Conditions to List for Kubectl Wait

PrismaPhonic:
This PR refactors conditions to be a list rather than map so they can be used with kubectl wait command.

### DIFF
--- a/pkg/apis/planetscale/v2/vitesskeyspace_methods.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_methods.go
@@ -167,11 +167,7 @@ func (p *VitessKeyspacePartitioning) TabletPools() []VitessShardTabletPool {
 // newStatus, then we do not update LastTransitionTime. However, if newStatus is different from current status, then
 // we update the status and update the transition time.
 func (s *VitessKeyspaceStatus) SetConditionStatus(condType VitessKeyspaceConditionType, newStatus corev1.ConditionStatus, reason, message string) {
-	if s.Conditions == nil {
-		s.Conditions = make(map[VitessKeyspaceConditionType]VitessKeyspaceCondition)
-	}
-
-	cond, ok := s.Conditions[condType]
+	cond, ok := s.GetCondition(condType)
 	if !ok {
 		cond = NewVitessKeyspaceCondition()
 	}
@@ -186,13 +182,13 @@ func (s *VitessKeyspaceStatus) SetConditionStatus(condType VitessKeyspaceConditi
 		cond.LastTransitionTime = &now
 	}
 
-	s.Conditions[condType] = cond
+	s.setCondition(cond)
 }
 
 // NewVitessKeyspaceCondition returns an init VitessKeyspaceCondition object.
-func NewVitessKeyspaceCondition() VitessKeyspaceCondition {
+func NewVitessKeyspaceCondition() *VitessKeyspaceCondition {
 	now := metav1.NewTime(time.Now())
-	return VitessKeyspaceCondition{
+	return &VitessKeyspaceCondition{
 		Status:             corev1.ConditionUnknown,
 		LastTransitionTime: &now,
 		Reason:             "",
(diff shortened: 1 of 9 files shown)


pkg/apis/planetscale/v2/vitesskeyspace_methods.go
```diff
@@ -212,15 +208,39 @@ func (c *VitessKeyspaceCondition) StatusDuration() time.Duration {
...
+	for _, condition := range s.Conditions {
+		out = append(out, *condition.DeepCopy())
 	}
 
 	return out
 }
+
+// GetCondition provides map style access to retrieve a condition from the conditions list by it's type
+// If the condition doesn't exist, we return nil, false.
+func (s *VitessKeyspaceStatus) GetCondition(ty VitessKeyspaceConditionType) (value *VitessKeyspaceCondition, exists bool) {
```
enisoc, line 223: I think it may be bug-prone to have an exported method that returns a pointer into the underlying array. Callers may think it's ok to retain the pointer, but that's not safe in general. Either the caller's value might get unintentionally mutated underneath them, or the caller might intend to mutate the value in the array and fail.

Consider this scenario:

```go
ptr := GetCondition(condType)
SetConditionStatus(otherCondType, ...)

// This may or may not update the value in the array,
// depending on whether SetConditionStatus() above
// allocated a new array while appending to the slice.
ptr.Status = corev1.ConditionTrue
```

I think it'll be useful to have an exported `GetCondition()` method, but it should return a deep copy to be safe. We could convert this version to an unexported method `getCondition()` that's used by both `SetConditionStatus()` and `GetCondition()`.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -300,12 +300,13 @@ type VitessKeyspaceStatus struct {
...
 	Resharding *ReshardingStatus `json:"resharding,omitempty"`
-	// FIXME: Convert this to a list for kubectl wait compatibility before merging the feature branch to master.
-	// Conditions is a map of all VitessKeyspace specific conditions we want to set and monitor.
+	// Conditions is a list of all VitessKeyspace specific conditions we want to set and monitor.
 	// It's ok for multiple controllers to add conditions here, and those conditions will be preserved.
-	Conditions map[VitessKeyspaceConditionType]VitessKeyspaceCondition `json:"conditions,omitempty"`
+	Conditions []VitessKeyspaceCondition `json:"conditions,omitempty"`
 }
 
+type VitessKeyspaceConditions = []VitessKeyspaceCondition
```
enisoc, line 308: Is this alias used anywhere?

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -368,9 +368,9 @@ type VitessShardStatus struct {
 	// to turn down the shard.
 	Idle corev1.ConditionStatus `json:"idle,omitempty"`
 
-	// Conditions is a map of all VitessShard specific conditions we want to set and monitor.
+	// Conditions is a list of all VitessShard specific conditions we want to set and monitor.
 	// It's ok for multiple controllers to add conditions here, and those conditions will be preserved.
-	Conditions map[VitessShardConditionType]VitessShardCondition `json:"conditions,omitempty"`
+	Conditions []VitessShardCondition `json:"conditions,omitempty"`
```
enisoc, line 373: I think we should wait to change the existing conditions field on VitessShard. I'm worried that this change might break existing operator deployments upon upgrade due to validation/deserialization issues. Let's just make VitessKeyspace conditions start on the new format and we'll figure out the VitessShard migration separately so it doesn't block resharding work.

pkg/apis/planetscale/v2/vitesskeyspace_methods.go
```diff
@@ -212,15 +208,39 @@ func (c *VitessKeyspaceCondition) StatusDuration() time.Duration {
 }
 
 // DeepCopyConditions deep copies the conditions map for VitessKeyspaceStatus.
-func (s *VitessKeyspaceStatus) DeepCopyConditions() map[VitessKeyspaceConditionType]VitessKeyspaceCondition {
-	if s == nil {
-		return nil
-	}
-	out := make(map[VitessKeyspaceConditionType]VitessKeyspaceCondition, len(s.Conditions))
+func (s *VitessKeyspaceStatus) DeepCopyConditions() []VitessKeyspaceCondition {
+	out := make([]VitessKeyspaceCondition, len(s.Conditions))
```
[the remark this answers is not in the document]
PrismaPhonic: ahh right. I was being lazy and converting from map to list and forgot to adjust this.

pkg/apis/planetscale/v2/vitesskeyspace_methods.go
PrismaPhonic: Ahh, I see what you're saying. Ok, will adjust.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Oops, no. I was playing around with trying to see if there was a way to create a custom type with overrides for map style access, which isn't possible in Golang. Missed this artifact - will remove.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Ok sounds good. I'll revert these changes then.

pkg/apis/planetscale/v2/vitesskeyspace_methods.go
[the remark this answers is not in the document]
PrismaPhonic: Updated.

pkg/apis/planetscale/v2/vitesskeyspace_methods.go
PrismaPhonic: Adjusted.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Removed.

pkg/apis/planetscale/v2/vitessshard_types.go
PrismaPhonic: Reverted.

enisoc (approved):
LGTM

PR #138, opened 2020-09-18
vtgate: Don't hard-code the default gateway implementation.

enisoc:
We shouldn't set this at the operator level because the recommended default value of this flag is different for different Vitess versions. Instead, we should allow the default from upstream Vitess to take effect if the user doesn't set this flag explicitly.


PrismaPhonic (approved):
LGTM

PR #139, opened 2020-09-23
Don't set ImagePullPolicy if we weren't asked to.

### LINKED ISSUE #111: Hot loop when using AlwaysPullImages admission controller
We've received a report of a hot loop that occurs when these conditions are met:

- The vttablet image is set to use the `:latest` tag.
- The `updateStrategy` is set to `type: Immediate`.
- The `imagePullPolicies` field is left unset.

I will try to reproduce this. In the meantime, a potential workaround is to explicitly fill in [`imagePullPolicies`](https://vitess-operator.planetscale.dev/api/#planetscale.com/v2.VitessImagePullPolicies) in the VitessCluster spec, making sure to set a non-empty value for any Vitess component for which you are using a `:latest` image (vtctld, vtgate, vttablet, vtbackup, mysqld, mysqldExporter).

enisoc:
Fixes #111

### DIFF
--- a/pkg/operator/update/podspec.go
+++ b/pkg/operator/update/podspec.go
@@ -130,6 +130,14 @@ func PodContainer(dst, src *corev1.Container) {
 	dstVolumeMounts := dst.VolumeMounts
 	VolumeMounts(&dstVolumeMounts, src.VolumeMounts)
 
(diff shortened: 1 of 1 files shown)


PrismaPhonic (approved):
LGTM

PR #142, opened 2020-09-26
Resharding: Update VTK Status To Include Copy Progress

PrismaPhonic:
This PR updates VTK status when resharding to include current copy progress.

Reviewer notes: Can someone check my logic for calculating percentage? Right now I cast both source shard row counts and target shard row counts to floating points for division, but we could be losing information here potentially as we go from uint64 to float64. Do we think this could be problematic? Should we just rely on floor division instead?

### DIFF
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -318,6 +318,7 @@ type ReshardingStatus struct {
 	// CopyProgress will indicate the percentage completion ranging from 0-100 as integer values.
 	// Once we are past the copy phase, this value will always be 100, and will never be 100 while we
 	// are still within the copy phase.
+	// If we can not compute the copy progress in a timely fashion, we will report -1 to indicate the progress is unknown.
 	CopyProgress int `json:"copyProgress,omitempty"`
 }
 
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -315,6 +315,10 @@ type ReshardingStatus struct {
 	SourceShards []string `json:"sourceShards,omitempty"`
 	// TargetShards is a list of target shards for the current resharding operation.
 	TargetShards []string `json:"targetShards,omitempty"`
+	// CopyProgress will indicate the percentage completion ranging from 0-100 as integer values.
+	// Once we are past the copy phase, this value will always be 100, and will never be 100 while we
+	// are still within the copy phase.
+	CopyProgress int `json:"copyProgress,omitempty"`
 }
 
 // WorkflowState represents the current state for the given Workflow.
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -184,8 +184,7 @@ func (r *reconcileHandler) shardsRowCount(ctx context.Context, shardNames []stri
 		if err != nil {
 			return 0, fmt.Errorf("failed to get schema for shard %v: %v", shardName, err)
 		}
-		for i := range schema.TableDefinitions {
-			tabletDef := schema.TableDefinitions[i]
+		for _, tabletDef := range schema.TableDefinitions {
 			rowCount += tabletDef.GetRowCount()
 		}
 	}
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -24,7 +24,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"vitess.io/vitess/go/vt/proto/topodata"
 	"vitess.io/vitess/go/vt/wrangler"
 
 	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
@@ -174,22 +173,14 @@ func (r *reconcileHandler) percentCopied(ctx context.Context, sourceShards, targ
 func (r *reconcileHandler) shardsRowCount(ctx context.Context, shardNames []string) (uint64, error) {
 	var rowCount uint64
 	for _, shardName := range shardNames {
-		tabletMap, err := r.ts.GetTabletMapForShard(ctx, r.vtk.Spec.Name, shardName)
+		shardInfo, err := r.ts.GetShard(ctx, r.vtk.Spec.Name, shardName)
 		if err != nil {
 			return 0, fmt.Errorf("failed to get tablets for shard %v: %v", shardName, err)
 		}
-		// Find the master and get data size from it.
-		var masterTabletAlias *topodata.TabletAlias
-		for _, tabletInfo := range tabletMap {
-			if tabletInfo.Type == topodata.TabletType_MASTER {
-				masterTabletAlias = tabletInfo.Alias
-				break
-			}
-		}
-		if masterTabletAlias == nil {
+		if shardInfo.MasterAlias == nil {
 			return 0, fmt.Errorf("could not find master tablet alias for determining row count of shard %v", shardName)
 		}
-		schema, err := r.wr.GetSchema(ctx, masterTabletAlias, make([]string, 0), make([]string, 0), false)
+		schema, err := r.wr.GetSchema(ctx, shardInfo.MasterAlias, make([]string, 0), make([]string, 0), false)
 		if err != nil {
 			return 0, fmt.Errorf("failed to get schema for shard %v: %v", shardName, err)
 		}
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -180,7 +180,7 @@ func (r *reconcileHandler) shardsRowCount(ctx context.Context, shardNames []stri
 		if shardInfo.MasterAlias == nil {
 			return 0, fmt.Errorf("could not find master tablet alias for determining row count of shard %v", shardName)
 		}
-		schema, err := r.wr.GetSchema(ctx, shardInfo.MasterAlias, make([]string, 0), make([]string, 0), false)
+		schema, err := r.wr.GetSchema(ctx, shardInfo.MasterAlias, nil, nil, false)
 		if err != nil {
 			return 0, fmt.Errorf("failed to get schema for shard %v: %v", shardName, err)
 		}
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -153,6 +153,10 @@ func (r *reconcileHandler) percentCopied(ctx context.Context, sourceShards, targ
 		r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "CopyProgressUnknown", "failed to aggregate row count for source shards: %v", err)
 		return -1
 	}
+	if sourceRowCount == 0 {
+		// If sourceRowCount is zero, then we are just waiting for workflow to transition to Running.
+		return 99
+	}
 	// Aggregate row counts for all target shards.
 	targetRowCount, err := r.shardsRowCount(ctx, targetShards)
 	if err != nil {
--- a/pkg/controller/vitesskeyspace/reconcile_resharding.go
+++ b/pkg/controller/vitesskeyspace/reconcile_resharding.go
@@ -117,29 +117,15 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 		}
 	}
 
+	progressCtx, cancel := context.WithTimeout(ctx, topoReconcileTimeout)
+	defer cancel()
 	switch workflowStatus.State {
 	case planetscalev2.WorkflowError:
+		workflowStatus.CopyProgress = r.percentCopied(progressCtx, workflowStatus.SourceShards, workflowStatus.TargetShards)
 		sort.Strings(errorMsgs)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Error", fmt.Sprintf("VReplication reported an error: %v", errorMsgs[0]))
 	case planetscalev2.WorkflowCopying:
-		// Aggregate row counts for all source shards.
-		sourceRowCount, err := r.shardsRowCount(ctx, workflowStatus.SourceShards)
-		if err != nil {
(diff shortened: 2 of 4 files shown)


pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -120,8 +122,26 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 		sort.Strings(errorMsgs)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Error", fmt.Sprintf("VReplication reported an error: %v", errorMsgs[0]))
 	case planetscalev2.WorkflowCopying:
+		// Aggregate tablet maps for all source shards.
+		sourceRowCount, err := r.shardsRowCount(ctx, workflowStatus.SourceShards)
+		if err != nil {
+			r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "SourceShardsRowCountFailed", "failed to aggregate row count for source shards: %v", err)
```
enisoc, line 128: I think the Reason string is more useful when we use it to group problems into high-level categories, rather than uniquely specifying exactly which case we hit. The Message string tells you the details. How about something like this?

```suggestion
			r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "CopyProgressUnknown", "failed to aggregate row count for source shards: %v", err)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -120,8 +122,26 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+			return resultBuilder.RequeueAfter(topoRequeueDelay)
+		}
+		targetRowCount, err := r.shardsRowCount(ctx, workflowStatus.TargetShards)
+		if err != nil {
+			r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "TargetShardsRowCountFailed", "failed to aggregate row count for target shards: %v", err)
```
enisoc, line 133:
```suggestion
			r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "CopyProgressUnknown", "failed to aggregate row count for target shards: %v", err)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -120,8 +122,26 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+			return resultBuilder.RequeueAfter(topoRequeueDelay)
+		}
+		percentComplete := int(math.Floor((float64(targetRowCount) / float64(sourceRowCount)) * 100.0))
```
enisoc, line 136: > we could be losing information here potentially as we go from uint64 to float64. Do we think this could be problematic?

I don't think it's a problem. Row counts up to 2^53 - 1 will be represented exactly in a float64, and even above that, the error as a percentage will almost never be enough to change the truncated result.

> Should we just rely on floor division instead?

Integer arithmetic would work too, although it technically risks overflow in the unlikely event that the row count is close to 2^64, since you'd have to multiply by 100 before dividing.

My only suggestion here is that there's no need to call `Floor()` if we're about to convert to an integer anyway:

```suggestion
		percentComplete := int(float64(targetRowCount) / float64(sourceRowCount) * 100.0)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -120,8 +122,26 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 		sort.Strings(errorMsgs)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Error", fmt.Sprintf("VReplication reported an error: %v", errorMsgs[0]))
 	case planetscalev2.WorkflowCopying:
+		// Aggregate tablet maps for all source shards.
+		sourceRowCount, err := r.shardsRowCount(ctx, workflowStatus.SourceShards)
+		if err != nil {
+			r.recorder.Eventf(r.vtk, corev1.EventTypeWarning, "SourceShardsRowCountFailed", "failed to aggregate row count for source shards: %v", err)
+			return resultBuilder.RequeueAfter(topoRequeueDelay)
+		}
```
enisoc, line 130: We should also check if `sourceRowCount == 0` so we don't divide by zero later. I guess in that case we can just return `99` since the copy is basically done and we're just waiting for the workflow to transition to Running.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+		// Find the master and get data size from it.
+		var masterTabletAlias *topodata.TabletAlias
+		for _, tabletInfo := range tabletMap {
+			if tabletInfo.Type == topodata.TabletType_MASTER {
+				masterTabletAlias = tabletInfo.Alias
+				break
+			}
+		}
+		if masterTabletAlias == nil {
+			return 0, fmt.Errorf("could not find master tablet alias for determining row count")
```
enisoc, line 176:
```suggestion
			return 0, fmt.Errorf("could not find master tablet alias for determining row count of shard %v", shardName)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+		}
+		schema, err := r.wr.GetSchema(ctx, masterTabletAlias, make([]string, 0), make([]string, 0), false)
+		if err != nil {
+			return 0, err
```
enisoc, line 180:
```suggestion
			return 0, fmt.Errorf("failed to get schema for shard %v: %v", shardName, err)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 
 	return resultBuilder.Result()
 }
+
+func (r reconcileHandler) shardsRowCount(ctx context.Context, shardNames []string) (uint64, error) {
+	var rowCount uint64
+	for _, shardName := range shardNames {
+		tabletMap, err := r.ts.GetTabletMapForShard(ctx, r.vtk.Spec.Name, shardName)
+		if err != nil {
+			return 0, err
```
enisoc, line 165: Let's include the loop var in all errors so we know which one failed:

```suggestion
			return 0, fmt.Errorf("failed to get tablets for shard %v: %v", shardName, err)
```

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 
 	return resultBuilder.Result()
 }
+
+func (r reconcileHandler) shardsRowCount(ctx context.Context, shardNames []string) (uint64, error) {
+	var rowCount uint64
+	for _, shardName := range shardNames {
+		tabletMap, err := r.ts.GetTabletMapForShard(ctx, r.vtk.Spec.Name, shardName)
```
enisoc, line 163: Since we're only interested in the master, we could just call GetShard and find the master_alias there. That way we don't have to fetch all the non-master tablet records.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+		for _, tabletInfo := range tabletMap {
+			if tabletInfo.Type == topodata.TabletType_MASTER {
+				masterTabletAlias = tabletInfo.Alias
+				break
+			}
+		}
+		if masterTabletAlias == nil {
+			return 0, fmt.Errorf("could not find master tablet alias for determining row count")
+		}
+		schema, err := r.wr.GetSchema(ctx, masterTabletAlias, make([]string, 0), make([]string, 0), false)
```
enisoc, line 178: Do we really need to make string arrays, or can we just pass `nil, nil`? If we do need to allocate something, the more common way to make an empty string slice is `[]string{}`.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -136,3 +156,33 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
+		if err != nil {
+			return 0, err
+		}
+		for i := range schema.TableDefinitions {
+			tabletDef := schema.TableDefinitions[i]
+			rowCount += tabletDef.GetRowCount()
+		}
```
enisoc, line 185:
```suggestion
		for _, tableDef := range schema.TableDefinitions {
			rowCount += tableDef.GetRowCount()
		}
```

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -315,6 +315,10 @@ type ReshardingStatus struct {
 	SourceShards []string `json:"sourceShards,omitempty"`
 	// TargetShards is a list of target shards for the current resharding operation.
 	TargetShards []string `json:"targetShards,omitempty"`
+	// CopyProgress will indicate the percentage completion ranging from 0-100 as integer values.
```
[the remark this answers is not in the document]
PrismaPhonic: I was thinking about this too. That being said I would still say that if we are past the copy phase we should report 100% and not even attempt to check.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Sounds good to me

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Ahhh, good catch. Will update.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Good point

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: wasn't sure if you can call `len()` on nil, but just tried in go playground and you can. Refactoring to `nil, nil`.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -120,8 +122,26 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
 		sort.Strings(errorMsgs)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Error", fmt.Sprintf("VReplication reported an error: %v", errorMsgs[0]))
 	case planetscalev2.WorkflowCopying:
+		// Aggregate tablet maps for all source shards.
```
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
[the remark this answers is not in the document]
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Updated.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_resharding.go
PrismaPhonic: Done.

enisoc (approved):
LGTM after one minor fix. Conditionally approved.

pkg/controller/vitesskeyspace/reconcile_resharding.go
```diff
@@ -115,13 +115,18 @@ func (r *reconcileHandler) reconcileResharding(ctx context.Context) (reconcile.R
...
 	switch workflowStatus.State {
 	case planetscalev2.WorkflowError:
+		workflowStatus.CopyProgress = r.percentCopied(progressCtx, workflowStatus.SourceShards, workflowStatus.TargetShards)
 		sort.Strings(errorMsgs)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Error", fmt.Sprintf("VReplication reported an error: %v", errorMsgs[0]))
 	case planetscalev2.WorkflowCopying:
+		workflowStatus.CopyProgress = r.percentCopied(progressCtx, workflowStatus.SourceShards, workflowStatus.TargetShards)
 		r.setConditionStatus(planetscalev2.VitessKeyspaceReshardingInSync, corev1.ConditionFalse, "Copying", "Existing data from the source shards is being backfilled on target shards")
 	case planetscalev2.WorkflowRunning:
+		workflowStatus.CopyProgress = 100
```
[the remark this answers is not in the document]
PrismaPhonic: That's a good point. I'll initialize it at -1 when I first create the workflow status.

PR #143, opened 2020-09-29
KeyspaceStatus: Include Partitioning Readiness

PrismaPhonic:
This PR expands the partitioning status section to include readiness information and provides a high-level condition to assess whether the partitioning that is serving writes has a ready tablet count that matches its desired tablet count.

### DIFF
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -462,7 +462,7 @@ const (
 	// VitessKeyspaceReshardingInSync indicates whether the keyspace has an active
 	// resharding operation whose target shards are ready to serve if traffic is switched.
 	VitessKeyspaceReshardingInSync VitessKeyspaceConditionType = "ReshardingInSync"
-	// VitessKeyspaceReady indicates whether the keyspace's serving partitionings tablets are all ready.
+	// VitessKeyspaceReady indicates whether the tablet Pods of the keyspace's serving partitioning are all Ready.
 	VitessKeyspaceReady VitessKeyspaceConditionType = "Ready"
 )
 
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -414,11 +414,20 @@ type VitessKeyspacePartitioningStatus struct {
 	// UpdatedTablets is the number of desired tablets that are up-to-date
 	// (have no pending changes).
 	UpdatedTablets int32 `json:"updatedTablets,omitempty"`
+	// DesiredShards is the number of desired shards. This is computed from
+	// information that's already available in the spec, but clients should
+	// use this value instead of trying to compute shard partitionings on their
+	// own.
+	DesiredShards int32 `json:"desiredShards,omitempty"`
+	// ReadyShards is the number of desired shards that are Ready.
+	ReadyShards int32 `json:"readyShards,omitempty"`
 }
 
 // NewVitessKeyspacePartitioningStatus creates a new status object with default values.
 func NewVitessKeyspacePartitioningStatus(partitioning *VitessKeyspacePartitioning) VitessKeyspacePartitioningStatus {
 	var desiredTablets int32
+	shards := partitioning.ShardNameSet()
+	desiredShards := int32(shards.Len())
 
 	tabletPools := partitioning.TabletPools()
 	for tpIndex := range tabletPools {
@@ -429,6 +438,7 @@ func NewVitessKeyspacePartitioningStatus(partitioning *VitessKeyspacePartitionin
 		ShardNames:     partitioning.ShardNameSet().List(),
 		ServingWrites:  corev1.ConditionUnknown,
 		DesiredTablets: desiredTablets,
+		DesiredShards:  desiredShards,
 	}
 }
 
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -401,13 +401,33 @@ type VitessKeyspacePartitioningStatus struct {
 	// possible that some shards in this partitioning are serving writes.
 	// Check the per-shard status for full details.
 	ServingWrites corev1.ConditionStatus `json:"servingWrites,omitempty"`
+	// DesiredTablets is the number of desired tablets. This is computed from
+	// information that's already available in the spec, but clients should
+	// use this value instead of trying to compute shard partitionings on their
+	// own.
+	DesiredTablets int32 `json:"desiredTablets,omitempty"`
+	// Tablets is the number of observed tablets. This could be higher or
+	// lower than desiredTablets if the state has not yet converged.
+	Tablets int32 `json:"tablets,omitempty"`
+	// ReadyTablets is the number of desired tablets that are Ready.
+	ReadyTablets int32 `json:"readyTablets,omitempty"`
+	// UpdatedTablets is the number of desired tablets that are up-to-date
+	// (have no pending changes).
+	UpdatedTablets int32 `json:"updatedTablets,omitempty"`
 }
 
 // NewVitessKeyspacePartitioningStatus creates a new status object with default values.
 func NewVitessKeyspacePartitioningStatus(partitioning *VitessKeyspacePartitioning) VitessKeyspacePartitioningStatus {
+	desiredTablets := int32(0)
+	tabletPools := partitioning.TabletPools()
+	for tpIndex := range tabletPools {
+		desiredTablets += tabletPools[tpIndex].Replicas
+	}
+
 	return VitessKeyspacePartitioningStatus{
-		ShardNames:    partitioning.ShardNameSet().List(),
-		ServingWrites: corev1.ConditionUnknown,
+		ShardNames:     partitioning.ShardNameSet().List(),
+		ServingWrites:  corev1.ConditionUnknown,
+		DesiredTablets: desiredTablets,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -401,13 +401,33 @@ type VitessKeyspacePartitioningStatus struct {
...
+	// ReadyTablets is the number of desired tablets that are Ready.
+	ReadyTablets int32 `json:"readyTablets,omitempty"`
+	// UpdatedTablets is the number of desired tablets that are up-to-date
+	// (have no pending changes).
+	UpdatedTablets int32 `json:"updatedTablets,omitempty"`
 }
 
 // NewVitessKeyspacePartitioningStatus creates a new status object with default values.
 func NewVitessKeyspacePartitioningStatus(partitioning *VitessKeyspacePartitioning) VitessKeyspacePartitioningStatus {
+	desiredTablets := int32(0)
```
avaidyanatha, line 421: Nit: The zero value is 0, so this is fine right?
```suggestion
	var desiredTablets int32
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -187,6 +199,48 @@ func allShardsServingWrites(shardNames []string, shardStatus map[string]planetsc
 	return result
 }
 
+func totalReadytablets(shardNames []string, shardStatus map[string]planetscalev2.VitessKeyspaceShardStatus) int32 {
```
avaidyanatha, line 202: My initial thought was that this looks copypasta-ish and all of this logic could sit in one subroutine or method? So you could call something like `updateTabletCounts()` to set all the values in one place.

Either way, this isn't that important to me and I won't block on it, just let me know what you feel about it.

avaidyanatha (approved):
LGTM, two thoughts.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: I was copying logic from another constructor function above in the same file. But I believe you're right on that.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: That's true to some extent but would require passing in the thing we want to mutate. I tend to avoid doing that because it can lead to buggy code that's unclear. I find it better usually to design pure functions that don't mutate data and return aggregated results. I agree it's less efficient, but I don't personally believe that two functions that do similar things are always bad.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Updated here, and also updated the place I copied it from.

pkg/controller/vitesskeyspace/reconcile_shards.go
avaidyanatha: I'm totally cool with keeping it how it is, just wanted to throw that option out there just in case you had thought of it yourself and were on the edge of making the change or not :)

pkg/apis/planetscale/v2/vitesskeyspace_types.go
avaidyanatha: Cool, thanks!

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -441,6 +462,8 @@ const (
 	// VitessKeyspaceReshardingInSync indicates whether the keyspace has an active
 	// resharding operation whose target shards are ready to serve if traffic is switched.
 	VitessKeyspaceReshardingInSync VitessKeyspaceConditionType = "ReshardingInSync"
+	// VitessKeyspaceReady indicates whether the keyspace's serving partitionings tablets are all ready.
```
enisoc, line 465:
```suggestion
	// VitessKeyspaceReady indicates whether the tablet Pods of the keyspace's serving partitioning are all Ready.
```

pkg/controller/vitesskeyspace/reconcile_shards.go
enisoc, line 202:
```suggestion
func totalReadyTablets(shardNames []string, shardStatus map[string]planetscalev2.VitessKeyspaceShardStatus) int32 {
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -152,6 +153,17 @@ func (r *reconcileHandler) reconcileShards(ctx context.Context) error {
 	for i := range r.vtk.Status.Partitionings {
 		status := &r.vtk.Status.Partitionings[i]
 		status.ServingWrites = allShardsServingWrites(status.ShardNames, r.vtk.Status.Shards)
+		status.Tablets = totalTablets(status.ShardNames, r.vtk.Status.Shards)
+		status.ReadyTablets = totalReadytablets(status.ShardNames, r.vtk.Status.Shards)
+		status.UpdatedTablets = totalUpdatedTablets(status.ShardNames, r.vtk.Status.Shards)
+
+		if status.ServingWrites == corev1.ConditionTrue {
+			if status.ReadyTablets != status.DesiredTablets {
+				r.setConditionStatus(planetscalev2.VitessKeyspaceReady, corev1.ConditionFalse, "NonReadyServingTablets", fmt.Sprintf("Desired serving tablet count: %d; Ready serving tablet count: %d", status.DesiredTablets, status.ReadyTablets))
```
enisoc, line 162: We don't need to update the condition message whenever the particular values change. The status values themselves give that info. Also "serving tablets" seems misleading; what we really mean is that the tablets are in the partitioning that's supposed to be serving, but the tablets themselves are not serving (because they aren't ready). Let's just keep the Reason string simple.

```suggestion
				r.setConditionStatus(planetscalev2.VitessKeyspaceReady, corev1.ConditionFalse, "TabletsNotReady", "Some tablet Pods for the serving partitioning are not Ready.")
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -152,6 +153,17 @@ func (r *reconcileHandler) reconcileShards(ctx context.Context) error {
...
+			} else {
+				r.setConditionStatus(planetscalev2.VitessKeyspaceReady, corev1.ConditionTrue, "ServingTabletsReady", "All serving tablets for keyspace are ready")
```
enisoc, line 164:
```suggestion
				r.setConditionStatus(planetscalev2.VitessKeyspaceReady, corev1.ConditionTrue, "TabletsReady", "All tablet Pods for the serving partitioning are Ready")
```

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -152,6 +153,17 @@ func (r *reconcileHandler) reconcileShards(ctx context.Context) error {
 	for i := range r.vtk.Status.Partitionings {
 		status := &r.vtk.Status.Partitionings[i]
 		status.ServingWrites = allShardsServingWrites(status.ShardNames, r.vtk.Status.Shards)
+		status.Tablets = totalTablets(status.ShardNames, r.vtk.Status.Shards)
+		status.ReadyTablets = totalReadytablets(status.ShardNames, r.vtk.Status.Shards)
+		status.UpdatedTablets = totalUpdatedTablets(status.ShardNames, r.vtk.Status.Shards)
+
+		if status.ServingWrites == corev1.ConditionTrue {
```
enisoc, line 160: If we get through the loop without finding any serving partitioning, we should set the Ready condition status to False with reason NoServingPartitioning. Otherwise it'll be left at Unknown/ReconcileFailed which doesn't represent what we actually know in that case.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -187,6 +199,48 @@ func allShardsServingWrites(shardNames []string, shardStatus map[string]planetsc
 	return result
 }
 
+func totalReadytablets(shardNames []string, shardStatus map[string]planetscalev2.VitessKeyspaceShardStatus) int32 {
+	// A partitioning with no shards has no ready tablets.
+	if len(shardNames) == 0 {
+		return 0
+	}
```
enisoc, line 206: This doesn't need to be special case. The loop will just iterate 0 times.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Done.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Removed from all.

pkg/controller/vitesskeyspace/reconcile_shards.go
```diff
@@ -187,6 +204,45 @@ func allShardsServingWrites(shardNames []string, shardStatus map[string]planetsc
...
+
+	return tabletCount
+}
+
+func totalReadyShards(shardNames []string, shardStatus map[string]planetscalev2.VitessKeyspaceShardStatus) int32 {
+	var readyShards int32
+	for _, shardName := range shardNames {
+		shard := shardStatus[shardName]
+		if shard.ReadyTablets == shard.Tablets {
```
enisoc, line 238: I think we should look at DesiredTablets so we don't think everything is fine if a tablet is missing. It looks like I got this wrong myself in `vitesscluster/reconcile_keyspaces.go`. Can you fix it there too?

```suggestion
		if shard.ReadyTablets == shard.DesiredTablets {
```

enisoc (approved):
LGTM if you agree on the one minor correction. Conditionally approved.

pkg/controller/vitesskeyspace/reconcile_shards.go
PrismaPhonic: Good point. I'll fix it in both spots!

PrismaPhonic: Done.

PR #145, opened 2020-09-30
docs/api: Fix links to k8s docs.

enisoc:
The v1.13 docs were removed. We should have updated to v1.16 anyway when we updated our operator-sdk version.


PrismaPhonic (approved):
LGTM

PR #146, opened 2020-09-30
Document compatibility policy.

### LINKED ISSUE #140: Document in README and websites the supported/recommended Kubernetes version
According to @enisoc on Slack:

> the recommendation of k8s 1.14 matches the pinned 2.0.0 version of the operator that the vitess docs currently use. if you use the latest operator from master, we recommend k8s 1.16 or something +/- 1 minor version.
> 
> i.e. 1.15 through 1.17

I don't see this documented on the README, or on the Vitess or PlanetScale websites.

Please also document the support policy schedule for Kubernetes versions:

- When will a Kubernetes version start being supported?
- When will a Kubernetes version start being the primary recommendation / the version that the operator tests against.
- When will a Kubernetes version stop being supported? (e.g. stop acting on compatibility issues, actively tell people to avoid that version, and/or known incompatibilities may start being introduced)
- For the clouds that you explicitly support (GKE, EKS, etc.), will you guarantee that you support at least one Kubernetes version that overlaps with the versions they offer?

enisoc:
Closes #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,41 @@
 - [Getting Started on GCP](https://docs.planetscale.com/vitess-operator/gcp-quickstart)
 - [VitessCluster CRD API Reference](https://docs.planetscale.com/vitess-operator/api)
 
+## Compatibility
+
+Vitess Operator depends on Vitess libraries and Kubernetes libraries that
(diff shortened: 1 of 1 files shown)


PrismaPhonic (approved):
LGTM

PR #147, opened 2020-10-01
Add an operator.yaml helper

mdkent:
This adds a quick helper to generate a current version of https://storage.googleapis.com/vitess-operator/install/operator.yaml - useful when doing local builds.

There's probably a better way to do this but I needed something to help with the testing of latest.


Makefile
```diff
@@ -36,6 +36,9 @@ generate:
 	find deploy/crds -name '*.yaml' | xargs go run ./cmd/trim-crd
 	go run github.com/ahmetb/gen-crd-api-reference-docs -api-dir ./pkg/apis -config docs/api/config.json -template-dir docs/api/template -out-file docs/api/index.html
 
+generate-operator-yaml:
+	go run github.com/kubernetes-sigs/kustomize build ./deploy > operator.yaml
```
enisoc, line 40: I like the idea, but could we place the file in `build/_output/operator.yaml` so we keep all our git-ignored build artifacts in one directory tree?

mdkent: For sure! That'd work.

enisoc (approved):
LGTM Thanks!

PR #148, opened 2020-10-02
Fix desiredTablets count in partitioning status.

enisoc:
We forgot to multiply by the number of shards in the case of an equal partitioning.


PrismaPhonic (approved):
LGTM

PR #149, opened 2020-10-05
Reset ReshardingInSync to Unknown when Workflow Names Change.

PrismaPhonic:
Added reset of resharding in sync in case where new workflow name doesn't match old workflow name. We can't assume in sync last transition time applies to both.


enisoc (approved):
LGTM

PR #150, opened 2020-10-07
Add extra label to PVC

emitamago:
Adding new extra company/account label to pvc

### DIFF
--- a/pkg/operator/vttablet/pvc.go
+++ b/pkg/operator/vttablet/pvc.go
@@ -26,15 +26,15 @@ import (
 // NewPVC creates a new vttablet PVC from a Spec.
 func NewPVC(key client.ObjectKey, spec *Spec) *corev1.PersistentVolumeClaim {
 	// Store labels in labels obj because we need to add extra label and avoid mutating spec.Labels value
-	Labels := map[string]string{}
-	update.Labels(&Labels, spec.Labels)
-	update.Labels(&Labels, spec.ExtraLabels)
+	labels := map[string]string{}
+	update.Labels(&labels, spec.Labels)
+	update.Labels(&labels, spec.ExtraLabels)
 
 	return &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: key.Namespace,
 			Name:      key.Name,
-			Labels:    Labels,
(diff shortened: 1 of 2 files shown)


pkg/operator/vttablet/pvc.go
```diff
@@ -24,7 +24,13 @@ import (
 )
 
 // NewPVC creates a new vttablet PVC from a Spec.
+// TODO: "Handle the case when labels are removed from ExtraLabels"
```
enisoc, line 27: We don't usually put quotes around the TODO comment. Also, let's move this to right before the point where the fix would need to go, which is line 49:

```go
	// update extra labels
	// TODO: Handle the case when labels are removed from ExtraLabels.
	update.Labels(&obj.Labels, spec.ExtraLabels)
```

pkg/operator/vttablet/pvc.go
```diff
@@ -24,7 +24,13 @@ import (
...
 func NewPVC(key client.ObjectKey, spec *Spec) *corev1.PersistentVolumeClaim {
+	// Store labels in labels obj because we need to add extra label and avoid mutating spec.Labels value
+	labels := map[string]string{}
```
enisoc, line 30: We should change `Labels: spec.Labels` below to use this var instead.

enisoc (commented):
Would you mind applying the same fix to `operator/etcd/pvc.go` as well? It should be the same change.

pkg/operator/vttablet/pvc.go
emitamago: moved to recommended place and removed quote

pkg/operator/vttablet/pvc.go
emitamago: sorry I do not why it was like this. fixed to use this var

emitamago (commented):
thank you so much for your review. 

question is that why you recommend to apply same fix to etcd/pvc.go as well?

enisoc:
> question is that why you recommend to apply same fix to etcd/pvc.go as well?

The operator manages etcd PVCs the same way it does vttablet PVCs, so we should be consistent in applying ExtraLabels in both cases.

enisoc (approved):
LGTM

PR #152, opened 2020-10-17
Reorder serviceaccount changes

bnu0:
The previous change (https://github.com/planetscale/vitess-operator/pull/151) left the wrong tokens mounted. This makes sure to remove them using the forked account name before substituting the new account.

@enisoc


enisoc (approved):
LGTM

PR #153, opened 2020-10-17
Allow configuring subPath for volume-based backup targets

bnu0:
When using a Volume as a backup target, which might be a clustered or other network filesystem, it is sometimes helpful to restrict the backups for a particular VitessCluster into a subfolder of this volume. For example, if multiple VitessClusters are using the same shared storage for backups, using subPath can avoid any risk of different clusters trampling (or seeing) each other's backups.

This makes it possible to set the `SubPath` for a `VitessBackupLocation`. I've tested it with glusterfs, which works great 👍

### DIFF
--- a/pkg/apis/planetscale/v2/vitessbackupstorage_types.go
+++ b/pkg/apis/planetscale/v2/vitessbackupstorage_types.go
@@ -85,6 +85,9 @@ type VitessBackupLocation struct {
 	// a shared host path for local testing.
 	// +kubebuilder:validation:EmbeddedResource
 	Volume *corev1.VolumeSource `json:"volume,omitempty"`
+	// VolumeSubPath gives the subpath in the volume to mount to the backups target.
+	// Only used for Volume-backed backup storage, ignored otherwise.
+	VolumeSubPath string `json:"volumeSubPath,omitempty"`
 	// Annotations can optionally be used to attach custom annotations to Pods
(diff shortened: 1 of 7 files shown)


pkg/apis/planetscale/v2/vitessbackupstorage_types.go
```diff
@@ -74,6 +74,9 @@ type VitessBackupLocation struct {
 	// +kubebuilder:validation:MaxLength=25
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([a-z0-9]*[a-z0-9])?$
 	Name string `json:"name,omitempty"`
+	// SubPath gives the subpath in the volume to mount to the backups target.
+	// Only used for Volume-backed backup storage, ignored otherwise.
+	SubPath     string `json:"subPath,omitempty"`
```
enisoc, line 79: Since this only applies to Volume-based backup locations, how about we call it `VolumeSubPath` to help signal that?

I really should have made `Volume` a struct with `Source *corev1.VolumeSource` inside it, so we could now add `SubPath`, but unfortunately I didn't think of it.

enisoc (commented):
Great idea!

bnu0:
@enisoc sorry it took me so long to circle back to this, i have updated it as you requested.

enisoc (approved):

PR #154, opened 2020-10-17
Include namespace in etcd client service address.

enisoc:
This allows the operator to still talk to the VitessCluster's etcd, even if the operator is in a different namespace.


PrismaPhonic (approved):
LGTM

PR #155, opened 2020-10-20
PsOp: Update External MySQL Defaults.

PrismaPhonic:
This PR updates the `enable_replication_reporter` flag to be false for externally managed MySQL. This has already bitten customers of ours twice, and we've decided this is a better default for this flag.


enisoc (approved):
LGTM

PR #158, opened 2020-10-28
Document WaitForFirstConsumer alternative to per-zone StorageClass

PrismaPhonic:
This PR Updates local lockserver suggestions in generated docs to include a newer and simpler approach.

### DIFF
--- a/pkg/apis/planetscale/v2/etcdlockserver_types.go
+++ b/pkg/apis/planetscale/v2/etcdlockserver_types.go
@@ -82,10 +82,9 @@ type EtcdLockserverTemplate struct {
 	// for each etcd instance to store its data files.
 	// This field is required.
 	//
-	// IMPORTANT: For a cell-local lockserver, you have two options. You either should set a storageClassName here for
-	// a StorageClass that's configured to only provision volumes in the Availability Zone that corresponds to the
-	// Vitess cell, or instead you may set `volumeBindingMode: WaitForFirstConsumer` in the StorageClass.
-	// This allows you to use the same StorageClass for any availability zone.
+	// IMPORTANT: For a cell-local lockserver in a Kubernetes cluster that spans
+	// multiple zones, you should ensure that `volumeBindingMode: WaitForFirstConsumer`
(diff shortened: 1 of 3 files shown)


pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -82,9 +82,10 @@ type EtcdLockserverTemplate struct {
 	// for each etcd instance to store its data files.
 	// This field is required.
 	//
-	// IMPORTANT: For a cell-local lockserver, you must set a storageClassName
-	// here for a StorageClass that's configured to only provision volumes in
-	// the Availability Zone that corresponds to the Vitess cell.
+	// IMPORTANT: For a cell-local lockserver, you have two options. You either should set a storageClassName here for
```
avaidyanatha, line 85:
```suggestion
	// IMPORTANT: For a cell-local lockserver, you have two options. Your first option is to set a storageClassName here for
```

pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -82,9 +82,10 @@ type EtcdLockserverTemplate struct {
...
+	// a StorageClass that's configured to only provision volumes in the Availability Zone that corresponds to the
+	// Vitess cell, or instead you may set `volumeBindingMode: WaitForFirstConsumer` in the StorageClass.
```
avaidyanatha, line 87: This just makes it slightly easier to read.

```suggestion
	// Vitess cell. Alternatively, you may set `volumeBindingMode: WaitForFirstConsumer` in the StorageClass.
```

avaidyanatha (commented):
Two comments, looks good otherwise.

enisoc:
I've pushed a comment to simplify this to only mentioning the new way. I'm not aware of any reason to use the old way except if you're on a super old k8s version, which we don't support anymore. I also applied the change to the tablet pool docs as well.

enisoc (approved):

PR #160, opened 2020-11-12
Support longer entity names with more character choices.

enisoc:
Cluster, keyspace, cell, and backup location names can now be longer and can contain more characters.

- VitessCluster names can be up to 63 chars and must match the regex `^[a-z0-9]([a-z0-9-.]*[a-z0-9])?$`.
- Keyspace and backup location names can be anything allowed as a k8s label value: up to 63 chars, matching the regex  `^[A-Za-z0-9]([A-Za-z0-9-_.]*[A-Za-z0-9])?$`.
- Cell names are similar, but cannot contain `-` because Vitess doesn't support `-` in cell names due to the tablet alias format.


PrismaPhonic (approved):
Nice work! LGTM. One nit.

pkg/operator/names/join_constraints_test.go
```diff
@@ -0,0 +1,166 @@
...
+		ValidFirstChar: isLowercaseAlphanumeric,
+	}
+	table := [][]string{
+		{"abc", "def"},
+		{"abc1", "2def3", "a-b-c"},
+		{"1abc", "2def3", "a-b-c"},
+		{"longnamepart1", "reallyreallyridiculouslylongnamepart2"},
+	}
+	for _, inputs := range table {
+		want := Join(inputs...)
```
avaidyanatha, line 136: Is the old `Join` function used elsewhere or does it just live here to test backwards compatibility?

avaidyanatha (approved):
Nice, looks good.

PR #166, opened 2020-12-23
Inject Affinities for Vtbackup-init Pod

PrismaPhonic:
This PR passes down affinities from the tablet spec to the vtbackup-init pod spec.

### DIFF
--- a/pkg/operator/vttablet/vtbackup_pod.go
+++ b/pkg/operator/vttablet/vtbackup_pod.go
@@ -179,6 +179,11 @@ func NewBackupPod(key client.ObjectKey, backupSpec *BackupSpec) *corev1.Pod {
 		pod.Spec.ServiceAccountName = planetscalev2.DefaultVitessServiceAccount
 	}
 
+	// Allow users to inject their own affinities.
+	if tabletSpec.Affinity != nil {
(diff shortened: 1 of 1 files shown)


pkg/operator/vttablet/vtbackup_pod.go
```diff
@@ -179,6 +179,11 @@ func NewBackupPod(key client.ObjectKey, backupSpec *BackupSpec) *corev1.Pod {
 		pod.Spec.ServiceAccountName = planetscalev2.DefaultVitessServiceAccount
 	}
 
+	// Allow users to inject their own affinities.
+	if tabletSpec.Affinity != nil {
+		pod.Spec.Affinity = tabletSpec.Affinity
```
enisoc, line 184: Since there is no default Affinity for vtbackup, and this is a New function (not Update on an existing Pod), you can just put `Affinity: tabletSpec.Affinity,` in the PodSpec literal above, like we do for `Tolerations`.

PrismaPhonic: Good point, moving.

PrismaPhonic: Done.

enisoc (approved):

PR #168, opened 2021-01-08
Fix: Update Deprecated RPCs

PrismaPhonic:
This PR updates `vtop` to no longer use the deprecated `SlaveStatus()` rpc call, and switch over to using the new `ReplicationStatus` rpc call instead. This PR also scrubs all usage of the word `slave` from `vtop` in comments and variable names.


deepthi (approved):
:100:

enisoc (approved):

PR #177, opened 2021-02-09
Fix: Sort Tolerations and Volume Names for Deterministic Hash

PrismaPhonic:
I wrote unit tests to catch a bug where we weren't deterministically sorting either tolerations or volumes which would cause a different hash if the order changed. I'm not certain it will fix this issue - we will have to try it out with a real deployment to see.

### DIFF
--- a/pkg/operator/contenthash/tolerations.go
+++ b/pkg/operator/contenthash/tolerations.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"sort"
 	"strconv"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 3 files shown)


enisoc (commented):
This would make sense if our goal was to ignore the ordering of these values, but I don't see a reason we would want to do that. In core k8s, I'm pretty sure changing the order of volumes or tolerations would be detected as a diff and would trigger a rollout for something like Deployment. That's part of the general rule of always assuming that order matters in any list in a k8s API. It's only maps where we can ignore order.

If you suspect there's an exception to the rule, you could do an experiment where you create a Deployment with tolerations and volumes, then edit the Deployment pod template to change the order and see if Deployment initiates a rollout. If it doesn't, then it would make sense for us to copy that behavior.

PrismaPhonic:
If the contents of the list are the same, why should order matter for tolerations or volumes? That seems odd to me to redeploy if the contents stay the same.

enisoc:
I don't know for sure whether the core k8s APIs trigger rollouts for reordered tolerations and volumes, but if they do, it seems like a reasonable decision to avoid having to understand every list as a special case. Since order matters for some lists, it's simpler to just assume that order matters for all lists.

jmoldow:
> we weren't deterministically sorting either tolerations or volumes

Can you elaborate on this?

Tolerations and volumes are user-defined lists, right? So what is not deterministic about that? Isn't it deterministic, namely "the exact order as specified by the user"?

> which would cause a different hash if the order changed

Ironically, would deploying this change not change the hash for users of this feature? As opposed to right now, where the hash would only change for users who explicitly choose to re-order their lists?

> it seems like a reasonable decision to avoid having to understand every list as a special case. Since order matters for some lists, it's simpler to just assume that order matters for all lists.

I would agree with this for anything that is a user-defined list, or a trivial extension of a user-defined list. But for any lists that are fully internal, wouldn't it make sense to sort them, if they aren't already? (e.g. AFAIK, when using an `equal` partition, the shard name have a deterministic order)

enisoc:
@jmoldow Great points, and I agree with your statement of what the general principle should be. We can close this since it was only an experiment while trying to debug an issue that we eventually tracked down to #178 instead.

PR #178, opened 2021-02-10
Fix: Copy ResourceRequirements so we aren't sharing pointers.

PrismaPhonic:
ResourcesRequirements internally holds maps, which are pointers. We found that this can create a bug if resource requirements gets mutated by another party. This PR tries to resolve that by copying resource requirements everywhere we set it to avoid shared memory issues.

### DIFF
--- a/pkg/operator/update/podspec.go
+++ b/pkg/operator/update/podspec.go
@@ -210,7 +210,6 @@ func ResourceList(dst, src *corev1.ResourceList) {
 	if *dst == nil {
 		// Allocate a new map to avoid sharing memory with src.
 		*dst = make(corev1.ResourceList)
-		return
 	}
 	for srcKey, srcVal := range *src {
 		(*dst)[srcKey] = srcVal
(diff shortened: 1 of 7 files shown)


pkg/operator/update/podspec.go
```diff
@@ -208,9 +208,8 @@ func ResourceRequirements(dst, src *corev1.ResourceRequirements) {
 // since those might be set by mutating admission webhooks or other controllers.
 func ResourceList(dst, src *corev1.ResourceList) {
 	if *dst == nil {
-		if len(*src) > 0 {
-			*dst = *src
-		}
+		// Allocate a new map to avoid sharing memory with src.
+		*dst = make(corev1.ResourceList)
```
enisoc, line 212: We need to also remove the `return` so we fall through to the copy phase.

enisoc (changes requested):

enisoc (approved):

pkg/operator/update/podspec.go
PrismaPhonic: Done.

PR #188, opened 2021-04-11
Bump example deployment to the newest v2.4.0 release tag.

aquarapid:

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: vitess-operator
       containers:
       - name: vitess-operator
-        image: planetscale/vitess-operator:v2.0.0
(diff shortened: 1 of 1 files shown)


askdba (changes requested):
I think the value needs to be set to `latest` instead of version-specific.
If an image needs to be pulled then manually edited this way no need to update every quarter

aquarapid:
> I think the value needs to be set to `latest` instead of version-specific.
> If an image needs to be pulled then manually edited this way no need to update every quarter

My opinion here is than pinning is best-practice for production;  and pulling latest can introduce unexpected changes;  and is a practice we want to discourage, unless you are a developer.

deepthi (approved):
OK, let's pin it.

PR #189, opened 2021-04-13
Remove excessive dependencies by upgrading to planetscale/tengo v3

shlomi-noach:
Matching https://github.com/vitessio/vitess/pull/7796, we upgrade to `planetscale/tengo` `v3`  where we removed docker-related code & dependencies.


askdba (approved):
LGTM

PR #190, opened 2021-04-14
Set a vtgate buffer size of 1000; default of 10 is likely way too small.

aquarapid:


deepthi (commented):
LGTM, but do we know what the memory impact will be?

deepthi (approved):

aquarapid:
Should not be more than a few megabytes for any reasonable workload.

PR #192, opened 2021-04-21
Support for VTTablet lifecycle hooks

artemvovk:
Adding support for container lifecycle hooks (https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) to VTTablet pods.

My main use case is to issue a `TabletExternallyReparented` command when scheduler decides to kill our primary tablets (unmanaged) - there could be other uses as well.

I'm not familiar with how integration tests are scoped, but I tested on my own cluster:

## Tests:

Screen Shot 2021-04-21 at 19 52 08

Logs from vtctld when above container is terminated:

Screen Shot 2021-04-21 at 20 11 30

Signed-off-by: Artem Vovk <avovk@recurly.com>

### DIFF
--- a/deploy/crds/planetscale.com_vitessshards.yaml
+++ b/deploy/crds/planetscale.com_vitessshards.yaml
@@ -505,6 +505,75 @@ spec:
                               type: string
                             type: object
                         type: object
+                      lifecycle:
+                        properties:
+                          postStart:
+                            properties:
+                              exec:
+                                properties:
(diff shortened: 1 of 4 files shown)


derekperkins (approved):

mattrobenolt (approved):
:+1:

mattrobenolt:
Thanks for the contribution @artemvovk.

PR #196, opened 2021-05-05
Bump version after v2.4.1 release

askdba:
Signed-off-by: Alkin Tezuysal <alkin.tezuysal@gmail.com>

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: vitess-operator
       containers:
       - name: vitess-operator
-        image: planetscale/vitess-operator:v2.4.0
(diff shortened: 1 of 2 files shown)


deploy/operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: vitess-operator
       containers:
       - name: vitess-operator
-        image: planetscale/vitess-operator:v2.4.0
+        image: planetscale/vitess-operator:v2.4.1
```
deepthi, line 19: This should also be v2.5.0?

askdba: No, we bump this to a release tag. Previously pinned here https://github.com/planetscale/vitess-operator/pull/188

PR #197, opened 2021-05-05
Relax consistency during vtbackup "catch up" phase.

aquarapid:
No reason to have strong consistency, since if there
are any errors during the backup, we should discard it
and start from scratch.

This allows fast(er) "catch up" from the binary logs, and
allow the backup to complete in a reasonable time, even
if the upstream write rate is high.


deepthi (approved):
LGTM

PR #203, opened 2021-06-18
Add kustomize as a direct go.mod dependency

jmoldow:
It is already installed as a transitive dependency and exists in
go.sum. But since it is used directly by `go run` in
`make generate-operator-yaml`, it is helpful to have it as a
direct dependency.

Signed-off-by: Jordan Moldow <jmoldow@box.com>


dctrwatson:
Please add it to this file too: https://github.com/planetscale/vitess-operator/blob/main/tools/tools.go

This way it won't be cleaned up by `go mod tidy`

jmoldow:
@dctrwatson Done! Thanks for calling that out.

dctrwatson (approved):
Thank you!

PR #205, opened 2021-06-18
Use github for two transitive dependencies

jmoldow:
The maintainer of github.com/fvbommel/util now recommends
including directly from that location.

On my corporate network, gomodules.xyz is blocked for some
reason. Since we're only downloading a single package from that
location, and since we're already downloading plenty of other
packages from github.com URLs, and since go should be able to
grab things from proxy.golang.org anyway, it seems fine to
replace the gomodules.xyz URL with the direct link to the github
repository. ¯\_(ツ)_/¯

Signed-off-by: Jordan Moldow <jmoldow@box.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,9 @@ require (
 // balancer.PickOptions from google.golang.org/grpc/balancer
 replace google.golang.org/grpc => google.golang.org/grpc v1.26.0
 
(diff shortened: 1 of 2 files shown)


dctrwatson (approved):
> The maintainer of github.com/fvbommel/util now recommends
> including directly from that location.

This has long been annoyance, so glad there's a fix forward for it.

Thank you!

PR #207, opened 2021-06-28
Adding TopologySpreadConstraints for VTTablets

artemvovk:
Reference:
https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#topologyspreadconstraint-v1-core

This adds support for managing/customizing more even spread of
VTTablets across kubernetes nodes
(in addition to affinity/tolerations).

Same feature is added to VTGates via `gateway` spec

## Test results from deploying:

- VTTablets

Screen Shot 2021-06-28 at 15 57 15

Screen Shot 2021-06-29 at 14 02 36

- VTGates

Screen Shot 2021-07-09 at 09 27 30

Signed-off-by: Artem Vovk <avovk@recurly.com>


deepthi:
What about older versions? We officially support 1.15, 1.16 and 1.17 and the docs say:
```
Note: In versions of Kubernetes before v1.18, you must enable the EvenPodsSpread feature gate on the API server and the scheduler in order to use Pod topology spread constraints.
```
Perhaps this is why CI tests failed.

artemvovk:
@deepthi Yeah, I assume it's OK with this field being optional? or: when is the plan to support 1.18 - 1.20 is GA for most PaaS providers.

aquarapid:
I think optional is fine, as long as we call that out, and it does not break existing PSC definitions on older versions.

This would be *really* useful for vtgate too;  since we don't have the built-in support for affinity there, since it sits behind a deployment.

artemvovk:
I can add it VTGates in follow up PR (to keep this small/isolated) or do you want it as part of this change?

aquarapid:
> I can add it VTGates in follow up PR (to keep this small/isolated) or do you want it as part of this change?

Together, if you don't mind.

artemvovk:
@aquarapid Done (updated PR description with VTGate screenshot)

pkg/apis/planetscale/v2/vitesscell_types.go
```diff
@@ -189,6 +189,11 @@ type VitessCellGatewaySpec struct {
 	// Tolerations allow you to schedule pods onto nodes with matching taints.
 	// +kubebuilder:validation:EmbeddedResource
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// TopologySpreadConstraint can optionally be used to
+	// specify how to spread vtgate pods among the given topology
+	// *kubebuilder:validation:EmbeddedResource
```
aquarapid, line 195: Does `*` vs `+` here have some special meaning I am not familiar with?

pkg/operator/vttablet/pod.go
```diff
@@ -269,6 +269,7 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	// Record a hash of desired tolerations to force the Pod to be recreated if
 	// one disappears from the desired list.
 	desiredStateHash.AddTolerations("tolerations", spec.Tolerations)
+	desiredStateHash.AddTopologySpreadConstraints("topologySpreadConstraints", spec.TopologySpreadConstraints)
```
aquarapid, line 272: Oh, I didn't understand this is how it worked.  Can we make a note in the auto-genned docs above that touching the tolerations or topologySpreadConstraints spec will restart the pod.

pkg/operator/vttablet/pod.go
```diff
@@ -287,6 +288,7 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	update.Volumes(&obj.Spec.Volumes, tabletVolumes.Get(spec))
 	update.Volumes(&obj.Spec.Volumes, spec.ExtraVolumes)
 	update.Tolerations(&obj.Spec.Tolerations, spec.Tolerations)
+	obj.Spec.TopologySpreadConstraints = spec.TopologySpreadConstraints
```
aquarapid, line 291: Why do we overwrite here vs update (as with the preceding code)?

aquarapid (changes requested):

pkg/operator/contenthash/topology_spread_constraints.go
```diff
@@ -0,0 +1,48 @@
...
+	h := md5.New()
+
+	for i := range in {
+		tsc := &in[i]
+
+		writeStringHash(h, string(tsc.MaxSkew))
+		writeStringHash(h, tsc.TopologyKey)
+		writeStringHash(h, string(tsc.WhenUnsatisfiable))
+		labelSelectors, err := metav1.LabelSelectorAsMap(tsc.LabelSelector)
+		if err == nil {
```
aquarapid, line 39: Should this not be `err != nil` ?

pkg/operator/vttablet/pod.go
artemvovk: It was the simplest/most straightforward way to add these - I am not aware of controllers or webhooks that would add their own `TopologySpreadConstraints` that we'd want to keep, but I can update to do a diff/merge similar to volumes/tolerations

artemvovk: I guess to compare/diff `TopologySpreadConstraints` - I'd ultimately have to compare their `LabelSelectors` and I am not sure if there's a sane way to do it without some edge cases

pkg/apis/planetscale/v2/vitesscell_types.go
artemvovk: from the docs - it should be `+`, so I fixed this one and the other stray `*`

https://book.kubebuilder.io/reference/markers/crd-validation.html

aquarapid (approved):
LGTM

bnu0 (approved):
This will be super helpful!

PR #208, opened 2021-06-29
Updates from make generate

jmoldow:
Signed-off-by: Jordan Moldow <jmoldow@alum.mit.edu>

### DIFF
--- a/deploy/crds/planetscale.com_vitessclusters.yaml
+++ b/deploy/crds/planetscale.com_vitessclusters.yaml
@@ -1222,6 +1222,109 @@ spec:
                                               additionalProperties:
                                                 type: string
                                               type: object
+                                            lifecycle:
+                                              properties:
+                                                postStart:
+                                                  properties:
+                                                    exec:
+                                                      properties:
+                                                        command:
+                                                          items:
+                                                            type: string
+                                                          type: array
+                                                      type: object
+                                                    httpGet:
+                                                      properties:
+                                                        host:
+                                                          type: string
+                                                        httpHeaders:
+                                                          items:
+                                                            properties:
+                                                              name:
+                                                                type: string
+                                                              value:
+                                                                type: string
+                                                            required:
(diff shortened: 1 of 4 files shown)


jmoldow:
The missing update in https://github.com/jmoldow/vitess-operator/blob/make-generate/pkg/apis/planetscale/v2/zz_generated.deepcopy.go#L2551 seems particularly problematic. That might be a bug in versions of the operator since that config was introduced. It'd be great to have all files kept up-to-date with each PR, but at least the yaml files being out-of-date won't cause the operator to misbehave, but out-of-date generate code might be fatal.

Could y'all consider adding a new Github workflow check, to prevent merges/releases if the generated files are out-of-date?

I've done something like this in some of my own projects that have code generation:

```bash
make generate
git add --all
git diff HEAD
echo 'If this test fails, it is because the git diff is non-empty after running "make generate".'
echo 'This indicates that either:'
echo ' - A source change has been made without running "make generate" and committing the corresponding generated code.'
echo ' - A manual change has been made to an auto-generated file, which is not supported or allowed.'
echo 'See the diff above for more information.'
echo 'To correct this, locally run "make generate", commit the changes, and re-run tests.'
git diff HEAD --quiet --exit-code
```

@dctrwatson @skottler 

I can submit that in a separate PR for your consideration.

jmoldow:
Added a proposal for that automated check in #209 .

skottler:
LGTM. Added @jordw for additional review.

skottler (approved):

jmoldow:
@skottler @jordw Can we merge this now?

jordw (approved):

jordw:
@jmoldow Sorry I was out of the office

jmoldow:
Thanks for approving @jordw . I don't have merge abilities, could someone do that for me? @skottler @jordw ?

jordw:
Done!

PR #213, opened 2021-07-23
Externalmaster failover

mavenraven:
s/already/always returns a node error from the first commit message

### DIFF
--- a/pkg/controller/vitessshardreplication/tablet_externally_reparent.go
+++ b/pkg/controller/vitessshardreplication/tablet_externally_reparent.go
@@ -53,14 +53,6 @@ func (r *ReconcileVitessShard) tabletExternallyReparent(ctx context.Context, vts
 	ctx, cancel := context.WithTimeout(ctx, externallyReparentTimeout)
 	defer cancel()
 
-	// Check actual shard record in case we are out of sync
-	// and bail if shard record says we have a master already.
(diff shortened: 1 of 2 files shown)


pkg/controller/vitessshardreplication/tablet_externally_reparent.go
```diff
@@ -42,9 +42,14 @@ func (r *ReconcileVitessShard) tabletExternallyReparent(ctx context.Context, vts
 		return resultBuilder.Result()
 	}
 
-	// If we already have a master we can bail early.
+	// If we already have a healthy master externalmaster we can bail early.
 	if vts.Status.HasMaster == corev1.ConditionTrue {
-		return resultBuilder.Result()
+		for _, tablet := range vts.Status.Tablets {
+			if tablet.IsExternalMaster() && tablet.Type == "master" && tablet.IsAvailable() {
```
deepthi, line 48: Should check Type against const instead of string value.
https://github.com/planetscale/vitess-operator/blob/5aa6228dc0d815eb05f4c5ca33e6b56bb8057f88/pkg/apis/planetscale/v2/vitessshard_types.go#L324

mavenraven: @deepthi that's not correct, it's not "externalmaster" but "master". Is there somehwere where table type constants already exist? I didn't see it.

PR #215, opened 2021-07-29
Adds readiness probe for mysqld container.

mavenraven:


aquarapid (changes requested):
We should probably adjust the probe default params (cf. https://pkg.go.dev/k8s.io/api/core/v1#Probe)

 * PeriodSeconds:  2 seconds instead of the default 10

rest can probably be left at defaults.

Otherwise, LGTM

aquarapid (approved):

PR #223, opened 2021-09-14
Upgrade to operator-sdk v0.18.0 and vitess

dctrwatson:
This requires updating vitess so that the k8s client wouldn't be downgraded. By updating vitess, this also forces the start of the migration to more inclusive terminology in the operator.


deepthi (approved):
LGTM

go.mod
```diff
@@ -4,32 +4,26 @@ go 1.15
...
+	k8s.io/kubectl v0.18.6
+	k8s.io/kubernetes v1.18.6
+	k8s.io/utils v0.0.0-20200603063816-c1c6865ac451
+	sigs.k8s.io/controller-runtime v0.6.4
+	sigs.k8s.io/controller-tools v0.3.0
 	sigs.k8s.io/kustomize v2.0.3+incompatible
-	sigs.k8s.io/yaml v1.1.0
-	vitess.io/vitess v0.0.0-20210504115542-f7304cd1893a
+	sigs.k8s.io/yaml v1.2.0
+	vitess.io/vitess v0.10.3-0.20210914162054-9df4dc751764
```
deepthi, line 24: Why is this `0.10.3`? I'd have expected `0.11.0`

dctrwatson: `go get vitess.io/vitess@HEAD` is what I ran

`go get vitess.io/vitess@main` got me `v0.11.1` but then that looks like not `HEAD` because it didn't have updated k8s client in it

askdba (approved):

go.mod
deepthi: that makes sense. We might have to fix something on the vitess side so that the number on `@HEAD` gets bumped.

deepthi:
https://vitess.slack.com/archives/CNE9WP677/p1631785758026500?thread_ts=1631654474.025500&cid=CNE9WP677

Will upgrading controller-gen.kubebuilder.io to v0.3 invalidate any current vitess resources?
Is v1.19 the best target? It’ll be end-of-life in 4 weeks

deepthi:
I'm going to merge this now, we can address any other concerns separately.

dctrwatson:
> Will upgrading controller-gen.kubebuilder.io to v0.3 invalidate any current vitess resources?

afaik, it should not invalidate any resources

> Is v1.19 the best target? It’ll be end-of-life in 4 weeks

It is not the best target, but highest target with the least amount of effort/time.

We need to look into the work it would take to continue to upgrade operator-sdk (there are breaking changes) or just divesting the operator from operator-sdk.

PR #224, opened 2021-09-23
pkg/operator/vttablet/datastore.go: Remove logic around recently depr…

gedgar:
…ecated demote_master_type flag

https://github.com/vitessio/vitess/pull/8511 removes the `demote_master_type` flag, which is breaking pools of `type: externalmaster`:
```
flag provided but not defined: -demote_master_type
Usage of /vt/bin/vttablet:
  -allowed_tablet_types value
    	Specifies the tablet types this vtgate is allowed to route queries to
```

Signed-off-by: Gary Edgar <gary@planetscale.com>

### DIFF
--- a/pkg/operator/vttablet/datastore.go
+++ b/pkg/operator/vttablet/datastore.go
@@ -18,7 +18,6 @@ package vttablet
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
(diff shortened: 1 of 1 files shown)


deepthi (commented):
You need to fix the build error, the rest LGTM.
`init_tablet_type` is being set correctly here:
https://github.com/planetscale/vitess-operator/blob/main/pkg/operator/vttablet/flags.go#L68

deepthi (approved):

Issue #237, opened 2021-12-01
abount #83  Restart tablet Pods to complete pending PVC filesystem resizes


huangzhiyong:
hi,
We use EKS+ EBS CSI. When the cluster has more shards (there are more EBS attached), when we expand EBS, due to CSI efficiency issues, many PVs are in the corev1.PersistentVolumeClaimFileSystemResizePending state, but this is expected.
We don't want to restart POD. How to avoid this situation?

PR #238, opened 2021-12-09
add flag --enable_online_fs_expansion: control whether pods that refe…

huangzhiyong:
…rence the resized volume need to be restarted.

Kubernetes v1.11 also introduces an alpha feature called online file system expansion. This feature enables file system expansion while a volume is still in-use by a pod. Because this feature is alpha, it requires enabling the feature gate, ExpandInUsePersistentVolumes. It is supported by the in-tree volume plugins GCE-PD, AWS-EBS, Cinder, and Ceph RBD. When this feature is enabled, pod referencing the resized volume do not need to be restarted. Instead, the file system will automatically be resized while in use as part of volume expansion. File system expansion does not happen until a pod references the resized volume, so if no pods referencing the volume are running file system expansion will not happen.


huangzhiyong:
#237

deepthi:
The DCO needs to be fixed. If the reviewers like the change, we can run CI.

huangzhiyong:
DCO -> fixed

pkg/controller/vitessshard/vitessshard_controller.go
```diff
@@ -50,6 +50,7 @@ const (
 var (
 	maxConcurrentReconciles = flag.Int("vitessshard_concurrent_reconciles", 10, "the maximum number of different vitessshards to reconcile concurrently")
 	resyncPeriod            = flag.Duration("vitessshard_resync_period", 30*time.Second, "reconcile vitessshards with this period even if no Kubernetes events occur")
+	onlineFileSystemExpansion = flag.Bool("enable_online_fs_expansion", true, "if true, pod referencing the resized volume do not need to be restarted, but provided that the volume plug-in supports, such as GCE-PD, AWS-EBS, Cinder, and Ceph RBD")
```
mattlord, line 53: IMO we should make this behavior opt-in because we have no way to know that this will work in a given environment, right? That would mean  making the default `false`. I would recommend:
```
onlineFileSystemExpansion = flag.Bool("enable_online_fs_expansion", false, "if true, pods referencing the resized volume do not need to be restarted; you must ensure that the CSI driver(s) in use support online resizing, e.g. GCE-PD, AWS-EBS, Cinder, or Ceph RBD")
```

mattlord (changes requested):
Thank you for the contribution! Aside from my 1 comment/request, it makes sense. I'll let @dctrwatson comment though as he's much more familiar with the operator than I am at this point.

pkg/controller/vitessshard/vitessshard_controller.go
dctrwatson: We can at least wait 2 releases before changing the default

pkg/controller/vitessshard/reconcile_disk.go
```diff
@@ -105,7 +105,9 @@ func (r *ReconcileVitessShard) reconcileDisk(ctx context.Context, vts *planetsca
 
 	// If disk size has changed and the changes are all ready, mark the shard as ready to cascade. Otherwise, skip this.
 	if anythingChanged {
-		rollout.Cascade(vts)
+		if !*onlineFileSystemExpansion {
```
dctrwatson, line 108: We can move this check to the beginning of `reconcileDisk` where it checks `vts.Spec.UpdateStrategy.Type` and just `return resultBuilder.Result()` because I don't think this function provides any value when using `onlineFileSystemExpansion`

mattlord:
Closing this as stale

PR #243, opened 2022-01-31
Add forcePathStyle setting for S3 backup

L3o-pold:
Could be a better idea to add support for vttablet `extraFlags` .

### DIFF
--- a/deploy/crds/planetscale.com_vitessbackupstorages.yaml
+++ b/deploy/crds/planetscale.com_vitessbackupstorages.yaml
@@ -125,6 +125,8 @@ spec:
                         type: string
                       endpoint:
                         type: string
+                      forcePathStyle:
+                        type: boolean
                       keyPrefix:
                         maxLength: 256
                         pattern: ^[^\r\n]*$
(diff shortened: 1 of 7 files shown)


deepthi:
Is the issue that vttablet supports `extraFlags` but they are not used by the `vitessbackup` object?

L3o-pold:
Main issue is that we can't set the `forcePathStyle` settings. A way to fix this is to do as I did in the PR.

An other way to fix the issue (and in a more durable and flexible way) could be to as you said use `extraFlags` in a `vitessbackup`.

dctrwatson (approved):
This seems like a straightforward/simple solution to the issue.

I agree adding `ExtraFlags` would be more flexible; just it would need to be added to each backup location type + methods.

L3o-pold:
I rebase my changes if we can move forward with this.

deepthi:
> I rebase my changes if we can move forward with this.

Yes, please

L3o-pold:
> > I rebase my changes if we can move forward with this.
> 
> Yes, please

done

Issue #240, opened 2022-01-25
initContainer defaults preventing existing YAML from working


aquarapid:
I have a (test) operator cluster YAML that has a keyspace section looking like this:

```
  keyspaces:
  - name: main
    partitionings:
    - equal:
        parts: 2
        shardTemplate:
          databaseInitScriptSecret:
            name: example-cluster-config
            key: init_db.sql
          replication:
            enforceSemiSync: true
          tabletPools:
          - cell: uscentral1a
            type: replica
            replicas: 1
            vttablet:
              extraFlags:
                db_charset: utf8mb4
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            mysqld:
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            dataVolumeClaimTemplate:
              accessModes: ["ReadWriteOnce"]
              resources:
                requests:
                  storage: 5Gi
              storageClassName: uscentral1a-ssd
          - cell: uscentral1c
            type: replica
            replicas: 1
            vttablet:
              extraFlags:
                db_charset: utf8mb4
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            mysqld:
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            dataVolumeClaimTemplate:
              accessModes: ["ReadWriteOnce"]
              resources:
                requests:
                  storage: 5Gi
              storageClassName: uscentral1c-ssd
          - cell: uscentral1f
            type: replica
            replicas: 1
            vttablet:
              extraFlags:
                db_charset: utf8mb4
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            mysqld:
              resources:
                requests:
                  cpu: 10m
                  memory: 256Mi
            dataVolumeClaimTemplate:
              accessModes: ["ReadWriteOnce"]
              resources:
                requests:
                  storage: 5Gi
              storageClassName: uscentral1f-ssd
```

This used to work just fine.  But since PR #216 I get the multiple errors (one for each vttablet) when applying this (can be seen with `kubectl get events` or in the operator logs):

```
[...]

PR #244, opened 2022-01-31
Revert "Set vttablet init container resource limits"

nickvanw:
Reverts planetscale/vitess-operator#216

This is causing issues (https://github.com/planetscale/vitess-operator/issues/240), let's revert it for now and we can figure out what the solution is. 

cc @deepthi & @aquarapid

### DIFF
--- a/pkg/operator/vttablet/mysqlctld.go
+++ b/pkg/operator/vttablet/mysqlctld.go
@@ -18,7 +18,6 @@ package vttablet
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/pointer"
 
 	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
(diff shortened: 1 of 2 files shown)


deepthi (approved):

deepthi:
@brandt we have to revert #216 until #240 can be fixed.
Would you be able to do another PR that can get you what you need without breaking the case in #240?

brandt:
Ah, I see in `update.ResourceRequirements(dst, src)` an undefined limit in `src` will not overwrite a defined limit in `dst`. 

This should be fixable by simply removing the default limit on the init container. If the cluster requires a limit to be set, the user will have to have already defined it on the vttablet container, and the init container limit will be automatically updated to use the same value. If the vttablet limit is _not_ set, the init container limit will remain unset as well.

Does that sound like a reasonable approach?

deepthi:
@dctrwatson can you comment on the proposed fix?

dctrwatson:
> Does that sound like a reasonable approach?

> @dctrwatson can you comment on the proposed fix?

Ah ya, that sounds like it would be a workable fix.

PR #252, opened 2022-02-21
Fixed `PriorityClass` API Version and Docker Image tag for vtop

frouioui:
## Description

Following the release of Kubernetes v1.22 many APIs were no longer served including the one used by our `PriorityClass` (`scheduling.k8s.io/v1beta1`). This pull request makes sure we use the `v1` of `scheduling.k8s.io`. More information on deprecated APIs [here](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes).

Additionally, the docker tag used for `planetscale/vitess-operator` changed from `v2.5.0` to `latest`. This change was motivated by the fact that we use `latest` for all our Docker images in the vitess repository. I propose that we use a pinned tag (i.e `v2.5.0`) when we are creating a new release commit of vtop, and then switch it back to `latest` on main.


deepthi (approved):

deepthi:
> I propose that we use a pinned tag (i.e v2.5.0) when we are creating a new release commit of vtop, and then switch it back to latest on main.

This sounds good.

PR #260, opened 2022-04-06
End to end upgrade testing

GuptaManan100:
This PR adds end to end upgrade testing for Vitess-operator. In the workflow we start with a `v2.5.1` version operator image and Vitess components of version `12.0.3`. We then upgrade the operator version to the latest version (Docker image built from the PR) and then we update the Vitess components to version `13.0.0`. After this upgrade, we go on to run the remainder of the example script at https://vitess.io/docs/14.0/get-started/operator/, performing a Move tables workflow and a reshard operation. This should give us enough confidence that we upgraded a running Vitess cluster (operator and all components), and performed other essential operations on the upgraded cluster.

Before we cut for any release of the Vitess-operator, we should keep the versions of Vitess and operator being used in these tests. For example, in the PR that updates the dependency of the Vitess-operator to Vitess version 13.0.0, we should update the tests to go from `13.0.0` to `latest` image of Vitess and use `v2.6.0` Vitess-operator image initially.


.github/workflows/upgrade-test.yaml
```diff
@@ -0,0 +1,48 @@
...
+  build:
+    name: Upgrade Test
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ~1.17
+      id: go
```
frouioui, line 22: nit: this can be removed

test/upgrade/operator/201_customer_tablets.yaml
```diff
@@ -0,0 +1,116 @@
...
+metadata:
+  name: example
+spec:
+  images:
+    vtctld: vitess/lite:latest
+    vtgate: vitess/lite:latest
+    vttablet: vitess/lite:latest
+    vtbackup: vitess/lite:latest
+    mysqld:
+      mysql56Compatible: vitess/lite:latest
```
frouioui, line 12: Shouldn't these be set to `v13.0.0` like `cluster_upgrade.yaml`?

test/upgrade/operator/cluster_upgrade.yaml
```diff
@@ -0,0 +1,191 @@
...
+type: Opaque
+stringData:
+  users.json: |
+    {
+      "user": [{
+        "UserData": "user",
+        "Password": ""
+      }]
+    }
+  init_db.sql: |
```
frouioui, line 99: This whole block can be removed

test/upgrade/test.sh
```diff
@@ -0,0 +1,341 @@
...
+  waitForKeyspaceToBeServing customer -80 1
+  waitForKeyspaceToBeServing customer 80- 1
+}
+
+function upgradeToLatest() {
+  echo "Apply operator-latest.yaml "
+  kubectl apply -f operator-latest.yaml
+
+  echo "Upgrade all the other binaries"
+  kubectl apply -f cluster_upgrade.yaml
```
frouioui, line 288: nit: if you don't want to maintain two pairs of identic files: `operator-latest.yaml`-`operator.yaml`, and `101_initial_cluster.yaml`-`cluster_upgrade.yaml`, you can `sed` the files to change the tag of the Docker images

frouioui:
Regarding the repeated test timeout when creating the initial cluster, hopefully, it won't timeout anymore on the self-hosted runners.

.github/workflows/upgrade-test.yaml
frouioui: it's not being highlighted, but i was mentioning line 22

GuptaManan100: Yeah, I got that

test/upgrade/operator/201_customer_tablets.yaml
GuptaManan100: Correct. good catch

test/upgrade/test.sh
GuptaManan100: Yes, I'll do that

GuptaManan100:
I don't think it will timeout on a self-hosted runner. Currently on my machine, it works in under 25 or so minutes. In any case, if it times out, I'll see how to increase the time limit.

test/upgrade/test.sh
GuptaManan100: On second thought, or rather the reason I did it this way in the first place, was because we will be introducing a new flag very soon, and then upgrading to a new version of the operator will also require specifying that flag, which will also need to be added to the `operator-latest.yaml`. So, I didn't generate these files, since we would be throwing that code away in a separate PR real soon

frouioui: Got it, makes sense! What will that flag do?

.github/workflows/upgrade-test.yaml
GuptaManan100: fixed

test/upgrade/operator/201_customer_tablets.yaml
GuptaManan100: fixed

test/upgrade/test.sh
GuptaManan100: Our beloved, `durability_policy` is going to make its appearance here

frouioui (approved):
Looks great to me! Very excited about this 🚀

I think we should hold this Pull Request until we try the new workflow against the new self-hosted runner

GuptaManan100:
I agree, I don't want to merge a CI workflow which keeps failing after timing out for 30 minutes

test/upgrade/test.sh
frouioui: Gotcha!

frouioui:
@deepthi we should make `Upgrade Test` a required CI check

GuptaManan100:
UPDATE: I will move this to Buildkite, since we have all the infra set there instead of self-hosted runner

frouioui (approved):
BuildKit change looks good to me

GuptaManan100:
Buildkite test hasn't run yet, we should merge only after it passes successfully

deepthi (approved):

PR #263, opened 2022-05-26
Upgrade Vitess dependency to 13.0.0

GuptaManan100:
## Description

This PR upgrades the operator dependency to release 13.0.0. As part of this upgrade, some build issues had to be fixed due to refactors in Vitess commands like (https://github.com/vitessio/vitess/pull/9015). 

We have also introduced a new semi-sync parameter in some RPCs, but it is not being used. In order to fix the build issues, we are passing a false value for this parameter. As and when we start using this parameter, we should also revisit this code. The upgrade path that has been decided is to not use this parameter in release-14.0 (making this change safe). We will store the durability_policy in the topo instead of a flag to binaries like vtctld. The RPCs for this have been introduced in https://github.com/vitessio/vitess/pull/10221.

We should use those RPCs in these places, to find the correct durability_policy and use that to set semi-sync. This will be needed to be done when we upgrade the Vitess dependency to release 14.0.0

We have also needed to set a durability_policy in the main() function. Currently we are setting the `none` duraibility`. When we upgrade the Vitess dependency this will also have to change.


pkg/controller/vitessshardreplication/init_restored_shard.go
```diff
@@ -231,7 +231,9 @@ func electInitialShardPrimary(ctx context.Context, keyspaceName, shardName strin
 		return nil, fmt.Errorf("lost topology lock; aborting: %v", err)
 	}
 	// Promote the candidate to primary.
-	_, err = wr.TabletManagerClient().PromoteReplica(ctx, candidatePrimary.tablet.Tablet)
+	// TODO (GuptaManan100): We are passing false for semiSync parameter since it is not being used currently.
+	// When we change this behaviour in Vitess, we should also revisit here.
```
deepthi, line 235: Can you document why we might change this behavior?

test/endtoend/operator/101_initial_cluster.yaml
```diff
@@ -8,12 +8,12 @@ metadata:
   name: example
 spec:
   images:
-    vtctld: vitess/lite:v12.0.3
-    vtgate: vitess/lite:v12.0.3
-    vttablet: vitess/lite:v12.0.3
-    vtbackup: vitess/lite:v12.0.3
+    vtctld: vitess/lite:v13.0.0
```
deepthi, line 11: should this be 13.0.0 or a later minor version?

test/endtoend/operator/101_initial_cluster_backup.yaml
```diff
@@ -15,12 +15,12 @@ spec:
           path: /backup
           type: Directory
   images:
-    vtctld: vitess/lite:v13.0.0
-    vtgate: vitess/lite:v13.0.0
-    vttablet: vitess/lite:v13.0.0
-    vtbackup: vitess/lite:v13.0.0
+    vtctld: vitess/lite:latest
```
deepthi, line 18: why is this latest and not v13.0.x?

deepthi (commented):
Nice work!
I like that we have started using VtctldServer.
A few questions, the rest lgtm.
We should get a review from @Verolop before merging.

test/endtoend/operator/101_initial_cluster.yaml
GuptaManan100: We can do any/all. I just added this to check that we are able to upgrade from the earliest release

test/endtoend/operator/101_initial_cluster_backup.yaml
GuptaManan100: This file is used for the backup test and not the upgrade test. In the backup test we are testing that the backups work with the latest version of Vitess. Earlier I had 13.0.0 there, because the code on the operator was supposed to only work with 13.0.0. After this PR, the code will be meant to be working with the release-14 of Vitess, so we use the latest image.

pkg/controller/vitessshardreplication/init_restored_shard.go
GuptaManan100: Do you want me to add it in the code as more comments?

Verolop (approved):
LGTM!

pkg/controller/vitessshardreplication/init_restored_shard.go
deepthi: link to a vitessio/vitess issue if we have one?

PR #269, opened 2022-06-16
Update compatibility table for v2.6*

Verolop:
add vitess v12.0.* compatibility for v2.6.*

PR #270, opened 2022-06-17
Fix end to end tests

GuptaManan100:
## Description

This PR fixes the end to end tests by changing the Vitess image they use. The tests are currently supposed to test that upgrading from 13.0.0 to the code in release-branch 14.0 is safe and the current version of operator supports it.
To do this, we were earlier using the image tag `latest` because the 14.0 code lived on main. After we cut release-14.0, we should start using the rc-1 images.

### DIFF
--- a/test/endtoend/operator/101_initial_cluster_backup.yaml
+++ b/test/endtoend/operator/101_initial_cluster_backup.yaml
@@ -15,12 +15,12 @@ spec:
           path: /backup
           type: Directory
   images:
-    vtctld: vitess/lite:latest
-    vtgate: vitess/lite:latest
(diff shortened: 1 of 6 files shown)


frouioui:
@GuptaManan100, do you have a link to the failing end-to-end test?

GuptaManan100:
@frouioui Sure, take a look at the CI of https://github.com/planetscale/vitess-operator/pull/269

frouioui (approved):

deepthi:
Can one of you create an issue to bump this again after vitess 14 GA? We should add it to the release checklist.

PR #272, opened 2022-06-21
Deploying VTAdmin in Vtop

GuptaManan100:
## Description

This PR adds VTAdmin to the operator and associated tests.
Currently only the specific vtadmin image can be used for deploying VTAdmin, but we'll get `vitess/lite` image working too later.

An example config looks like the following - 
```yaml
spec:
  images:
    vtadmin: vitess/vtadmin:latest
  vtadmin:
    rbac:
      name: example-cluster-config
      key: rbac.yaml
    cells:
      - zone1
    apiAddresses:
      - http://localhost:14001
    replicas: 1
    readOnly: false
    apiResources:
      limits:
        memory: 128Mi
      requests:
        cpu: 100m
        memory: 128Mi
    webResources:
      limits:
        memory: 128Mi
      requests:
        cpu: 100m
        memory: 128Mi
```

The VTAdmin spec includes the following fields - 
- `Rbac` as a secret source. It is the role based access control rules to use for VTAdmin API
- `Cells` as a list of strings. It is the cells where VTAdmin must be deployed. Defaults to deploying everywhere
- `APIAddresses` as a list of strings. Since the VTAdmin web UI runs on the client side, it needs the API address to use to query the API. We can't use the internal kubernetes service address or the ip of the pod, since they aren't visible outside the cluster. The API Address must be provided by the user based on how they export the service of VTAdmin API. In our tests we port-forward the service to port `14001` and therefore that is the address provided here. This value is a list because the address to be used for VTAdmin web in each cell might be different. If only 1 value is provided then, that is used for all the cells. The ideal way to deploy this would be to export each individual VTAdmin service (that we create) in each cell and attach external IPs to them and provide those IPs here.
- `Replicas` - Number of VTAdmin deployments required per cell. We setup a service on top of the web and API connections, so load-balancing comes out of the box
- `WebResources` - Resource requests and limits for the container running the VTAdmin web server
- `APIResources` - Resource requests and limits for the container running the VTAdmin API server
- `ReadOnly` - Configuration to set the VTAdmin web UI to be read only.

## Documentation
https://github.com/vitessio/website/pull/1073


docs/api/index.html
```diff
@@ -3752,6 +3780,19 @@ <h3 id="planetscale.com/v2.VitessClusterStatus">VitessClusterStatus
...
 <td>
+<code>vtadmin</code></br>
+<em>
+<a href="#planetscale.com/v2.VtadminStatus">
+VtadminStatus
+</a>
+</em>
+</td>
+<td>
+<p>Vtadmin is a summary of the status of the vtctld deployment.</p>
```
deepthi, line 3791:
```suggestion
<p>Vtadmin is a summary of the status of the vtadmin deployment.</p>
```

docs/api/index.html
```diff
@@ -4335,6 +4376,19 @@ <h3 id="planetscale.com/v2.VitessImagePullPolicies">VitessImagePullPolicies
...
 <td>
+<code>vtadmin</code></br>
+<em>
+<a href="https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#pullpolicy-v1-core">
+Kubernetes core/v1.PullPolicy
+</a>
+</em>
+</td>
+<td>
+<p>Vtadmin is the container image pull policy to use for Vtadmin instances.</p>
```
deepthi, line 4387:
```suggestion
<p>vtadmin is the container image pull policy to use for Vtadmin instances.</p>
```

docs/api/index.html
```diff
@@ -370,6 +370,19 @@ <h3 id="planetscale.com/v2.VitessCluster">VitessCluster
...
 <td>
+<code>vtAdmin</code></br>
+<em>
+<a href="#planetscale.com/v2.VtAdminSpec">
+VtAdminSpec
+</a>
+</em>
+</td>
+<td>
+<p>Vtadmin deploys a set of VtAdmin servers for the Vitess cluster.</p>
```
deepthi, line 381:
```suggestion
<p>VtAdmin deploys a set of Vitess Admin servers for the Vitess cluster.</p>
```

docs/api/index.html
```diff
@@ -4443,6 +4497,17 @@ <h3 id="planetscale.com/v2.VitessImages">VitessImages
 </tr>
 <tr>
 <td>
+<code>vtadmin</code></br>
+<em>
+string
+</em>
+</td>
+<td>
+<p>Vtadmin is the container image (including version tag) to use for Vtadmin instances.</p>
```
deepthi, line 4506:
```suggestion
<p>Vtadmin is the container image (including version tag) to use for Vitess Admin instances.</p>
```

deepthi: Not quite sure what this should _really_ be called, so this is just a suggestion.

deepthi: Is this optional? My recollection is that for other images (like vttablet), we default to `vitess/lite:latest` if the image is not specified. But since vtadmin is still experimental, if this image is not provided we should skip the deployment.

docs/api/index.html
```diff
@@ -3577,6 +3592,19 @@ <h3 id="planetscale.com/v2.VitessClusterSpec">VitessClusterSpec
...
 <td>
+<code>vtAdmin</code></br>
+<em>
+<a href="#planetscale.com/v2.VtAdminSpec">
+VtAdminSpec
+</a>
+</em>
+</td>
+<td>
+<p>Vtadmin deploys a set of VtAdmin servers for the Vitess cluster.</p>
```
deepthi, line 3603:
```suggestion
<p>Vtadmin deploys a set of Vitess Admin servers for the Vitess cluster.</p>
```

deepthi: Alternatively: `Vitess Admin servers` -> `Admin servers`

docs/api/index.html
GuptaManan100: Fixed

docs/api/index.html
GuptaManan100: Fixed

docs/api/index.html
GuptaManan100: Fixed. Also added a check in the code for the same. If the image is not specified, we log an error and don't deploy vtadmin

docs/api/index.html
GuptaManan100: Fixed

.buildkite/pipeline.yml
```diff
@@ -39,3 +39,21 @@ steps:
           propagate-environment: true
           volumes:
             - "/var/run/docker.sock:/var/run/docker.sock"
+
+  - name: "VTAdmin Test"
+    command:
+    - apk add g++ make bash curl mysql mysql-client chromium
```
GuptaManan100, line 45: This line is different from the other test setup lines. It is also installing chromium browser. We need this to run it in a headless mode to download the VTAdmin web pages. We can't use `curl` (which we use for the API pages) or `wget` because they do not support JavaScript, so a headless browser was the easiest solution.

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -0,0 +1,373 @@
...
+		return resultBuilder.Result()
+	}
+
+	if len(vt.Spec.VtAdmin.APIAddresses) == 0 {
+		log.Errorf("Not deploying vtadmin since api addresses field is not specified. Atleast 1 value is required")
+	}
+
+	if len(vt.Spec.VtAdmin.APIAddresses) != 1 && len(vt.Spec.VtAdmin.APIAddresses) != len(vt.Spec.VtAdmin.Cells) {
+		log.Errorf("Not deploying vtadmin since api addresses field doesn't align with cells field")
+	}
```
GuptaManan100, line 58: These are user input validations. There exist 3 - 
- The VTAdmin image should be specifically provided. We do not add a default for `vitess:lite` since that won't work
- There should be at least 1 APIAddress provided
- If more than 1 APIAddresses are provided, then the length of this list should match the length of the cells list.

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -0,0 +1,373 @@
...
+		discoverySecret, err := r.createDiscoverySecret(ctx, vt, cell)
+		if err != nil {
+			return nil, err
+		}
+
+		// We have already checked that atleast 1 value should be available in APIAddresses
+		apiAddress := vt.Spec.VtAdmin.APIAddresses[0]
+		if len(vt.Spec.VtAdmin.APIAddresses) > 1 {
+			apiAddress = vt.Spec.VtAdmin.APIAddresses[idx]
+		}
```
GuptaManan100, line 192: This is safe because of the user input validation above

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -0,0 +1,373 @@
...
+    ],
+    "vtgates": [
+        {
+            "host": {
+                "hostname": "%s:%d"
+            }
+        }
+    ]
+}`, vtctldServiceIP, vtctldServiceWebPort, vtctldServiceIP, vtctldServiceGrpcPort, vtgateServiceIP, vtgateServiceGrpcPort)
+	secretName := vtadmin.DiscoverySecretName(vt.Name, cell.Name)
```
GuptaManan100, line 302: The discovery configuration file is autogenerated using the global service of `vtctld` and per cell service of `vtgates`.

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -0,0 +1,373 @@
...
+	configKey := vtadmin.WebConfigFileName
+	configVal := fmt.Sprintf(`window.env = {
+     'REACT_APP_VTADMIN_API_ADDRESS': "%s",
+     'REACT_APP_FETCH_CREDENTIALS': "omit",
+     'REACT_APP_ENABLE_EXPERIMENTAL_TABLET_DEBUG_VARS': false,
+     'REACT_APP_BUGSNAG_API_KEY': "",
+     'REACT_APP_DOCUMENT_TITLE': "",
+     'REACT_APP_READONLY_MODE': %s,
+ };`, apiAddress, convertReadOnlyFieldToString(vt.Spec.VtAdmin.ReadOnly))
+	secretName := vtadmin.WebConfigSecretName(vt.Name, cell.Name)
```
GuptaManan100, line 328: Some parameters of the config are not user controllable like `REACT_APP_ENABLE_EXPERIMENTAL_TABLET_DEBUG_VARS` is always set to false. Since we do not have a good way of discovering vttablets yet, this doesn't work. `http-tablet-url-tmpl` is not flexible to allow this right now. To enable this, we would have to setup services for all the vttablets and then encode them in the flag. Other complexities like new vttablets requiring change in config, etc will also have to be dealt with.
Similarly, `REACT_APP_FETCH_CREDENTIALS` is always set to `omit` because we aren't enabling CORS. If we did want to do that, then we would also need to ask the user to provide the hosts that are allowed to access the API server

pkg/operator/secrets/volume_mount.go
```diff
@@ -41,6 +41,9 @@ type VolumeMount struct {
 	// DirName is the name of the directory under VolumeMountDir in which to
 	// mount this SecretSource.
 	DirName string
+	// AbsolutePath stores the absolute path to use instead of the generated path
+	// with /vt/secret as the prefix
+	AbsolutePath string
```
GuptaManan100, line 46: This had to be added to allow mounting files to a specific path instead of the `/vt/secrets` where they usually are. For the Vtadmin web config, the path has to be `/vt/web/vtadmin/build/config`

pkg/operator/vtadmin/deployment.go
```diff
@@ -0,0 +1,362 @@
...
+			},
+			InitialDelaySeconds: 300,
+			FailureThreshold:    30,
+		},
+		VolumeMounts: spec.ExtraVolumeMounts,
+		Env:          spec.ExtraEnv,
+	}
+	update.ResourceRequirements(&apiContainerResources, &spec.APIResources)
+	updateRbac(spec, apiFlags, vtadminAPIContainer, &obj.Spec.Template.Spec)
+	updateDiscovery(spec, apiFlags, vtadminAPIContainer, &obj.Spec.Template.Spec)
```
GuptaManan100, line 214: Adds the file mounts for Rbac and discovery config

pkg/operator/vtadmin/deployment.go
```diff
@@ -0,0 +1,362 @@
...
+					Port: intstr.FromString(planetscalev2.DefaultWebPortName),
+				},
+			},
+			InitialDelaySeconds: 300,
+			FailureThreshold:    30,
+		},
+		VolumeMounts: spec.ExtraVolumeMounts,
+		Env:          spec.ExtraEnv,
+	}
+	updateWebConfig(spec, vtadminWebContainer, &obj.Spec.Template.Spec)
```
GuptaManan100, line 262: Adds the file mount for the web config

GuptaManan100:
Also, the vtadmin test needs to restarted after the image `vitess/vtadmin:latest` is pushed to docker hub. The associated PR for this change is https://github.com/vitessio/vitess/pull/10543

deploy/crds/planetscale.com_vitessclusters.yaml
```diff
@@ -1989,6 +1993,184 @@ spec:
                   tolerations:
                     x-kubernetes-preserve-unknown-fields: true
                 type: object
+              vtAdmin:
```
L3o-pold, line 1996: should be lowercase I think (same in the PR description)

GuptaManan100: Yes, this makes sense. Fixed it now

pkg/controller/vitesscluster/reconcile_vtadmin.go
frouioui: That makes sense to me

test/endtoend/operator/101_initial_cluster_vtadmin.yaml
```diff
@@ -0,0 +1,231 @@
...
+# Please refer to the operator documentation for recommendations on
+# production settings.
+apiVersion: planetscale.com/v2
+kind: VitessCluster
+metadata:
+  name: example
+spec:
+  images:
+    vtctld: vitess/lite:v14.0.0-rc1
+    vtadmin: vitess/vtadmin:latest
```
frouioui, line 12: Yeah, I think `latest` is fine here. We will change it tomorrow during the release

pkg/operator/vtadmin/deployment.go
```diff
@@ -0,0 +1,362 @@
...
+				Name:          planetscalev2.DefaultWebPortName,
+				Protocol:      corev1.ProtocolTCP,
+				ContainerPort: planetscalev2.DefaultWebPort,
+			},
+		},
+		Command: []string{webDir + "/node_modules/.bin/serve"},
+		Args: []string{
+			// Symlinks are required because the web config file is mounted as a
+			// secret which happens to be mounted as symlink instead of an actual file
+			"--symlinks",
```
frouioui, line 233: 👍🏻

docs/api/index.html
```diff
@@ -3577,6 +3592,19 @@ <h3 id="planetscale.com/v2.VitessClusterSpec">VitessClusterSpec
...
 <td>
+<code>vtadmin</code></br>
+<em>
+<a href="#planetscale.com/v2.VtAdminSpec">
+VtAdminSpec
+</a>
+</em>
+</td>
+<td>
+<p>VtAdmin deploys a set of Vitess Admin servers for the Vitess cluster.</p>
```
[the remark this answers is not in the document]
GuptaManan100: This is all generated code, we'll have to change the comments to change these and then they'll complain about formatting

pkg/controller/vitesscluster/reconcile_vtadmin.go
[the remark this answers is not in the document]
GuptaManan100: Yes I tried to, but it got cluttered and I wanted different error messages from both anyways.

test/endtoend/operator/101_initial_cluster_vtadmin.yaml
GuptaManan100: Yes, exactly

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -0,0 +1,373 @@
...
+
+	if len(vt.Spec.VtAdmin.APIAddresses) == 0 {
+		log.Errorf("Not deploying vtadmin since api addresses field is not specified. Atleast 1 value is required")
+	}
+
+	if len(vt.Spec.VtAdmin.APIAddresses) != 1 && len(vt.Spec.VtAdmin.APIAddresses) != len(vt.Spec.VtAdmin.Cells) {
+		log.Errorf("Not deploying vtadmin since api addresses field doesn't align with cells field")
+	}
+
+	key := client.ObjectKey{Namespace: vt.Namespace, Name: vtadmin.ServiceName(vt.Name)}
```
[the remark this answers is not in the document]
GuptaManan100: That is a function which deterministically gives the service name to use. We are creating the service just down below. We always create services for vtctld and vtgates in cells, so that if there are multiple deployments then users can just assign a public IP address to the service and it will load balance across all the available pods

frouioui (approved):

PR #279, opened 2022-07-06
VTAdmin cluster-config cleanup

GuptaManan100:
## Description

VTop currently passes the cluster configuration as a comma-separated list of values, but VTAdmin has a flag to allow it to pass the same information in a YAML file. Since the YAML file is the recommended way of providing this information, we should be using that instead. It also adds additional security, in the sense that the cluster configuration is now a secret instead of being readable from the flags provided to the binary.
This PR accomplishes this change.

### DIFF
--- a/pkg/controller/vitesscluster/reconcile_vtadmin.go
+++ b/pkg/controller/vitesscluster/reconcile_vtadmin.go
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"path/filepath"
 	planetscalev2 "planetscale.dev/vitess-operator/pkg/apis/planetscale/v2"
 	"planetscale.dev/vitess-operator/pkg/operator/conditions"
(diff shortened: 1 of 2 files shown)


GuptaManan100:
This PR is based on top of changes in https://github.com/planetscale/vitess-operator/pull/278 and should be merged after it.

pkg/controller/vitesscluster/reconcile_vtadmin.go
```diff
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"path/filepath"
```
frouioui, line 26: nit: we should add a go-import check to that repository

frouioui (approved):

pkg/controller/vitesscluster/reconcile_vtadmin.go
GuptaManan100: Yep, we should. I'll fix this for now

GuptaManan100: Fixed

PR #282, opened 2022-07-10
Introduce Durability Policy in VTop

GuptaManan100:
## Description

This PR introduces the durability policy in the Vitess Operator. Durability Policy is introduced as a configuration in the Keyspace specification. 
Furthermore, the reconciliation logic for the same has also been added which ensures that the durability policy configured matches the durability policy in the topo server.

If the durability_policy is left unspecified, then VTop will not set a default or configure it at all. This is required for upgrade considerations. Since the previous version of VTop didn't support this configuration, when VTop is upgraded, it is expected that the durability policy configuration will be empty. If it is, we do not want to presume its value and set a default because that could be different than the durability policy that the user has already set in v14 using SetKeyspaceDurabilityPolicy.
Once VTop is upgraded, the durability policy config can be added and VTop will ensure that it is configured properly henceforth.

### DIFF
--- a/test/endtoend/operator/operator-latest.yaml
+++ b/test/endtoend/operator/operator-latest.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
--- a/test/endtoend/operator/operator-latest.yaml
+++ b/test/endtoend/operator/operator-latest.yaml
@@ -1,3 +1,5 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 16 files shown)


Verolop (approved):
✨

test/endtoend/operator/operator-latest.yaml
```diff
@@ -1,3 +1,5 @@
+
```
frouioui, line 1: nit: is that extra line needed?

frouioui (approved):

test/endtoend/operator/operator-latest.yaml
GuptaManan100: No not really, must have come while copy-pasting the crds

GuptaManan100: Fixed

PR #283, opened 2022-07-12
Augment VTAdmin test to also test VTOrc setup

GuptaManan100:
## Description

This PR augments the VTAdmin end-to-end test to also test VTOrc. The initial cluster-setup file is changed to also launch VTOrc. After verifying that VTAdmin is working as expected, we also go on to verify that VTOrc runs as expected. This is done by stopping replication on all the valid replicas. Following this, we try and write to the primary, which only succeeds if VTOrc can fix the failure and repair replication. 
An additional flag `disable-active-reparent` has been added to the vttablet in the test setup, just as how we recommend users to run. Furthermore, VTop only repairs replication if the source set on a tablet doesn't match the primary information. All of these pieces together mean that our test for the VTOrc setup is correct.
I have also manually verified that if VTOrc is not running, the writes are stalled indefinitely. 

There is a scope for improving the test to timeout after some time. Currently, that part is left to the CI timeout configured to 1 hour in the BuildKite pipeline configuration.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -47,9 +47,9 @@ steps:
     - tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz
     - export PATH=$PATH:/usr/local/go/bin
     - rm go1.18.3.linux-amd64.tar.gz
-    - make vtadmin-test
+    - make vtorc-vtadmin-test
     concurrency: 1
-    concurrency_group: 'vtop/vtadmin-test'
+    concurrency_group: 'vtop/vtorc-vtadmin-test'
     timeout_in_minutes: 60
     plugins:
       - docker#v3.12.0:
--- a/test/endtoend/vtorc_vtadmin_test.sh
+++ b/test/endtoend/vtorc_vtadmin_test.sh
@@ -3,19 +3,20 @@
 source ./tools/test.env
 source ./test/endtoend/utils.sh
 
-# get_started_vtadmin:
-function get_started_vtadmin() {
+# get_started_vtorc_vtadmin:
+function get_started_vtorc_vtadmin() {
     echo "Apply latest operator-latest.yaml"
     kubectl apply -f "operator-latest.yaml"
     checkPodStatusWithTimeout "vitess-operator(.*)1/1(.*)Running(.*)"
 
-    echo "Apply 101_initial_cluster_vtadmin.yaml"
-    kubectl apply -f "101_initial_cluster_vtadmin.yaml"
+    echo "Apply 101_initial_cluster_vtorc_vtadmin.yaml"
+    kubectl apply -f "101_initial_cluster_vtorc_vtadmin.yaml"
     checkPodStatusWithTimeout "example-zone1-vtctld(.*)1/1(.*)Running(.*)"
     checkPodStatusWithTimeout "example-zone1-vtgate(.*)1/1(.*)Running(.*)"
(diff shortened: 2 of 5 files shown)


GuptaManan100:
I don't think 2 are required. I actually think 1 is preferable since we should have a test that launches all the Vitess components like the users would, instead of launching only some of them in each separate test.
If I add a separate test it would pretty much be the same test without VTAdmin running and without its test. So there would just be the overhead of setting up a new cluster and nothing different.

@frouioui Is there a reason why you would prefer 2 separate tests?

frouioui:
@GuptaManan100 okay sounds good to me! I was thinking about this in case we wanted to have "more" dedicated tests for vtorc and vtadmin. But setting one big cluster with all the components is good!

test/endtoend/vtorc_vtadmin_test.sh
```diff
@@ -111,6 +112,19 @@ function verifyVtadminSetup() {
...
+function verifyVTOrcSetup() {
+  # Stop replication using the vtctld and wait for VTOrc to repair
+  allReplicaTablets=$(getAllReplicaTablets)
+  for replica in $(echo "$allReplicaTablets") ; do
+    vtctlclient StopReplication "$replica"
+  done
+  # Now that we have stopped replication on both the tablets, we know that this will
+  # only succeed if VTOrc is able to fix it since we are running vttablet with disable active reparent
+  # and semi-sync durability policy
+  mysql -e "insert into customer(email) values('newemail@domain.com');"
```
frouioui, line 125: I am not sure how fast `StopReplication` is, but if it is not, maybe we should have some delay before calling the `mysql` command so we can ensure that the replication is stopped in the replicas.

.buildkite/pipeline.yml
```diff
@@ -40,16 +40,16 @@ steps:
...
     - apk add g++ make bash curl mysql mysql-client chromium
     - wget https://golang.org/dl/go1.18.3.linux-amd64.tar.gz
     - tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz
     - export PATH=$PATH:/usr/local/go/bin
     - rm go1.18.3.linux-amd64.tar.gz
-    - make vtadmin-test
+    - make vtorc-vtadmin-test
     concurrency: 1
-    concurrency_group: 'vtop/vtadmin-test'
+    concurrency_group: 'vtop/vtorc-vtadmin-test'
```
frouioui, line 52: Do we plan on adding more components to `vtorc-vtadmin-test`? If so, we could name it to something more generic such as `vitess-components-test`. Same for the Makefile rule, and the test script filename.

frouioui (approved):
Left two comments, otherwise, it looks good to me!

test/endtoend/vtorc_vtadmin_test.sh
GuptaManan100: The command won't return until replication is stopped. We don't need any wait after it.

frouioui: Okay that's perfect!

.buildkite/pipeline.yml
GuptaManan100: No, I don't there are any other Vitess components to be added. I think the vtorc-vtadmin name is more descriptive, but I don't have any strong preference between the two

frouioui: makes sense! i don't have strong preference too :)

PR #280, opened 2022-07-06
Upgrade Vitess dependency to 14.0.0

GuptaManan100:
## Description

This PR upgrades the operator dependency to release 14.0.0. As part of this upgrade, we also clean up all the TODOs inserted during the version upgrade to 13.0.0 https://github.com/planetscale/vitess-operator/pull/263

PR #284, opened 2022-07-12
Prepare the `release-2.7` branch

frouioui:
## Description

This Pull Request updates the `release-2.7` branch to prepare it for the future releases of `2.7.*`. It is part of the changes required to adopt a new release workflow that includes release branches.

It changes the `version.go` file and reverts the recent changes to vtorc tests (#282, #283) and the upgrade of dependencies (#280). The Pull Request #281 is also reverted.

@GuptaManan100, I have reverted the above Pull Request to restore the branch to its state when releasing `v2.7.1`. I have left the most recent changes to VTAdmin: #278 and #279. Let me know if this is an issue.

## Changes that are reverted

- #280 
- #281 
- #282 
- #283 

## Changes that are kept

- #278 
- #279


GuptaManan100:
@frouioui Shouldn't we start the branch with the tagged version 2.7 and just change the version information there. Then we know that the branch would be at exactly the state that we did the release on. We should then backport the relevant PRs. Just my thoughts on the matter, this is fine too. I'll add the vtorc test again though on that branch, just to be sure that works.

frouioui:
@GuptaManan100, I agree, this is a similar logic here. The two PRs #278 & #279 are already "backported"/included.

GuptaManan100 (approved):
Okay! Great

PR #293, opened 2022-07-21
Fix: `init_db.sql` file to be in sync with `main` on Vitess

GuptaManan100:
# Description

https://github.com/vitessio/vitess/pull/10080 changed the `init_db.sql` file on `main`. This PR makes the equivalent changes in the examples and tests in VTop


Verolop (approved):
lgtm

PR #295, opened 2022-07-28
CherryPick: Upgrade Vitess dependency to 14.0.0 (#280)

GuptaManan100:
## Description

This is a cherry-pick of #280 to release-2.7.

The upgrade of vitess images in test is reverted in a commit following the cherry-pick since we want `release-2.7` to continue to be tested for upgrading from v13 to v14 of Vitess.


.buildkite/pipeline.yml
```diff
@@ -7,10 +7,10 @@ steps:
   - name: "Upgrade Test"
     command:
     - apk add g++ make bash curl mysql mysql-client
-    - wget https://golang.org/dl/go1.17.2.linux-amd64.tar.gz
-    - tar -C /usr/local -xzf go1.17.2.linux-amd64.tar.gz
+    - wget https://golang.org/dl/go1.18.3.linux-amd64.tar.gz
+    - tar -C /usr/local -xzf go1.18.3.linux-amd64.tar.gz
     - export PATH=$PATH:/usr/local/go/bin
-    - rm go1.17.2.linux-amd64.tar.gz
+    - rm go1.18.3.linux-amd64.tar.gz
```
frouioui, line 13: Should we hold off and directly upgrade to `go1.18.4` `main` and `release-2.7`?

frouioui: This newer version contains security fixes that might impact the operator.

frouioui (approved):

.buildkite/pipeline.yml
GuptaManan100: Why would we need to hold off? Haven't we already upgraded the dependency of release-14.0. I can increment the tagged version to 14.0.1

frouioui: What I mean is, should we also upgrades the go version used in the operator to be `go1.18.4`

GuptaManan100: The operator itself is on 1.17. 🤣

frouioui: Ouch

frouioui: Maybe in the near future we want to upgrade to a newer version such a `go1.17.12` or directly `go.1.18.4`

GuptaManan100: Yeah for sure. Let me do both of those now. We can forward port those commits then.

GuptaManan100: Upgraded both.

frouioui (approved):

Issue #302, opened 2022-08-11
backup pod unable to start with failed to fetch service account token


michaellee8:
Consistently got this error event when trying to startup a vitess cluster on GKE Autopilot running k8s v1.22.10 and vitess 14.0.1, tried both operator version 2.7.1 and 2.7.2 both failed. The cluster config I am using is similar to the exampledb as provided.

```
MountVolume.SetUp failed for volume "kube-api-access-x6hpj" : failed to fetch token: serviceaccounts "vitess-operator" is forbidden: the UID in the bound object reference (4d8d96fd-b40a-484d-90ad-a55e46df2f01) does not match the UID in record. The object might have been deleted and then recreated
```

I have found that the error messages comes from https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/noderestriction/admission.go#L537 so I am not sure if this is a operator problem or a k8s problem.

michaellee8:
Some additional information:

I have tried v1.22 on minikube and it works, but v1.21 on GKE Autopilot stable and v1.22 on GKE Autopilot regular both got the exact same MountVolume.SetUp error. GKE Standard v1.20 works flawlessly through, will try to see if GKE Standard v1.22 works.

PR #303, opened 2022-08-12
fix: k8s 1.21+ bounded service account token

michaellee8:
In k8s 1.21+, bounded service account token are enfoced which requries a different prefix for service account token mounts, and are not removed in `reconcile_subcontroller.go`. My patch fixes the issue which allows the subcontroller to work.

Closes https://github.com/planetscale/vitess-operator/issues/302


michaellee8:
Still got the problem after this patch, no idea why, investigating.

deepthi:
@michaellee8 let us know once you have a working fix. At that time we can review and run CI.

michaellee8:
@deepthi It seems that even with my patch it still doesn't work, are you familiar with other places in the codebase that may cause the issue?

mattlord:
Closing this as stale

PR #324, opened 2022-10-01
VTOrc Deployment fix and miscellaneous changes.

GuptaManan100:
## Description

This PR fixes the VTOrc deployment for the vitess operator. The changes made recently to VTOrc make it backward in-compatible with Vitess operator. i.e. The latest VTOrc image won't work with the previous version of VTop and vice-versa.

The config file is totally dropped in this release and instead the use of flags is promoted. We have already introduced flags for all the possible configurations in https://github.com/vitessio/vitess/pull/11356. With these changes, VTOrc like other vitess components will be deployed on the 15000 port instead of the 3000 port before.

Apart from VTOrc changes, this PR also fixes the usage of deprecated flags which were removed in https://github.com/vitessio/vitess/pull/11336

A couple of flags in vtbackup which were already deprecated and unused have also been removed

The tests have also been changed to use `vtctldclient` from version `14.0.3` of Vitess instead of `vtctlclient` from `v13.0.0`

### Related Issues
- Fixes https://github.com/planetscale/vitess-operator/issues/326


GuptaManan100:
The backup tests are failing currently because we dropped the flag `backup_engine_implementation` from `vtbackup` binary, even though it uses it. The fix will be on the vitess repo.

deepthi:
> I have also taken the opportunity to rename the VTOrc deployment name from vitessOrchestrator to vtorc

This is the only thing I'll quibble with. We've tried to keep the names in the yaml config as descriptive as possible. That's why we have vitessDashboard, gateway etc. instead of vtctld and vtgate.
Let us change it back.

GuptaManan100:
@deepthi I changed it back 👍 and updated the description

frouioui (approved):

frouioui:
I am not seeing anything that seems off, looks good to me!

PR #358, opened 2023-01-13
Bug: Fix vtadmin resources

nianfei97:
Currently, the resources requests/limits for the vtadmin component is not reflected in the spec for vtadmin

This is because the resource spec is copied to a local variable and discarded instead of the actual container spec

Signed-off-by: Teh Nian Fei <tehnianfei@gmail.com>

### DIFF
--- a/pkg/operator/vtadmin/deployment.go
+++ b/pkg/operator/vtadmin/deployment.go
@@ -185,8 +185,6 @@ func UpdateDeployment(obj *appsv1.Deployment, spec *Spec) {
 	update.PodTemplateContainers(&obj.Spec.Template.Spec.InitContainers, spec.InitContainers)
 	update.PodTemplateContainers(&obj.Spec.Template.Spec.Containers, spec.SidecarContainers)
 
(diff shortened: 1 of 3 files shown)


GuptaManan100 (commented):
Yep, this looks like a bug. Thankyou for catching this. Could you add a test for this in the vtorc-vtadmin test that we have. You can change the config for the spec of vtadmin and verify in the test if the resources match the ones set.

nianfei97:
I have added a test to ensure that the vtadmin resources are set. Unfortunately, I couldn't seem to get the end2end tests running on my system, so I was unable to verify that the test works as expected.

GuptaManan100:
@nianfei97 I verified running the test locally, it works just as expected. There was a bug fix which was merged into main recently which is required to get the CI to pass, so I have merged main into the PR. Once the tests pass I shall merge the PR.

GuptaManan100 (approved):

GuptaManan100:
Thank-you for finding the bug and for your contributions @nianfei97! 🚀

PR #360, opened 2023-01-16
Remove removed flags in vtctld

GuptaManan100:
## Description

This PR removes the flags that were removed in https://github.com/vitessio/vitess/pull/12085.


frouioui (approved):
Encountered the same issue just now! LGTM

PR #361, opened 2023-01-18
Support upgrade of minor MySQL version

frouioui:
This PR adds the code that supports upgrade of minor MySQL versions. The operator runs `set global innodb_fast_shutdown = 0` in the mysqld container before it shutdown to enable proper upgrade.

Additionally, replica tablets are the first ones to be rolled out, followed by the primary.


pkg/operator/vttablet/pod.go
```diff
@@ -370,6 +370,16 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 	}
 }
 
+func detectMySQLVersionUpgradeDowngrade() *corev1.Lifecycle {
```
GuptaManan100: Shouldn't the name be `LifeCycleForMySQLVersionChange` or something along those line? What does the function `detect`?

pkg/apis/planetscale/v2/labels.go
```diff
@@ -37,6 +37,8 @@ const (
 	TabletTypeLabel = LabelPrefix + "/" + "tablet-type"
 	// TabletIndexLabel is the key for identifying the index of a Vitess tablet within its pool.
 	TabletIndexLabel = LabelPrefix + "/" + "tablet-index"
+	// TabletMySQLVersion is the key for identifying the version of the VTTablet's MySQL
+	TabletMySQLVersion = LabelPrefix + "/" + "tablet-mysql-version"
```
GuptaManan100, line 41: This is unused

GuptaManan100 (commented):
We might be able to use `obj.SetAnnotations`

PR #368, opened 2023-02-01
Makes VTOrc mandatory

GuptaManan100:
### Description

This PR makes the changes to the operator to always run VTOrc for a given keyspace as it is now a mandatory component of Vitess. This PR also removes code that repairs replication from the operator since this will be handled by VTOrc now.


test/endtoend/operator/101_initial_cluster_vtorc_vtadmin.yaml
```diff
@@ -71,15 +71,6 @@ spec:
...
     turndownPolicy: Immediate
-    vitessOrchestrator:
-      resources:
-        limits:
-          memory: 128Mi
-        requests:
-          cpu: 100m
-          memory: 128Mi
-      extraFlags:
-        recovery-period-block-duration: 5s
```
GuptaManan100, line 82: I have removed this to verify that we still start a VTOrc instance irrespective of whether it is part of the config or not.

frouioui (approved):

PR #371, opened 2023-02-03
Update Go dependency, compatibility table, Vitess dependency and fix upgrades

GuptaManan100:
### Description

This PR updates the go and Vitess dependencies, and updates the compatibility table to prepare the release of 2.9.0-rc1 version of the operator.

It also reverts some of the changes made in https://github.com/planetscale/vitess-operator/pull/361 wherein the operator would set `innodb_fast_shutdown=0` in vttablet container lifecycle. This solution wasn't working out and has been removed. Instead docs for how to upgrade the MySQL version are added in #373  

### Related Issues
- https://github.com/vitessio/vitess/issues/11749

PR #373, opened 2023-02-06
Update the release instructions after v16 code freeze

frouioui:
This PR is part of https://github.com/vitessio/vitess/issues/12233. It moves the "update vitess dependency" section of the release instruction to the "Cut Release" section. We now update the vitess dependency right before cutting the release, not during the code freeze.

This PR also updates the do_release script to update the `/deploy/operator.yaml` file before doing the release.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -19,6 +19,27 @@ build/Dockerfile.release
 go.mod
 ```
 
+### Update Compatibility Table
+
+Add a new entry for the planned minor version to the [compatibility table](https://github.com/planetscale/vitess-operator/blob/main/README.md#compatibility)
+in the README file.
+
+The recommended Kubernetes versions depend on the version of the Kubernetes
+client libs that Vitess Operator uses:
+
+https://github.com/planetscale/vitess-operator/blob/111ac173e1c473853e66e270486ba8a9a47ecc54/go.mod#L35
+
+The recommended server versions are those that match the minor version for the
+client libraries, as well as server versions that are +/- one minor version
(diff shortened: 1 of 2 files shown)


GuptaManan100:
I have a couple more things to add/update. Also this won't work until #371 is merged.

GuptaManan100 (approved):

docs/release-process.md
```diff
@@ -19,6 +19,27 @@ build/Dockerfile.release
...
 
+### Update Compatibility Table
+
+Add a new entry for the planned minor version to the [compatibility table](https://github.com/planetscale/vitess-operator/blob/main/README.md#compatibility)
+in the README file.
+
+The recommended Kubernetes versions depend on the version of the Kubernetes
+client libs that Vitess Operator uses:
+
+https://github.com/planetscale/vitess-operator/blob/111ac173e1c473853e66e270486ba8a9a47ecc54/go.mod#L35
```
mattlord, line 30: Do we want this to be a permanent/static link?

mattlord (approved):

GuptaManan100:
Force merging since the changes are only markdown and release script which don't affect the tests. The tests are currently failing and being worked on in #371

docs/release-process.md
frouioui: I think it is fine to keep it that way. The link redirects to an old commit but at least it will always point to the same line of code. That line:
```
// Pinned to kubernetes-1.16.2
```

Is visible in all the versions of our `go.mod`, with the `1.16.2` being the only thing we change.

PR #379, opened 2023-02-09
Fix tests in release-2.9

GuptaManan100:
### Description

This PR fixes the tests in release-2.9

### DIFF
--- a/test/endtoend/operator/101_initial_cluster.yaml
+++ b/test/endtoend/operator/101_initial_cluster.yaml
@@ -8,13 +8,13 @@ metadata:
   name: example
 spec:
   images:
-    vtctld: vitess/lite:v15.0.2
-    vtgate: vitess/lite:v15.0.2
-    vttablet: vitess/lite:v15.0.2
-    vtorc: vitess/lite:v15.0.2
-    vtbackup: vitess/lite:v15.0.2
+    vtctld: vitess/lite:v15.0.2-mysql80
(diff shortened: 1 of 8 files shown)


test/endtoend/operator/101_initial_cluster.yaml
```diff
@@ -8,13 +8,13 @@ metadata:
...
-    vtctld: vitess/lite:v15.0.2
-    vtgate: vitess/lite:v15.0.2
-    vttablet: vitess/lite:v15.0.2
-    vtorc: vitess/lite:v15.0.2
-    vtbackup: vitess/lite:v15.0.2
+    vtctld: vitess/lite:v15.0.2-mysql80
+    vtgate: vitess/lite:v15.0.2-mysql80
+    vttablet: vitess/lite:v15.0.2-mysql80
+    vtorc: vitess/lite:v15.0.2-mysql80
+    vtbackup: vitess/lite:v15.0.2-mysql80
```
frouioui, line 15: This should be ported over to `main` as well right?

frouioui (approved):

test/endtoend/operator/101_initial_cluster.yaml
GuptaManan100: No, on main we are now using v16 images and testing upgrade from them. So they are already from 8.0 to 8.0. Actually on main even the minor version is not being upgraded.

PR #391, opened 2023-03-02
Improve release instructions and script after the `2.9.0` release

frouioui:
This PR updates the release instructions and the release script after the `2.9.0` release.

- Instructions on what to do on `main` is added.
- The backup tests will now use the `mysql57` image.
- We no longer change the `operator.yaml` file during the release on the release branch but rather on `main`.
- The instruction output is fixed.


docs/release-process.md
```diff
@@ -103,6 +103,14 @@ Follow the instructions prompted by the `do_release.sh` script. You will need to
 > Make sure to Normal Merge the pull request i.e. merge the pull request with merge commit and not a squash merge. This is required because we create the tag
 from the pull request, so in order to have the tag on the release branche's history, it has to be a normal merge.
 
+##### On `main`
+
+Once you have done the release on the release branch, there are several steps to follow on `main`.
+
+- The `vitess/lite` image tag must be changed in [101_initial_cluster.yaml](..%2Ftest%2Fendtoend%2Foperator%2F101_initial_cluster.yaml). The latest Vitess release tag must be used.
+- We must copy the [operator.yaml](..%2Ftest%2Fendtoend%2Foperator%2Foperator.yaml) file we created during the release onto `main`.
```
GuptaManan100, line 111: We need to copy the `operator-latest.yaml` file into `operator.yaml` in `main` and change the operator version field to `2.9.0`, etc.

GuptaManan100 (approved):
Rest looks good to me

docs/release-process.md
frouioui: Good point

PR #398, opened 2023-03-20
Pin Docker image version in Buildkite

GuptaManan100:
### Description

This PR pins the docker version in buildkite because of recent bug we ran into. The bug has been described in detail in https://github.com/docker/cli/issues/4103.

This PR also updates the init_db.sql files for the examples after the changes made for super-read-only in https://github.com/vitessio/vitess/pull/12206

EDIT - The bug linked above had a response and the fix is to install `gcompat`. That has also been fixed in this PR.

Fixes: https://github.com/vitessio/vitess/issues/12678

### DIFF
--- a/runrunrun.sh
+++ b/runrunrun.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+for i in {1..500} ; do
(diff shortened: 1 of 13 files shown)


frouioui (approved):

rsajwani:
I think there is test failure still due  to some user issue ..

https://buildkite.com/planetscale/vitess-operator/builds/474#01870310-a19c-4865-b9de-ebc4e9abf9db

rsajwani:
should we do the change with 
deploy/example.yaml as well

frouioui:
> should we do the change with deploy/example.yaml as well

Yes we should

GuptaManan100:
@frouioui We'll have to make corresponding changes to the operator example in the Vitess repo too. Let's do that after we fix everything on this PR.

frouioui:
> @frouioui We'll have to make corresponding changes to the operator example in the Vitess repo too. Let's do that after we fix everything on this PR.

Yes, I have an action item for this

runrunrun.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
```
frouioui, line 2: ah, forgot to remove this before the merge. I'll fix it.

PR #420, opened 2023-05-08
Complete the Reshard before deleting the shard

GuptaManan100:
### Description

We should complete the reshard process before deleting the shard. If we don't then the workflow remains to be seen in the `vtctlclient Workflow customer listall` command and Completing it after deleting the shards fails with following error -

```
Reshard Error: rpc error: code = Unknown desc = node doesn't exist: /vitess/example/global/keyspaces/customer/shards/-/Shard
E0508 17:32:26.613887   52392 main.go:105] remote error: rpc error: code = Unknown desc = node doesn't exist: /vitess/example/global/keyspaces/customer/shards/-/Shard
```

This PR fixes the test to complete the workflow first.

### DIFF
--- a/test/endtoend/upgrade_test.sh
+++ b/test/endtoend/upgrade_test.sh
@@ -163,6 +163,14 @@ COrder
 +----------+-------------+----------+-------+
 EOF
 
+  # Complete the reshard process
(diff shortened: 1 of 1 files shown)


rohit-nayak-ps (approved):
lgtm

PR #432, opened 2023-06-14
Add a new field called "index" to VitessShardTabletPool as part of x-kubernetes-list-map-keys

### LINKED ISSUE #428: Feature Request: Allow more than one VitessShardTabletPool with a certain tablet type in a given cell
### Feature Description

We want to create [more than one VitessShardTabletPool with a certain tablet type in a given cell](https://github.com/planetscale/vitess-operator/blob/685391b42447b97e1e9dcb7640ced2a8ca01c291/pkg/apis/planetscale/v2/vitessshard_types.go#L105-L116).
This would allow us to support multiple unmanaged MySQL replicas with minimal changes like the one below.

```yaml
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: example
spec:
  cells:
    - name: zone1
  keyspaces:
    - name: unsharded
      partitionings:
      - equal:
          parts: 1
          shardTemplate:
            tabletPools:
              - cell: zone1
                type: externalreplica
                replicas: 2
                vttablet:
                externalDatastore:
                  host: replica-host-1
              - cell: zone1
                type: externalreplica
                replicas: 2
                vttablet:
                externalDatastore:
                  host: replica-host-2
```

With this feature, we don't have to set up a (vtgate, vttablet) pair for each additional MySQL replica like the following, which requires client redirection to a different endpoint.

```yaml
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: example
spec:
  cells:
    - name: zone1
    - name: zone2
  keyspaces:
    - name: unsharded
      partitionings:
        - equal:
            parts: 1
            shardTemplate:
              tabletPools:
                - cell: zone1
                  type: externalreplica
                  replicas: 2
                  vttablet:
                  externalDatastore:
                    host: replica-host-1
                - cell: zone2
                  type: externalreplica
                  replicas: 2
                  vttablet:
                  externalDatastore:
                    host: replica-host-2
```

### Use Case(s)

- When you want to build a redundant system with unmanaged tablets

### Discussion

- https://vitess.slack.com/archives/CNE9WP677/p1685090821858559
- https://vitess.slack.com/archives/CNE9WP677/p1685083619890119

yoheimuta:
## Description

This PR introduces a new field named "index" to the `x-kubernetes-list-map-keys` of TabletPools. It enables the presence of multiple TabletPool instances with a specific tablet type in a given cell.

By assigning a default value of 0 to the index, the existing users and most use cases remain unaffected. The index value can be explicitly set to a value greater than 0 only when building a redundant system with unmanaged tablets.

## Related Issue(s)

- Fixes #428

### DIFF
--- a/pkg/operator/vttablet/uid.go
+++ b/pkg/operator/vttablet/uid.go
@@ -43,11 +43,30 @@ lead time to develop a smarter way to handle tablet identity and MySQL server
 IDs in Vitess itself.
 
 WARNING: DO NOT change the behavior of this function, as that may result in
-         the deletion and recreation of all tablets.
+
+	the deletion and recreation of all tablets.
 */
 func UID(cellName, keyspaceName string, shardKeyRange planetscalev2.VitessKeyRange, tabletPoolType planetscalev2.VitessTabletPoolType, tabletIndex uint32) uint32 {
 	h := md5.New()
 	fmt.Fprintln(h, cellName, keyspaceName, shardKeyRange.String(), string(tabletPoolType), tabletIndex)
+	sum := h.Sum(nil)
+	return binary.BigEndian.Uint32(sum[:4])
+}
+
+/*
+UIDWithPoolIndex function generates a 32-bit unsigned integer similar to the UID function above.
+
+However, it additionally takes the poolIndex as an input.
+This allows the generation of a unique UID for a tablet that belongs to a different pool
+but shares other common attributes.
+
+To preserve the existing UID, it is recommended to use the UID function instead of this function
+when the poolIndex is set to its default value of 0.
+*/
+func UIDWithPoolIndex(cellName, keyspaceName string, shardKeyRange planetscalev2.VitessKeyRange,
+	tabletPoolType planetscalev2.VitessTabletPoolType, tabletIndex uint32, poolIndex uint32) uint32 {
+	h := md5.New()
+	fmt.Fprintln(h, cellName, keyspaceName, shardKeyRange.String(), string(tabletPoolType), tabletIndex, poolIndex)
 	sum := h.Sum(nil)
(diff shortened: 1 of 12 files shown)


deepthi:
`index` seems like an anti-pattern. any code that is fetching the objects will get a slice with each object having its own index in that slice. Why not use `name` as the identifier?

yoheimuta:
Thank you for the review! I will address these two points and make the necessary fixes.

yoheimuta:
@deepthi I've made the fixes about naming and limitation: https://github.com/planetscale/vitess-operator/pull/432/commits/a042c49f75c53a2d2d54b975a9a43b59a64d21ce https://github.com/planetscale/vitess-operator/pull/432/commits/36957219fe4b5338e86e027d232ff244f7b38193

pkg/operator/vttablet/uid.go
```diff
@@ -43,11 +43,30 @@ lead time to develop a smarter way to handle tablet identity and MySQL server
 IDs in Vitess itself.
 
 WARNING: DO NOT change the behavior of this function, as that may result in
-         the deletion and recreation of all tablets.
+
```
frouioui, line 46: nit: we can remove this extra line i think

frouioui (approved):
LGTM! thank you @yoheimuta

pkg/operator/vttablet/uid.go
yoheimuta: Thank you for your review!
I fixed it: [62ea41d](https://github.com/planetscale/vitess-operator/pull/432/commits/62ea41ded28f11b4fa183f6b2a35f7f9c717ecad)

GuptaManan100 (approved):
lgtm

PR #483, opened 2023-10-04
Support safe MySQL upgrades

frouioui:
This Pull Request adds code logic that will set the `global.innodb_fast_shutdown` MySQL variable to `0` if there is a need for a safe MySQL upgrade.

Part of https://github.com/vitessio/vitess/issues/14160

### DIFF
--- a/pkg/controller/vitessshardreplication/reconcile_drain.go
+++ b/pkg/controller/vitessshardreplication/reconcile_drain.go
@@ -591,7 +591,7 @@ func safeMysqldUpgrade(currentImage, desiredImage string) (bool, error) {
 	// as patch release could also include on-disk data format changes. This happened a number of times
 	// in practice as well, so this concern is real.
 	//
-	// MySQL though as announced a new release strategy, see:
+	// MySQL though has announced a new release strategy, see:
 	// https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/
 	//
 	// With that release strategy, it will become possible that patch releases will be safe to downgrade
--- a/pkg/controller/vitessshardreplication/reconcile_drain.go
+++ b/pkg/controller/vitessshardreplication/reconcile_drain.go
@@ -261,7 +261,7 @@ func (r *ReconcileVitessShard) reconcileDrain(ctx context.Context, vts *planetsc
 		}
 	}
 
-	// 4. Check if we need to any operations like disabling fast shutdown
+	// 4. Check if we need to perform any operations like disabling fast shutdown
 	// for upgrades here.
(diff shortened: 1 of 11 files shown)


pkg/controller/vitessshardreplication/reconcile_drain_test.go
```diff
@@ -0,0 +1,150 @@
...
+			current:   "docker.io/vitess/mysql:8.0.23",
+			desired:   "docker.io/vitess/mysql:8.0.23",
+			needsSafe: false,
+		},
+		{
+			name:      "equal version with different registry",
+			current:   "docker.io/vitess/mysql:8.0.23",
+			desired:   "docker.io/vitess/mysql:8.0.23",
+			needsSafe: false,
+		},
```
GuptaManan100, line 58: They are the same cases

GuptaManan100 (approved):
Rest LGTM!

pkg/controller/vitessshardreplication/reconcile_drain_test.go
frouioui: Fixed via c30e759

pkg/controller/vitessshardreplication/reconcile_drain.go
```diff
@@ -255,8 +261,16 @@ func (r *ReconcileVitessShard) reconcileDrain(ctx context.Context, vts *planetsc
 		}
 	}
 
+	// 4. Check if we need to any operations like disabling fast shutdown
```
deepthi, line 264:
```suggestion
	// 4. Check if we need to perform any operations like disabling fast shutdown
```

pkg/controller/vitessshardreplication/reconcile_drain.go
```diff
@@ -458,6 +472,151 @@ func candidatePrimary(ctx context.Context, wr *wrangler.Wrangler, shard *topo.Sh
...
+	if dstParts[0] == curParts[1] && dstParts[1] < curParts[1] {
+		return false, fmt.Errorf("cannot downgrade minor version from %s to %s", current, desired)
+	}
+
+	// Alright, here it gets more tricky. MySQL has had a complicated release history. For the 8.0 series,
+	// up to 8.0.34 at least (known at this point), it was not supported to downgrade patch releases
+	// as patch release could also include on-disk data format changes. This happened a number of times
+	// in practice as well, so this concern is real.
+	//
+	// MySQL though as announced a new release strategy, see:
```
deepthi, line 594:
```suggestion
	// MySQL though has announced a new release strategy, see:
```

deepthi (approved):
Nice work! I only found a couple of typos in comments. Please fix them and then merge.

PR #406, opened 2023-04-06
Add support for container lifecycle hooks in VTGate

stankevich:
Can be used to add `postStart` and `preStop` [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) to the VTGate container. If a longer `preStop` hook execution is needed, this PR also adds support for customizing `terminationGracePeriodSeconds` of the VTGate pods.

Example use cases: delaying the shutdown of VTGate pods to let the application complete any in-flight queries, or notifying the application to gracefully disconnect from the VTGate pod that is being shut down.

Similar to https://github.com/planetscale/vitess-operator/pull/192.

PR #500, opened 2023-11-06
Add support for customizing terminationGracePeriodSeconds of the VTTablet pods

yoheimuta:
This PR introduces support for customizing the terminationGracePeriodSeconds of the VTTablet pods. Currently, this value is set to 30 minutes.

Having [a fairly high terminationGracePeriodSeconds](https://github.com/planetscale/vitess-operator/blob/9cb55b422e21e6fbea1c3614c4bfdedd3a2f963f/pkg/operator/vttablet/constants.go#L51) can lead to issues when adding sidecars that don't handle SIGTERM properly, resulting in long delays during the termination process. At least, this default constant value is not suitable for the use case of unmanaged tablets.

Similar to https://github.com/planetscale/vitess-operator/pull/406.


pkg/operator/vttablet/pod.go
```diff
@@ -313,7 +313,11 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 		obj.Spec.SecurityContext.FSGroup = pointer.Int64Ptr(planetscalev2.DefaultVitessFSGroup)
 	}
 
-	obj.Spec.TerminationGracePeriodSeconds = pointer.Int64Ptr(terminationGracePeriodSeconds)
+	if spec.Vttablet.TerminationGracePeriodSeconds != nil {
+		obj.Spec.TerminationGracePeriodSeconds = spec.Vttablet.TerminationGracePeriodSeconds
+	} else {
+		obj.Spec.TerminationGracePeriodSeconds = pointer.Int64Ptr(terminationGracePeriodSeconds)
```
mattlord, line 319: Nit, but I feel like we should rename this variable to `defaultTerminationGracePeriodSeconds` now.

mattlord (approved):
Thanks, @yoheimuta ! I just had one nit about the existing var name. Let me know what you think.

frouioui (approved):
Looks good to me, thank you @yoheimuta

frouioui:
Closing and re-opening to re-trigger the CI tests that are in `Expected`

PR #502, opened 2023-11-06
update v18 vitess version to vtop main

harshit-gangal:


frouioui (approved):
Looks good to me, I restarted the failing test it was not finding the `planetscale/vitess-operator:v2.11.0` image which i think is just a timeline issue

GuptaManan100 (approved):

PR #511, opened 2023-12-11
Add tolerable replication lag value

GuptaManan100:
### Description

In the operator, we are calling PlannedReparentShard from the wrangler and an argument to that was recently added to it that specifies the amount of replication lag considered to be tolerable when initiating a PRS.

The vitess change was made in the PR https://github.com/vitessio/vitess/pull/14634


deepthi (commented):
Should this made a Keyspace/Shard property in the operator so that people can override it?
Or is that a bad idea?

frouioui (approved):

GuptaManan100:
Vitess operator only issues PlannedReparentShard when it needs to drain a tablet. The timeout for that PRS is not yet configurable, so I think it is okay to have a hard-coded value for now. If we feel the need to make them configurable, we can add them as a follow-up.

systay (approved):

Issue #383, opened 2023-02-14
AWS IAM Role for Service Account support


klagroix:
We're trying to use vitess-operator with an S3 backup spec defined in our VitessCluster manifest. Example as follows:

```
spec:  
  backup:
    engine: xtrabackup
    locations:
      - s3:
          bucket: my-s3-bucket
          region: us-east-1
```

From the logs, it appears that `vttablet` attempts to read the backup S3 bucket to see it it needs to restore from the latest backup. As such, we need to provide AWS credentials to allow the pod access to read the S3 bucket. 

Typically, we use IAM Roles for Service Accounts (IRSA) which allows annotating a Service Account with an IAM Role arn: https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html
* Using IRSA, when a pod is created using this annotated service account, `AWS_*` environment variables are added to the pod so the pod can authenticate and use the the IAM role.

The issue here is that vitess-operator seems to see these `AWS_` environment variables on the pod and attempts to `$patch: delete` these variables.

Example log taken from vitess-operator:

```
[...]

Issue #514, opened 2023-12-20
Operator misbehaving when a imagePullSecrets is defined on the default service account of the namespace


niv8:
When adding an **imagePullSecrets** in the **default** service account on the namespace the operator is installed on, the operator will bring all the pods up and then it will continuously recreate the vttablet pods...
It seems like the addition of imagePullSecrets by k8s is making the operator think that the pod spec has changed...

frouioui:
Hey @niv8, thank you for opening this issue. Has this been fixed with #518?

PR #518, opened 2024-01-04
Pass extra env to all vttablet containers

jhampson-dbre:
This allows for the mysqld-exporter to be configured with the same
environment variables as the vttablet and mysqld containers. This
is useful for configuring an AWS IAM-based service account, which
automatically injects additional AWS_* environment variables into
the pod. Without this, the operator continuously restarts the vttablet
pod due to trying to remove the AWS_* environment variables from
the mysqld-exporter container.

Related to #383

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -6675,7 +6675,7 @@ created for this component.</p>
 <td>
 <p>ExtraEnv can optionally be used to override default environment variables
 set by the operator, or pass additional environment variables.
-These values are applied to both the vttablet and mysqld containers.</p>
+These values are applied to the vttablet, mysqld, and mysqld-exporter containers.</p>
 </td>
 </tr>
 <tr>
--- a/pkg/operator/vttablet/env_vars.go
+++ b/pkg/operator/vttablet/env_vars.go
@@ -17,6 +17,7 @@ limitations under the License.
 package vttablet
 
 import (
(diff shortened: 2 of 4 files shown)


mattlord (approved):
Thank you for the contribution, @jhampson-dbre ! This seems like a safe and sensible change to me.

I wonder if this is also related to: https://github.com/planetscale/vitess-operator/issues/514

dbmurphy:
> Thank you for the contribution, @jhampson-dbre ! This seems like a safe and sensible change to me.
> 
> I wonder if this is also related to: #514

Yes, this seems like we need some way to allow "extraVars" diffs to be ignored and not assume the pod needs to be recreated I suppose.  This likely is the same thing on #514, but it seems a temporary fix, and any new containers in the pod in the future could hit this again.

jhampson-dbre:
I believe setting pull secret on the service account causes the pod's  spec.imagePullSecrets to be defined automatically. So similar issue where some k8s magic causes some implicit changes to the pod, but in this case I think the option for defining the pull secret in the VitessCluster spec should cause the operator expected state to match. In our case with env vars, they aren't passed through to all the containers in the pod so the spec will never match.

frouioui (approved):

frouioui:
Thank you for this contribution, it looks great to me!

PR #541, opened 2024-03-12
Ensure that binary logs for PITR are restored to a shared location

mattlord:
Without doing this, the binary logs cannot be accessed by `mysqlbinlog` in the `mysqld` container.

Related issues and PRs:
  - https://github.com/vitessio/vitess/pull/15451
  - https://github.com/vitessio/vitess/issues/14765
  - https://github.com/vitessio/vitess/issues/15452


shlomi-noach (commented):
Nice! Should we also provide the flag in `yaml` files?

mattlord:
> Nice! Should we also provide the flag in `yaml` files?

Yeah. I *think* this does it. https://github.com/planetscale/vitess-operator/pull/541/commits/e2e5e8b1e749b6f550ee72e77dce2f8ef326408d

shlomi-noach:
> Yeah. I think this does it.

How is the value being set, and to what specific value?

mattlord:
> How is the value being set, and to what specific value?

The user would specify the flag and value in their cluster yaml definition using the `extraFlags` parameter, just as they do for `mysqld` flags, e.g. If they don't specify a value then we enforce the default within the operator.

shlomi-noach (approved):
Seems to me like it's feature complete and can be taken out of `Draft`?

mattlord:
> > How is the value being set, and to what specific value?
> 
> The user would specify the flag and value in their cluster yaml definition using the `extraFlags` parameter, just as they do for `mysqld` flags, e.g. If they don't specify a value then we enforce the default within the operator.

The flag ended up being for `vttablet` and `vtbackup`, not `mysqlctld` (although vtbackup is a modified mysqlctld). I will leave the mysqlctld extra flags support though as that may come to be useful.

pkg/operator/vttablet/pod.go
```diff
@@ -87,6 +87,11 @@ func UpdatePod(obj *corev1.Pod, spec *Spec) {
 		key = strings.TrimLeft(key, "-")
 		vttabletAllFlags[key] = value
 	}
+	// Ensure that binary logs are restored to/from a location that all containers
+	// in the pod can access if no location was explicitly provided.
+	if _, ok := vttabletAllFlags["builtinbackup-incremental-restore-path"]; !ok {
+		vttabletAllFlags["builtinbackup-incremental-restore-path"] = vtDataRootPath
+	}
```
frouioui, line 94: What would happen if the path specified in `--builtinbackup-incremental-restore-path` is not accessible to all containers in the pod?

frouioui: I am guessing it is up to the user to set the same value on all components too? `mysqlctl`, `vttablet` and `vtbackup`

pkg/operator/vttablet/mysqlctld.go
```diff
@@ -29,6 +29,7 @@ const (
 	vtRootInitScript = `set -ex
 mkdir -p /mnt/vt/bin
 cp --no-clobber /vt/bin/mysqlctld /mnt/vt/bin/
+cp --no-clobber $(command -v mysqlbinlog) /mnt/vt/bin/ || true
```
frouioui, line 32: The directory `/mnt/.../` is shared across all the containers in the pod I am assuming? In which case, this line would resolve what you wrote in the PR's description:

> `/tmp` is not a shared mount point within the pod so when `mysqlbinlog` subsequently tries to read them from within the mysqld container it cannot find them in its container's `/tmp` directory and it fails with an error

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -315,6 +315,19 @@ type MysqldSpec struct {
...
+	// Resources specify the compute resources to allocate for just the MySQL Control Daemon.
+	Resources corev1.ResourceRequirements `json:"resources"`
+
+	// ExtraFlags can optionally be used to override default flags set by the
+	// operator, or pass additional flags to mysqlctld. All entries must be
+	// key-value string pairs of the form "flag": "value". The flag name should
+	// not have any prefix (just "flag", not "-flag"). To set a boolean flag,
+	// set the string value to either "true" or "false".
+	ExtraFlags map[string]string `json:"extraFlags,omitempty"`
+}
```
frouioui, line 329: Unless I am missing something, this does not seem to be used anywhere. I think we should remove it to avoid unrequired changes

pkg/operator/vttablet/pod.go
mattlord: The same thing that happens now to every user. It doesn't work. PITR does not generally work in the operator.

pkg/operator/vttablet/mysqlctld.go
mattlord: This is simply about copying the mysqlbinlog binary from the vitess/lite container image to the mysqlctld/vtbackup pod, as it looks like we'll need to keep that around in the lite image because the MySQL images do not contain that binary and it's needed for PITR.

pkg/apis/planetscale/v2/vitessshard_types.go
mattlord: I'll remove it if it indeed is not needed after I do some testing to confirm that things are now working.

frouioui: Coming back on what I said. I think it would be nice to keep that around so that people can use custom mysqlctld flags.

test/endtoend/utils.sh
```diff
@@ -103,17 +103,22 @@ function takeBackup() {
 
   sleep 10
 
+  mysql 'commerce:-' -e "flush binary logs" || echo "flushing binlogs failed"
```
shlomi-noach: The incremental backup will `flush binary logs` and there is no need to to that manually. If anything, it's better to _not_ `flush binary logs` so that the backup's behavior is better tested.

test/endtoend/utils.sh
```diff
@@ -71,17 +71,83 @@ function printMysqlErrorFiles() {
...
 
   # Issue the BackupShard command to vtctldclient.
-  vtctldclient BackupShard "$keyspaceShard"
+  vtctldclient BackupShard "${keyspaceShard}"
 
-  if [[ $? -ne 0 ]]; then
-    echo "Backup failed"
+  for i in {1..600} ; do
+    out=$(kubectl get vtb -n example --no-headers | wc -l)
+    if echo "${out}" | grep -c "${finalBackupCount}" >/dev/null; then
```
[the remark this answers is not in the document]
mattlord: Good catch! Fixed.

PR #451, opened 2023-07-25
Expose flags to configure TLS

ggalakhov:
### Related Issue: 
[Feature Request: configure operator to communicate with Vitess components using TLS #449](https://github.com/planetscale/vitess-operator/issues/449)

PR #573, opened 2024-07-11
[release-2.10] Expose flags to configure TLS

jcarrig:
Backport of https://github.com/planetscale/vitess-operator/pull/451

### DIFF
--- a/pkg/operator/controllermanager/flags.go
+++ b/pkg/operator/controllermanager/flags.go
@@ -72,10 +72,12 @@ func InitFlags() {
 	}
 	vttabletFlags.VisitAll(func(f *pflag.Flag) {
 		_, isRequired := tlsFlags[f.Name]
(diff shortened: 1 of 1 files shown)


mattlord:
@jcarrig 2.10 was paired with Vitess 17.0 which is now EOL. We won't be doing another release of 2.10. Does this still have value to you given that? If so, it certainly doesn't hurt anything project side. Thanks!

jcarrig:
> @jcarrig 2.10 was paired with Vitess 17.0 which is now EOL. We won't be doing another release of 2.10. Does this still have value to you given that? If so, it certainly doesn't hurt anything project side. Thanks!

hi @mattlord, yes it would give value as we could build an image from the release branch if this gets merged

mattlord (approved):

frouioui (approved):

Issue #577, opened 2024-07-16
Enhancement: Add an e2e test to ensure unmanaged tablet runs on the future Vitess operator


yoheimuta:
## Description

Currently, we don't have an e2e test that ensures unmanaged tablets will run on the next version of the operator. This can cause issues when Vitess components remove flags and the operator misses the breaking changes.

To prevent this, we should have an e2e test that sets up and checks the cluster using an external datasource.

## Related Issue(s)

- https://github.com/planetscale/vitess-operator/issues/575

PR #576, opened 2024-07-16
Remove 'vreplication_tablet_type' flag to run unmanaged vttablet v20

yoheimuta:
## Description

This PR removes a flag called `vreplication_tablet_type` that is deleted on Vitess v20.
See https://github.com/vitessio/vitess/blob/main/changelog/20.0/20.0.0/release_notes.md#vreplication-tablet-type-deletion.

## Related Issue(s)

fixes: https://github.com/planetscale/vitess-operator/issues/575

Issue #575, opened 2024-07-16
Bug Report: Unable to run unmanged vttablet v20 due to "Error: unknown flag: --vreplication_tablet_type"


yoheimuta:
### Overview of the Issue

We've encountered the following error when trying to run **unmanaged** vttablet v20 on vitess-operator:v2.13.0.

```
Error: unknown flag: --vreplication_tablet_type
--
  | Usage:
  | vttablet [flags]
...
```

### Reproduction Steps

1. Set up a cluster running vitess-operator:v2.13.0.

2. Deploy the following yaml:

```yaml
# The following example is minimalist. The security policies
# and resource specifications are not meant to be used in production.
# Please refer to the operator documentation for recommendations on
# production settings.
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: example
spec:
  images:
    vtctld: vitess/lite:latest
    vtadmin: vitess/vtadmin:latest
    vtgate: vitess/lite:latest
    vttablet: vitess/lite:latest
    vtbackup: vitess/lite:latest
    vtorc: vitess/lite:latest
    mysqld:
      mysql80Compatible: mysql:8.0.30
    mysqldExporter: prom/mysqld-exporter:v0.11.0
  cells:
  - name: zone1
    gateway:
      authentication:
        static:
          secret:
            name: example-cluster-config
            key: users.json
      replicas: 1
      resources:
        requests:
          cpu: 10m
          memory: 128Mi
        limits:
          memory: 256Mi
  vitessDashboard:
    cells:
    - zone1
    extraFlags:
      security_policy: read-only
    replicas: 1
    resources:
      limits:
        memory: 128Mi
      requests:
        cpu: 10m
        memory: 64Mi

  keyspaces:
  - name: commerce
    durabilityPolicy: none
    turndownPolicy: Immediate
    vitessOrchestrator:
      resources:
        limits:
          memory: 128Mi
        requests:
          cpu: 10m
          memory: 64Mi
      extraFlags:
        recovery-period-block-duration: 5s
    partitionings:
    - equal:
        parts: 1
        shardTemplate:
          databaseInitScriptSecret:
            name: example-cluster-config
            key: init_db.sql
          tabletPools:
          - cell: zone1
            type: externalmaster
            replicas: 1
            vttablet:
              extraFlags:
                db_charset: utf8mb4
              resources:
                limits:
                  memory: 256Mi
                requests:
                  cpu: 10m
                  memory: 128Mi
[...]

PR #578, opened 2024-07-16
Add unmanaged_tablet_test.sh to run a new e2e build on CI

yoheimuta:
## Description

This PR adds a new e2e test to ensure unmanaged tablet runs on the latest operator.

You see the following output on my personal build space:

- Test failed due to https://github.com/planetscale/vitess-operator/issues/575
   - Link: https://buildkite.com/yoheimuta/vitess-operator/builds/26#0190b9d3-8f76-4f75-99c1-e68baae22494/54-575
- Test passed after cherry-picking https://github.com/planetscale/vitess-operator/pull/576
   - Link: https://buildkite.com/yoheimuta/vitess-operator/builds/27
- 📝 I needed to apply [a tweak](https://github.com/planetscale/vitess-operator/commit/e74b533421eff3d4417651da3e9ede66638eb608) to the PR to run a build on my space due to a machine resource issue.

## Related Issue(s)

fixes: https://github.com/planetscale/vitess-operator/issues/577

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -107,7 +107,7 @@ from the pull request, so in order to have the tag on the release branche's hist
 
 ##### Update the test output
 
-The `upgrade_test.sh`, `backup_restore_test.sh` and `vtorc_vtadmin_test.sh` files must be updated with the proper release increment. Change the `verifyVtGateVersion` function calls to use the proper version (current version being released and latest previous version (only used in `upgrade_test.sh`)).
+The `upgrade_test.sh`, `backup_restore_test.sh`, `vtorc_vtadmin_test.sh` and `unmanaged_tablet_test.sh` files must be updated with the proper release increment. Change the `verifyVtGateVersion` function calls to use the proper version (current version being released and latest previous version (only used in `upgrade_test.sh`)).
 
 ##### CI Failures
 
--- a/test/endtoend/unmanaged_tablet_test.sh
+++ b/test/endtoend/unmanaged_tablet_test.sh
@@ -19,7 +19,7 @@ function get_started_unmanaged() {
     sleep 10
     echo "Creating vschema and commerce SQL schema"
 
-    ./pf_vtadmin.sh > /dev/null 2>&1 &
+    ./pf.sh > /dev/null 2>&1 &
     sleep 5
 
     waitForKeyspaceToBeServing commerce - 0
--- a/test/endtoend/unmanaged_tablet_test.sh
+++ b/test/endtoend/unmanaged_tablet_test.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+source ./tools/test.env
+source ./test/endtoend/utils.sh
+
+# get_started_unmanaged:
+function get_started_unmanaged() {
+    echo "Apply latest operator-latest.yaml"
+    kubectl apply -f "operator-latest.yaml"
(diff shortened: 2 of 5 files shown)


docs/release-process.md
```diff
@@ -107,7 +107,7 @@ from the pull request, so in order to have the tag on the release branche's hist
 
 ##### Update the test output
 
-The `upgrade_test.sh`, `backup_restore_test.sh` and `vtorc_vtadmin_test.sh` files must be updated with the proper release increment. Change the `verifyVtGateVersion` function calls to use the proper version (current version being released and latest previous version (only used in `upgrade_test.sh`)).
+The `upgrade_test.sh`, `backup_restore_test.sh`, `vtorc_vtadmin_test.sh` and `unmanaged_tablet_test.sh` files must be updated with the proper release increment. Change the `verifyVtGateVersion` function calls to use the proper version (current version being released and latest previous version (only used in `upgrade_test.sh`)).
```
frouioui, line 110: note to self, we will also need to update the [vitess-releaser](https://github.com/vitessio/vitess-releaser) with the new file

test/endtoend/unmanaged_tablet_test.sh
```diff
@@ -0,0 +1,102 @@
...
+      exit 1
+    fi
+
+    insertWithRetry
+
+    assertSelect ../common/select_commerce_data.sql "commerce" << EOF
+Using commerce
+Customer
++-------------+--------------------+
+| customer_id | email              |
```
frouioui, line 57: Not necessarily in this Pull Request, but ideally we should extract some of the reused code into sharded shell functions. For instance all the code between line 28 and line 82 is mostly the same as in the four other tests. The function `get_started()` (in `utils.sh`) already extract all this code, but makes assumptions that are not always valid for every test, like the fact that vtorc is always here and that we expect a fixed number of vttablets.

frouioui:
It seems the test is failing because the branch does not contain the bug fix. I will merge `main` into this branch, which should turn the test green.

test/endtoend/unmanaged_tablet_test.sh
```diff
@@ -0,0 +1,102 @@
...
+    checkPodStatusWithTimeout "example-zone1-vtctld(.*)1/1(.*)Running(.*)"
+    checkPodStatusWithTimeout "example-zone1-vtgate(.*)1/1(.*)Running(.*)"
+    checkPodStatusWithTimeout "example-etcd(.*)1/1(.*)Running(.*)" 3
+    checkPodStatusWithTimeout "example-vttablet-zone1(.*)1/1(.*)Running(.*)"
+
+    sleep 10
+    echo "Creating vschema and commerce SQL schema"
+
+    ./pf_vtadmin.sh > /dev/null 2>&1 &
+    sleep 5
```
frouioui, line 23: Since we are not defining vtadmin in the yaml file, we should use the normal port forwarding instead of the one that includes vtadmin.

```suggestion
    ./pf.sh > /dev/null 2>&1 &
    sleep 5
```

frouioui (approved):
Looks good to me! Thanks a bunch for adding this test.

deepthi (approved):
I agree with @frouioui about refactoring duplicated code.

frouioui:
I am marking the new workflow as requied.

yoheimuta:
Thanks!

PR #589, opened 2024-07-25
Extend ability to set topologySpreadhConstraints to all component workloads

relu:
This should allow users to configure topologySpreadhConstraints for all operator-managed workloads.


pkg/apis/planetscale/v2/etcdlockserver_types.go
```diff
@@ -197,6 +197,12 @@ type EtcdLockserverTemplate struct {
 	// +kubebuilder:validation:Schemaless
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// TopologySpreadConstraint can optionally be used to
+	// specify how to spread etcd pods among the given topology
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
```
frouioui, line 205: Any reason why `TopologySpreadConstraint` was added to only a few resources (`etcd`, `VitessDashboard` ,`VtAdmin`, `VtOrc`) ?

frouioui (commented):
the `operator-latest.yaml` file has to be updated with the latest CRD too. Additionally, a follow-up PR will have to be made on the vitess repository to update the given `operator.yaml` file in the examples.

pkg/apis/planetscale/v2/etcdlockserver_types.go
relu: Those were the only resources that didn't have it already.

It was implemented previously in:
* VitessCell (configures VtGate resources): https://github.com/planetscale/vitess-operator/blob/1e0acc1444d166434da2d822e9232e57f78324f0/pkg/apis/planetscale/v2/vitesscell_types.go#L202
* VitessShard (configures VtTablet resources): https://github.com/planetscale/vitess-operator/blob/1e0acc1444d166434da2d822e9232e57f78324f0/pkg/apis/planetscale/v2/vitessshard_types.go#L277

I think with these changes we are covering all of them now. Unless I've missed something? 🤔

pkg/operator/contenthash/topology_spread_constraints_test.go
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright 2020 PlanetScale Inc.
```
mattlord, line 2: We should use 2024 for new files.

pkg/operator/update/podspec.go
```diff
@@ -302,20 +301,10 @@ func TopologySpreadConstraints(dst *[]corev1.TopologySpreadConstraint, src []cor
...
-			srcMap = map[string]string{}
-		}
 		// If this item is already there, update it.
 		for dstIndex := range *dst {
 			dstObj := &(*dst)[dstIndex]
-			dstMap, dstErr := metav1.LabelSelectorAsMap(dstObj.LabelSelector)
-			if dstErr != nil {
-				dstMap = map[string]string{}
-			}
-			if reflect.DeepEqual(dstMap, srcMap) {
```
mattlord, line 318: I'm not sure why this code was added, so also not sure that this is safe. Reads to me like this was primarily about handling nils and invalid/unexpected data which we lose here. Can you please explain the reasoning?

mattlord (commented):
Thanks, @relu ! It looks good to me. I only had a few minor questions and suggestions. I can come back to this quickly when you have a chance to respond.

Thanks again!

frouioui:
Hey @relu! Are you still working on this? It would be nice to include this in the upcoming `v2.14.0` release of the vitess-operator if we can. The RC-1 is scheduled in a week, and GA late October.

pkg/operator/update/podspec.go
relu: The way this was intended to work was that the list of `corev1.TopologySpreadConstraint` items would be matched exclusively on the `labelSelector`, causing any previously existing items having the same `labelSelector` to be overridden.
This behavior doesn't make sense because the `labelSelector` will most likely be the same across all topology spread constraint items. Instead, we should only append to the list and not touch any existing values.

I've simplified the code a bit and added some test examples.

docs/api.md
```diff
@@ -1227,6 +1227,20 @@ ServiceOverrides
 <p>Tolerations allow you to schedule pods onto nodes with matching taints.</p>
 </td>
 </tr>
+<tr>
+<td>
+<code>topologySpreadConstraints</code></br>
```
frouioui, line 1232: The way we generate documentation changed in https://github.com/planetscale/vitess-operator/pull/625. We no longer use `</br>` as it can break in Markdown. Do you mind merging `main` onto your branch and re-generating the docs?

pkg/operator/contenthash/topology_spread_constraints.go
```diff
@@ -42,7 +43,16 @@ func TopologySpreadConstraints(in []corev1.TopologySpreadConstraint) string {
...
+		writeStringHash(h, strings.Join(tsc.MatchLabelKeys, ""))
+		if tsc.MinDomains != nil {
+			writeStringHash(h, string(*tsc.MinDomains))
+		}
+		if tsc.NodeAffinityPolicy != nil {
+			writeStringHash(h, string(*tsc.NodeAffinityPolicy))
+		}
+		if tsc.NodeTaintsPolicy != nil {
+			writeStringHash(h, string(*tsc.NodeTaintsPolicy))
+		}
```
frouioui, line 55: That is a good catch. I am guessing the hash we were previously computing were not encapsulating all the information we could add?

pkg/operator/contenthash/topology_spread_constraints_test.go
```diff
@@ -0,0 +1,64 @@
...
+	nilHash := TopologySpreadConstraints([]corev1.TopologySpreadConstraint{
+		{MinDomains: nil},
+	})
+	zeroHash := TopologySpreadConstraints([]corev1.TopologySpreadConstraint{
+
+		{MinDomains: ptr.To(int32(0))},
+	})
+	if nilHash == zeroHash {
+		t.Errorf("nilHash = zeroHash = %v; expected different values", nilHash)
+	}
```
frouioui, line 37: Instead of checking the equality manually we should use `require.NotEqual`. It is something we have adopted a while ago on `vitessio/vitess`. We should start doing the same for new tests on the operator. Same applies to `TestTopologySpreadConstraintNodeAffinityPolicysNil` and `TestTopologySpreadConstraintNodeTaintsPolicysNil`.

pkg/operator/contenthash/topology_spread_constraints_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func TestTopologySpreadConstraintNodeAffinityPolicysNil(t *testing.T) {
+	// Make sure nil is distinguishable from 0.
```
frouioui, line 41: The comment does not entirely make sense here, same for the following test

pkg/operator/update/podspec.go
frouioui: I think it's reasonable to do a DeepEqual here as it will catch any potential change to the Topology Spread Constraints.

However, I am a bit afraid of the performance implication that `reflect.DeepEqual` brings. Since we are calling it in a nested loop, if the user defines many constraints on their vttablet or etcd pods we might be left with a longer update. To avoid this potential issue, could we create a method to compare two constraints?

frouioui (commented):
Besides the comments I have left it looks good to me. Do you have a manual E2E test for this, if so, could you share the script and output with us?

frouioui:
Hi @relu! Any update on this PR? It has been left without update for more than a month.

mattlord:
Closing this as stale

PR #590, opened 2024-07-25
Replace deprecated package k8s.io/utils/pointer

relu:
Using k8s.io/utils/ptr as advised instead.


frouioui (approved):
Thank you!

mattlord (approved):
Thanks, @relu !

PR #597, opened 2024-08-28
Bump Kubernetes version to 1.31.0 and add CI matrix

frouioui:
This PR bumps the supported version of Kubernetes to `v1.31.0`. To do that, all the Kubernetes dependencies were updated to `v1.31.0`, and the changes made to the K8S API were reported back into our code.

I also added a new matrix of test in BuildKite to ensure vitess-operator works with the three latest version of Kubernetes. I did not find a way to reuse the string constants that define the K8S versions we want to test, so for now they look a bit "re-used". Only solution I could find was to generate the pipeline YAML within the runtime of the pipeline, which did not feel scalable and clean to read.

Moreover, the versions are hardcoded right now, but I hope in the future to add some automation to automatically bump those constants, kind of like we have in Vitess with the Golang version.


README.md
```diff
@@ -19,13 +19,17 @@ compatible with certain Vitess and Kubernetes versions, as shown in this table:
...
 | `v2.13.*`               | `v20.0.*`                   | `v1.25.*`, `v1.26.*`, `v1.27.*`, or `v1.28.*` |
-| `latest`                | `latest`                    | `v1.25.*`, `v1.26.*`, `v1.27.*`, or `v1.28.*` |
+| `latest`                | `latest`                    | `v1.29.*`, `v1.30.*`, `v1.31.*`               |
 
 If for some reason you must attempt to use versions outside the recommend
 window, we still welcome bug reports since a workaround might be possible.
 However, in some cases we may be unable to overcome the underlying limitations
 in our dependencies.
 
+Each major release of the vitess-operator will support the three latest major release of Kubernetes at the time
```
mattlord, line 29: Releases plural

README.md
```diff
@@ -19,13 +19,17 @@ compatible with certain Vitess and Kubernetes versions, as shown in this table:
...
+of the vitess-operator's major release RC-1. This will ensure that the major release of the vitess-operator will
+always support at least one major release of Kubernetes for its whole lifetime (1 year).
```
mattlord, line 31: Would be good to link to the EOL policy so it’s clear what this 1 year listing is about.

README.md
frouioui: Fixed via 4ccd230

README.md
frouioui: fixed via 4ccd230

README.md
```diff
@@ -19,13 +19,18 @@ compatible with certain Vitess and Kubernetes versions, as shown in this table:
 | `v2.11.*`               | `v18.0.*`                   | `v1.22.*`, `v1.23.*`, `v1.24.*`, or `v1.25.*` |
 | `v2.12.*`               | `v19.0.*`                   | `v1.25.*`, `v1.26.*`, `v1.27.*`, or `v1.28.*` |
 | `v2.13.*`               | `v20.0.*`                   | `v1.25.*`, `v1.26.*`, `v1.27.*`, or `v1.28.*` |
-| `latest`                | `latest`                    | `v1.25.*`, `v1.26.*`, `v1.27.*`, or `v1.28.*` |
+| `latest`                | `latest`                    | `v1.29.*`, `v1.30.*`, `v1.31.*`               |
```
GrahamCampbell, line 22: Don't we need an overlapping Kubernetes version with the previous release, to make upgrading easier/less risky?

GrahamCampbell: That is, to support v1.28 through v1.31.

frouioui: The main branch does support 1.28 technically

GuptaManan100 (approved):

Issue #608, opened 2024-10-08
user-defined per-keyspace images feature is unusable


bluecrabs007:
The [issue](https://github.com/planetscale/vitess-operator/issues/446) and the [PR](https://github.com/planetscale/vitess-operator/pull/524) brings in the ability to create a new keyspace and override the mysqld container image.

This lets the user deploy a new keyspace with mysql8 in a vitesscluster deployed with mysql57 with the goal of letting the user `MoveTables` the data from the 57 keyspace to 80.

There are 2 issues with the current implementation and the version this feature was released in.

Issue 1:
This feature lets the user override the mysqld image only. This causes incompatibilities between the vttablet container and the mysqld container due to the version mismatch.
For example, the vttablet would be running the mysql57 version of xtrabackup which doesn't work with mysql80
it fails with error 
```
xtrabackupengine.go:369] xtrabackup stderr: Error: MySQL 8.0 and Percona Server 8.0 are not supported by Percona Xtrabackup 2.4.x series. Please use Percona Xtrabackup 8.0.x for backups and restores.
```
Suggestion: A bug fix to enable support for user-definable per-keyspace images for vttablet, vtorc, myqsld exporter.

Issue 2:
This feature landed in vitess-operator [v2.12.0](https://github.com/planetscale/vitess-operator/releases/tag/v2.12.0) which corresponds to vitess v19, which drops support for mysql57.
Suggestion: Backport this feature to v2.11.x so that users can migrate off mysql 57 from the last vitess version that supports mysql57

frouioui:
Hey @bluecrabs007, thanks for opening this. This is a reasonable thing to do in vitess-operator.

I have [opened a backport](https://github.com/planetscale/vitess-operator/pull/609) to `release-2.11` for the original fix made in `v2.12`. This should solve `Issue 2`.

While it gets merged, I am working on a fix for `Issue 1`.

PR #524, opened 2024-01-23
Add support for user-defined per-keyspace images

maxenglander:
## Description

Currently the way to upgrade from mysql 57 to 80 in the operator is to change `mysqld` image version. This causes operator to do a rolling upgrade of tablets. There's a lot that could go wrong here, and it would be nice if there were an option to do a controlled migration from 57 to 80.

One way this could work in theory is to create a new keyspace configured with mysql 80, and do a `MoveTables` from the 57 keyspace to the 80. This would allow carefully testing and validating that the upgrade succeeded before doing `SwitchTraffic`, and in theory would allow switching back to 57 if issues were discovered in 80 after `SwitchTraffic`.

This PR allows the user to specify per-keyspace images, relaxing this prohibition: https://github.com/planetscale/vitess-operator/blob/5bf4279c968a365395e2ae8cf6c4cf61c172ad69/docs/api.md#L5280-L5283

## Issues

Fixes: https://github.com/planetscale/vitess-operator/issues/446

## Validation

I validated the change by deploying a local build to a local `kind` cluster, and deploying a `VitessCluster` with one keyspace inheriting images from the cluster spec, and the other using user-defined overrides.

_kind.yaml_

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: planetscale-vitess-operator-dev
networking:
  ipFamily: ipv4  # macos only supported v4
nodes:
- role: control-plane
  image: kindest/node:v1.24.15@sha256:7db4f8bea3e14b82d12e044e25e34bd53754b7f2b0e9d56df21774e6f66a70ab
  labels:
    role: master
```

_vitesscluster.yaml_

```
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: dev-vitess-cluster
spec:
  images:
    vtadmin: vitess/lite:mysql57
    vtbackup: vitess/lite:mysql57
    vtctld: vitess/lite:mysql57
    vtgate: vitess/lite:mysql57
    vtorc: vitess/lite:mysql57
    vttablet: vitess/lite:mysql57
    mysqld:
      mysql56Compatible: vitess/lite:mysql57
  cells:
  - name: localhost
    gateway:
      authentication:
        static:
          secret:
            name: dev-vitess-cluster-config
            key: users.json
      secureTransport:
        tls:
          certSecret:
            name: dev-vitess-cluster-config
            key: cert.pem
          keySecret:
            name: dev-vitess-cluster-config
            key: key.pem
  keyspaces:
[...]

PR #609, opened 2024-10-08
[release-2.11] Add support for user-defined per-keyspace images

frouioui:
This is a backport of https://github.com/planetscale/vitess-operator/pull/524 into `release-2.11`.

The motivation for this backport stems from https://github.com/planetscale/vitess-operator/issues/608, as explained by @bluecrabs007, the goal of the original PR is to support two versions of MySQL in the cluster instead of using the same one everywhere, this is very useful as we were deprecating MySQL 5.7. However, we deprecated MySQL 5.7 in v18.0.0 (equivalent of v2.11 in vtop) and removed it entirely from the code base in v19.0.0 (equivalent v2.12 in vtop).

We want to backport this to `v2.11` so users can use the feature, and thus migrate more easily, on a version of Vitess where MySQL 5.7 is still supported.

This PR is part of https://github.com/planetscale/vitess-operator/issues/608. It fixes the `Issue 2` described.


maxenglander (approved):

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -244,9 +244,21 @@ type VitessImages struct {
...
+type MysqldImageNew struct {
+	// Mysql56Compatible is a container image (including version tag) for mysqld
+	// that's compatible with the Vitess "MySQL56" flavor setting.
+	Mysql56Compatible string `json:"mysql56Compatible,omitempty"`
+	// Mysql80Compatible is a container image (including version tag) for mysqld
+	// that's compatible with the Vitess "MySQL80" flavor setting.
+	Mysql80Compatible string `json:"mysql80Compatible,omitempty"`
+}
+
+// TODO: Remove this once everything is migrated to MysqldImageNew.
```
mattlord, line 261: This isn't right, is it? I thought that the migration was going to happen in the other direction and New will be removed?

mattlord (approved):

pkg/apis/planetscale/v2/vitesscluster_types.go
frouioui: We will be removing `MysqldImage` in `v2.15` and renaming `MysqldImageNew` to `MysqldImage` at the same time.

PR #612, opened 2024-10-09
Update `docs/api.md` and automatically update it

frouioui:
Each time we change the CRDs we do update part of our documentation in `./docs/api/index.html`, but this document is hardly readable for anyone and does not render well on the GitHub UI either. The better way of showing this documentation is by just changing the file extension to Markdown and GitHub will know how to render it. We do have the Markdown version of `./docs/api/index.html`, which is located in `./docs/api.md`. However, we have not updated this file in several years despite the many changes to vitess-operator.

This is an issue as `./docs/api.md` is supposed to be the only very-up-to-date documentation we have for vitess-operator.

This PR fixes this issue by adding a phase to `make generate` that will copy the newly generated `./docs/api/index.html` into `./docs/api.md` and remove the first two line of the file (which contain HTML headers that do not render well in Markdown).

Adding this as part of `make generate` will ensure that the documentation will always be consistent as we have a CI checks that makes sure `make generate` has been run on any new commit.


GuptaManan100 (approved):
LGTM!

deepthi (approved):

PR #447, opened 2023-07-20
Upgrade etcd version from 3.3.13 to 3.5.9

MrFabio:
This PR aims to upgrade the etcd version.

PR #654, opened 2025-01-17
Upgrade default etcd version to 3.5.17

GrahamCampbell:
This PR upgrades the default etcd version to 3.5.17.

Corresponding PR on vitess: https://github.com/vitessio/vitess/pull/17653

### DIFF
--- a/docs/release-notes/2_15_0_summary.md
+++ b/docs/release-notes/2_15_0_summary.md
@@ -1,0 +1,20 @@
+## Major Changes
+
+### etcd Upgrade Path
+
+We have changed the default etcd version to `3.5.17`.
+
+You can upgrade by changing your YAML file to use the new Docker Image (`quay.io/coreos/etcd:v3.5.17`).
+
+### MySQL Upgrade Path
+
+With the latest version of Vitess (`v22.0.0`) the default MySQL version changed from `8.0.30` to `8.0.40`.
+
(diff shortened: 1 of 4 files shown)


GuptaManan100:
The DCO check seems to be failing 😕. Did you add the `-s` flag while committing?

GuptaManan100 (approved):

GrahamCampbell:
Made the commit from the GitHub webapp. 🤷

frouioui (commented):
The etcd version should be changed everywhere both in vitess and vtop.

frouioui:
Have you tested upgrade paths with a vitess cluster on K8S using the previous etcd version, upgrading to the other etcd version? Same thing with downgrades.

GrahamCampbell:
Re-pushed with a signed off commit.

frouioui (dismissed):
LGTM. Can you modify the release notes summary for the next release (`2.15.0`) to include a note on how to proceed with this newer etcd version, https://github.com/planetscale/vitess-operator/pull/447#issuecomment-1646150438.

frouioui (commented):
Let's open a PR on vitess to make the etcd version match too please, we use `v3.3.10` in the tests. We also use the `ETCD_VER` env variable in several places, in the vitess repo, which should be set to the same version as here.

frouioui (approved):

PR #553, opened 2024-05-13
Add the ability to automate and schedule backups

frouioui:
## Description

This Pull Request adds a new CRD called `VitessBackupSchedule`. Its main goal is to automate and schedule backups of Vitess, taking backups of the Vitess cluster at regular intervals based on a given cron `schedule` and `Strategy`. This new CRD is managed by the `VitessCluster`, like most other components of the vitess-operator, the `VitessCluster` controller is responsible for the whole lifecycle (creation, update, deletion) of the `VitessBackupSchedule` object in the cluster. Inside the `VitessCluster` it is possible to define several `VitessBackupSchedule`s as a list, allowing for multiple concurrent backup schedules.

Among other things, the `VitessBackupSchedule` object is responsible for creating Kubernetes's Job at the desired time, based on the user-defined `schedule`. It also keeps track of older jobs and delete them if they are too old, according to user-defined parameters (`successfulJobsHistoryLimit` & `failedJobsHistoryLimit`). The jobs created by the `VitessBackupSchedule` object will use the `vtctld` Docker Image and will execute a shell command that is generated based on the user-defined `strategies`. The end user can define as many backup strategy per schedule, each of them mocks what `vtctldclient` is able to do, the `Backup` and `BackupShard` commands are available, a map of extra flags enable the user to give as many flag as they want to `vtctldclient`.

A new end-to-end test is added to our BuildKite pipeline as part of this Pull Request to test the proper behavior of this new CRD.

## Related PRs

- Documentation: https://github.com/vitessio/website/pull/1746
- Vitessio/vitess: https://github.com/vitessio/vitess/pull/15969

## Demonstration

For this demonstration I have setup a Vitess cluster by following the steps in the [getting started guide](https://vitess.io/docs/20.0/get-started/operator/), until the very last step where we must apply the `306_down_shard_0.yaml` file. My cluster is then composed of 2 keyspaces: `customer` with 2 shards, and `commerce` unsharded. I then modify the `306...` yaml file to contain the new backup schedule, as seen in the snippet right below. We want to create two schedules, one for each keyspace. The keyspace `customer` will have two backup strategies: one for each shard.

```yaml
apiVersion: planetscale.com/v2
kind: VitessCluster
metadata:
  name: example
spec:
  backup:
    engine: xtrabackup
    locations:
      - volume:
          hostPath:
            path: /backup
            type: Directory
    schedules:
      - name: "every-minute-customer"
        schedule: "* * * * *"
        resources:
          requests:
            cpu: 100m
[...]

PR #658, opened 2025-01-28
Use vtbackup in scheduled backups

frouioui:
## Context

This Pull Request modifies the scheduled backup feature that was introduced in https://github.com/planetscale/vitess-operator/pull/553. The implementation in the original PR is rather simple and was a "v1". The CRD `VitessBackupSchedule` was introduced, giving the option to end-users to configure different `VitessBackupScheduleStrategy`, which instruct vitess-operator which keyspace/shard the user wants to backup. All the strategies in a `VitessBackupSchedule` were merged together to form a single `vtctldclient` command that would be executed in a K8S's `Job` following the schedule set in the `VitessBackupSchedule` CRD.

While this approach works, it has limitation: it uses `vtctldclient Backup` which will temporarily take out a serving tablet from the serving tablet pool and use it to take the backup. While this limitation is documented in the release notes, we want to encourage people to use `vtbackup` as an alternative.

## Proposed Change

This Pull Request modifies the `VitessBackupSchedule` controller to create one Kubernetes `Job` per `VitessBackupScheduleStrategy` defined the `VitessBackupSchedule` by the user - instead of merging all strategies into a single `Job`. Each `Job` will now create a `vtbackup` pod - instead of creating a pod where a `vtctldclient` command was executed.

### Changes

- Since `vtbackup` can only backup a single keyspace/shard, we need one `Job` per strategy, meaning that now each `VitessBackupSchedule` manages multiple `Job` resources (one per strategy), which was not the case before.

- The previous code logic relied heavily on the CRD's Status to schedule next runs, this logic was removed to make Status purely informative to only the end users, not the vitess-operator. Moreover, the previous logic was updating the Status of the resource on, almost, every request without knowing if we actually need to update the Status. This was modified to only update the Status when needed and make it ignore conflicts.

- The reconciling loop of `VitessBackupSchedule` was, at times, not returning the correct result+error expected by the controller-runtime `TypedReconciler` interface, leading to unnecessary re-queue of requests or missed re-queues. The logic has been modified to return correct values. A periodic resync of the controller has also been added to safeguard potential issues and to ensure resources (`Jobs`, PVCs, etc) are cleaned up correctly no matter what.

- The upgrade test has been modified to also run the `401_scheduled_backups.yaml` manifest.

- The default `ConcurrentPolicy` of the `Jobs` has been modified from `Allow` to `Forbid`. With the new strategies requiring far more resources and time to complete, this new default makes sense. This value is still configurable by end-users.

- Previously, the `vtbackup` created with every `VitessShard` was always attempting to do an initial backup (`--initial_backup`), which would fail if the `VitessBackup` definition was added after the `VitessShard` started running. That logic has been slightly modified so the vitess-operator is more aware of the current state of the `VitessShard` before creating the `vtbackup` pod. If there is already a primary in the shard, it will not be an initial backup, but an `--allow_first_backup` backup - as we only create this pod if there is no backup for this shard. The `VitessBackupSchedule` relies on the same logic to determine the type of backup it has to run, either: initial, allow first, or normal. Concurrency issues between the initial `vtbackup` pod and the first scheduled run of a `VitessBackupSchedule` is, depending on the schedule, likely. In such situation both `vtbackup` pods come up with the `--initial_backup` flag set to true. One of the two pods will simply fail early.

## Related PRs

- Documentation: https://github.com/vitessio/website/pull/1946
- Vitess examples: https://github.com/vitessio/vitess/pull/17869

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2591,7 +2591,7 @@ The PullPolicy used will be the same as the one used to pull the vtctld image.</
 <td>
 <em>(Optional)</em>
 <p>A list of pointers to currently running jobs.
-This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.</p>
+This field is deprecated and no longer used in versions &gt;= v2.15. It will be removed in a future release.</p>
 </td>
 </tr>
 <tr>
@@ -2606,7 +2606,7 @@ Kubernetes meta/v1.Time
 <td>
 <em>(Optional)</em>
 <p>Information when was the last time the job was successfully scheduled.
-This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.
+This field is deprecated and no longer used in versions &gt;= v2.15. It will be removed in a future release.
 Please use lastScheduledTimes instead which maps the last schedule time to each VitessBackupScheduleStrategy
 in the VitessBackupSchedule.</p>
 </td>
@@ -2622,8 +2622,8 @@ map[string]*k8s.io/apimachinery/pkg/apis/meta/v1.Time
 </td>
 <td>
 <em>(Optional)</em>
-<p>LastScheduledTimes lists for every VitessBackupScheduleStrategy what is the last schedule we executed.
-It does not list when the last execution started, but to which scheduled time it corresponds.</p>
+<p>A list of the last schedule we executed for each VitessBackupScheduleStrategy.
+Note that these are not the times when the last execution started, only the scheduled times.</p>
 </td>
 </tr>
 </tbody>
@@ -2749,7 +2749,7 @@ string
 <p>Strategy defines how we are going to take a backup.
 If you want to take several backups within the same schedule you can add more items
 to the Strategy list. Each VitessBackupScheduleStrategy will be executed within different
-K8S Jobs. This is useful if you want to have a single schedule backing up multiple shards
+kubernetes jobs. This is useful if you want to have a single schedule backing up multiple shards
 at the same time.</p>
 </td>
 </tr>
--- a/docs/api.md
+++ b/docs/api.md
@@ -647,16 +647,6 @@ SecretSource
 </tr>
 </tbody>
 </table>
-<h3 id="planetscale.com/v2.BackupStrategyName">BackupStrategyName
-(<code>string</code> alias)</p></h3>
-<p>
-(<em>Appears on:</em>
-<a href="#planetscale.com/v2.VitessBackupScheduleStrategy">VitessBackupScheduleStrategy</a>)
-</p>
-<p>
-<p>BackupStrategyName describes the vtctldclient command that will be used to take a backup.
-When scheduling a backup, you must specify at least one strategy.</p>
-</p>
 <h3 id="planetscale.com/v2.CephBackupLocation">CephBackupLocation
 </h3>
 <p>
@@ -2600,7 +2590,8 @@ The PullPolicy used will be the same as the one used to pull the vtctld image.</
 </td>
 <td>
 <em>(Optional)</em>
-<p>A list of pointers to currently running jobs.</p>
+<p>A list of pointers to currently running jobs.
+This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.</p>
 </td>
 </tr>
 <tr>
@@ -2614,7 +2605,25 @@ Kubernetes meta/v1.Time
 </td>
 <td>
 <em>(Optional)</em>
-<p>Information when was the last time the job was successfully scheduled.</p>
+<p>Information when was the last time the job was successfully scheduled.
+This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.
+Please use lastScheduledTimes instead which maps the last schedule time to each VitessBackupScheduleStrategy
+in the VitessBackupSchedule.</p>
+</td>
+</tr>
+<tr>
+<td>
+<code>lastScheduledTimes</code><br>
+<em>
+<a href="https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#*k8s.io/apimachinery/pkg/apis/meta/v1.time--">
+map[string]*k8s.io/apimachinery/pkg/apis/meta/v1.Time
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>LastScheduledTimes lists for every VitessBackupScheduleStrategy what is the last schedule we executed.
+It does not list when the last execution started, but to which scheduled time it corresponds.</p>
 </td>
 </tr>
 </tbody>
@@ -2642,9 +2651,7 @@ command line that will be executed in the Job&rsquo;s pod.</p>
 <td>
 <code>name</code><br>
 <em>
-<a href="#planetscale.com/v2.BackupStrategyName">
-BackupStrategyName
-</a>
+string
 </em>
 </td>
 <td>
@@ -2741,11 +2748,9 @@ string
 <td>
 <p>Strategy defines how we are going to take a backup.
 If you want to take several backups within the same schedule you can add more items
-to the Strategy list. Each VitessBackupScheduleStrategy will be executed by the same
-kubernetes job. This is useful if for instance you have one schedule, and you want to
-take a backup of all shards in a keyspace and don&rsquo;t want to re-create a second schedule.
-All the VitessBackupScheduleStrategy are concatenated into a single shell command that
-is executed when the Job&rsquo;s container starts.</p>
+to the Strategy list. Each VitessBackupScheduleStrategy will be executed within different
+K8S Jobs. This is useful if you want to have a single schedule backing up multiple shards
+at the same time.</p>
 </td>
 </tr>
 <tr>
--- a/docs/api.md
+++ b/docs/api.md
@@ -2828,8 +2828,8 @@ ConcurrencyPolicy
 <em>(Optional)</em>
 <p>ConcurrencyPolicy specifies ho to treat concurrent executions of a Job.
 Valid values are:
-- &ldquo;Allow&rdquo; (default): allows CronJobs to run concurrently;
-- &ldquo;Forbid&rdquo;: forbids concurrent runs, skipping next run if previous run hasn&rsquo;t finished yet;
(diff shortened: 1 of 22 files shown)


pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
```diff
@@ -466,76 +549,83 @@ func (r *ReconcileVitessBackupsSchedule) createJob(ctx context.Context, vbsc *pl
...
+	pvc := &corev1.PersistentVolumeClaim{}
+	key := client.ObjectKey{
+		Namespace: job.Namespace,
+		Name:      name,
+	}
+	err = r.client.Get(ctx, key, pvc)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, err
+		}
```
mattlord, line 571: This could just be:
```
		if err != nil && !apierrors.IsNotFound(err) {
			return nil, err
		}
```

Then the rest of the code would be more clear and easier to read IMO. If we don't get an error, meaning that it already exists, we just re-use it? Do we ensure it's in a certain state in this case?

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
```diff
@@ -466,76 +549,83 @@ func (r *ReconcileVitessBackupsSchedule) createJob(ctx context.Context, vbsc *pl
...
+	if err != nil {
+		return nil, nil, err
 	}
 
-	for i, strategy := range vbsc.Spec.Strategy {
-		vtctldclientServerArg, err := getVtctldServiceName(vbsc.Spec.Cluster)
-		if err != nil {
-			return corev1.PodSpec{}, err
-		}
+	_, completeBackups, err := vitessbackup.GetBackups(ctx, vbsc.Namespace, vbsc.Spec.Cluster, strategy.Keyspace, vkr.SafeName(),
```
mattlord, line 598: Nit, but I think the var name should be `completedBackups`, no?

mattlord (approved):
LGTM! ❤️ I only had a few minor comments/questions.

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
```diff
@@ -46,19 +53,18 @@ import (
...
 
 const (
-	controllerName   = "vitessbackupschedule-controller"
-	vtctldclientPath = "/vt/bin/vtctldclient"
+	controllerName = "vitessbackupschedule-controller"
 )
 
 var (
 	maxConcurrentReconciles = flag.Int("vitessbackupschedule_concurrent_reconciles", 10, "the maximum number of different vitessbackupschedule resources to reconcile concurrently")
+	resyncPeriod            = flag.Duration("vitessbackupschedule_resync_period", 1*time.Minute, "reconcile vitessbackupschedules with this period even if no Kubernetes events occur")
```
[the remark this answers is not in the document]
frouioui: We do not have many flags in vitess-operator and all of them use underscores. It is fine to keep it like that.

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
frouioui: good point, cases where the PVC already exist (thus no error) should be extremely rare as the name of the resource is salted and prefixed by the schedule UNIX time. but i have added an `else` condition to error in such situation: 072d109

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
frouioui: fixed via 072d109

docs/api.md
```diff
@@ -2600,7 +2590,8 @@ The PullPolicy used will be the same as the one used to pull the vtctld image.</
 </td>
 <td>
 <em>(Optional)</em>
-<p>A list of pointers to currently running jobs.</p>
+<p>A list of pointers to currently running jobs.
+This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.</p>
```
deepthi, line 2594:
```suggestion
This field is deprecated and no longer used in versions&gt;= v2.15. It will be removed in a future release.</p>
```

docs/api.md
```diff
@@ -2614,7 +2605,25 @@ Kubernetes meta/v1.Time
...
+<code>lastScheduledTimes</code><br>
+<em>
+<a href="https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#*k8s.io/apimachinery/pkg/apis/meta/v1.time--">
+map[string]*k8s.io/apimachinery/pkg/apis/meta/v1.Time
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>LastScheduledTimes lists for every VitessBackupScheduleStrategy what is the last schedule we executed.
```
deepthi, line 2625:
```suggestion
<p>A list of the last schedule we executed for each VitessBackupScheduleStrategy. 
```

docs/api.md
```diff
@@ -2614,7 +2605,25 @@ Kubernetes meta/v1.Time
...
+It does not list when the last execution started, but to which scheduled time it corresponds.</p>
```
deepthi, line 2626:
```suggestion
Note that these are not the times when the last execution started, only the scheduled times.</p>
```

docs/api.md
```diff
@@ -2741,11 +2748,9 @@ string
 <td>
 <p>Strategy defines how we are going to take a backup.
 If you want to take several backups within the same schedule you can add more items
-to the Strategy list. Each VitessBackupScheduleStrategy will be executed by the same
-kubernetes job. This is useful if for instance you have one schedule, and you want to
-take a backup of all shards in a keyspace and don&rsquo;t want to re-create a second schedule.
-All the VitessBackupScheduleStrategy are concatenated into a single shell command that
-is executed when the Job&rsquo;s container starts.</p>
+to the Strategy list. Each VitessBackupScheduleStrategy will be executed within different
+K8S Jobs. This is useful if you want to have a single schedule backing up multiple shards
```
deepthi, line 2752:
```suggestion
kubernetes jobs. This is useful if you want to have a single schedule backing up multiple shards
```

docs/api.md
```diff
@@ -2828,8 +2833,8 @@ ConcurrencyPolicy
 <em>(Optional)</em>
 <p>ConcurrencyPolicy specifies ho to treat concurrent executions of a Job.
 Valid values are:
-- &ldquo;Allow&rdquo; (default): allows CronJobs to run concurrently;
-- &ldquo;Forbid&rdquo;: forbids concurrent runs, skipping next run if previous run hasn&rsquo;t finished yet;
+- &ldquo;Allow&rdquo;: allows CronJobs to run concurrently;
+- &ldquo;Forbid&rdquo; (default): forbids concurrent runs, skipping next run if previous run hasn&rsquo;t finished yet;
```
deepthi, line 2837: 👍

docs/api/index.html
```diff
@@ -2602,7 +2592,8 @@ <h3 id="planetscale.com/v2.VitessBackupScheduleStatus">VitessBackupScheduleStatu
 </td>
 <td>
 <em>(Optional)</em>
-<p>A list of pointers to currently running jobs.</p>
+<p>A list of pointers to currently running jobs.
+This field is deprecated and no longer used in &gt;= v2.15. It will be removed in a future release.</p>
```
deepthi, line 2596: Is this file generated? I'm not repeating my comments from api.md here.

deepthi (commented):
I only had a few comments on the diff in this PR, but it's unclear to me how AllowFirstBackup works. Is that functionality that was already present, and hence not showing up as a diff in this PR? If I'm just missing it, please point me to the file where I can see it.

frouioui:
@deepthi `AllowFirstBackup` is new code, the flag is set here https://github.com/planetscale/vitess-operator/blob/072d109cd9bf4aa0f017278738c76adcba57ade5/pkg/operator/vttablet/flags.go#L107 on the `vtbackup` pod. The decision whether to set the flag or not is done in `vtbackupSpec`.

frouioui:
We have this constant:
https://github.com/planetscale/vitess-operator/blob/e9e15940d5b94180ea48b721f88ccdd49a91854d/pkg/operator/vitessbackup/labels.go#L28

That is used in both places, once for the vtbackup init pod, and another for the scheduled backup:

https://github.com/planetscale/vitess-operator/blob/072d109cd9bf4aa0f017278738c76adcba57ade5/pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go#L611

https://github.com/planetscale/vitess-operator/blob/28875f9c973d4534ef696ef647a198d4ec032408/pkg/controller/vitessshard/reconcile_backup_job.go#L87

docs/api/index.html
frouioui: `docs/api.md` and `docs/api/index.html` are both generated using the same source.

deepthi:
> We have this constant:
> 
> https://github.com/planetscale/vitess-operator/blob/e9e15940d5b94180ea48b721f88ccdd49a91854d/pkg/operator/vitessbackup/labels.go#L28
> 
> That is used in both places, once for the vtbackup init pod, and another for the scheduled backup:
> 
> https://github.com/planetscale/vitess-operator/blob/072d109cd9bf4aa0f017278738c76adcba57ade5/pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go#L611
> 
> https://github.com/planetscale/vitess-operator/blob/28875f9c973d4534ef696ef647a198d4ec032408/pkg/controller/vitessshard/reconcile_backup_job.go#L87
> 
> This flag will be used if: we have no backup and we already have a primary running in the shard.

Based on an out-of-band discussion, we should not allow people to take "first" backups on a long-running cluster. If they have not been using VItessBackupSchedules and they want to start using them, there must be a usable backup that can be used as the basis for the first scheduled backup.

docs/api.md
frouioui: fixed via 423e959

docs/api.md
frouioui: fixed via 423e959

docs/api.md
frouioui: fixed via 423e959

docs/api.md
frouioui: fixed via 423e959

deepthi (approved):

PR #662, opened 2025-02-11
pass on the sidecarDbName when creating a keyspace

### LINKED ISSUE #660: [Feature] Allow setting custom sidecar-db-name in VitessCluster spec
## Description
Allow setting `--sidecar-db-name` for vttablet pods created by the Vitess operator. By enabling this feature, users can define different sidecar databases for a single unmanaged MySQL instance to support multiple unmanaged keyspaces effectively. 

This will prevent conflicts when multiple external keyspaces share the same MySQL instance, particularly for workflows such as `MoveTables` or reverse replication, which store metadata in the sidecar database.

## Use case
Organizations using Vitess may operate one single unmanaged MySQL instance with multiple databases that map to separate Vitess external keyspaces. In these scenarios, each external keyspace must use its own sidecar database to avoid conflicts when managing workflows or metadata.
This gives the unmanaged MySQL instance with multiple sidecar dbs

For examples
- `external_ks_db1` uses `db1` schema and sidecar database `sidecar_db1`.
- `external_ks_db2` uses `db2` schema and sidecar database `sidecar_db2`.

Without the ability to configure `--sidecar-db-name`, multiple keyspaces on the same MySQL instance will share the default `_vt` sidecar database, leading to potential conflicts during operations like reverse replication.

mcrauwel:
fixes #660

### DIFF
--- a/deploy/crds/planetscale.com_vitessclusters.yaml
+++ b/deploy/crds/planetscale.com_vitessclusters.yaml
@@ -2562,6 +2562,8 @@ spec:
                       maxItems: 2
                       minItems: 1
                       type: array
+                    sidecarDbName:
+                      type: string
                     turndownPolicy:
                       enum:
                       - RequireIdle
--- a/deploy/crds/planetscale.com_vitesskeyspaces.yaml
+++ b/deploy/crds/planetscale.com_vitesskeyspaces.yaml
@@ -1120,6 +1120,8 @@ spec:
                 maxItems: 2
                 minItems: 1
                 type: array
+              sidecarDbName:
+                type: string
               topologyReconciliation:
                 properties:
                   pruneCells:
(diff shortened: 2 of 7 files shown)


frouioui (commented):
LGTM. Only thing missing is to re-generate the `operator-latest.yaml` file which we use to run our E2E tests and examples. To do this, you can run the following command:
```
make generate && kustomize build ./deploy > build/_output/operator.yaml
```

The output (`build/_output/operator.yaml`) will have to be copied to `test/endtoend/operator/operator-latest.yaml`. Once copied, you will want to revert unwanted changes on lines: `7698` and `7702`.

Finally, the new changes brought to `operator-latest.yaml` will need to be ported over to the vitess OSS repository to the file: `examples/operator/operator.yaml`.

mattlord (approved):
Thanks, @mcrauwel ! ❤️

PR #665, opened 2025-02-25
Correct CreateKeyspace behavior to support custom sidecar DB names

mattlord:
This is a follow-up to https://github.com/planetscale/vitess-operator/pull/662.

We did not update any tests in that PR so we did not realize that while the code added there was correct, the new feature/support added there did not work due to a latent bug that had been lurking in the operator for a long time but gone unnoticed. That latent bug was here: https://github.com/planetscale/vitess-operator/blob/f31cb56d616b877145612cc9c064d25d8f43c187/pkg/controller/vitesskeyspace/reconcile_keyspace_information.go#L48-L59

It *should* have issued the `CreateKeyspace` RPC if the error returned *was* the topo `NoNode` error. But instead, it had that check negated. Due to this, that `CreateKeyspace` block was never executed and instead the keyspace was later created by the first tablet that started up within the keyspace. The other keyspace configuration options were then later applied when the next reconcile job ran. The SidecarDB name, however, is immutable in a Keyspace and thus could not be updated and was thus never applied.

In this PR we correct that logic and add a test to confirm that the support for the custom sidecar DB name works from end to end.

### DIFF
--- a/test/integration/vitesscluster/topology_test.go
+++ b/test/integration/vitesscluster/topology_test.go
@@ -113,13 +113,9 @@ func populateTopo(f *framework.Fixture, ts *topo.Server) {
 		f.Fatalf("Can't create keyspace: %v", err)
 	}
 
-	// To test shard and tablet pruning, we have to first create the actual
-	// desired keyspace and desired shards (which won't be pruned) because the
-	// operator doesn't do that. Normally the vttablets do that, but Vitess is
-	// not actually running here.
-	if err := ts.CreateKeyspace(f.Context(), "keyspace1", &topodatapb.Keyspace{}); err != nil {
-		f.Fatalf("Can't create keyspace: %v", err)
-	}
+	// To test shard and tablet pruning, we have to first create the desired
+	// shards (which won't be pruned) because the operator doesn't do that.
+	// Normally the vttablets do that, but Vitess is not actually running here.
 	if err := ts.CreateShard(f.Context(), "keyspace1", "-80"); err != nil {
 		f.Fatalf("Can't create shard: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


test/integration/vitesscluster/topology_test.go
```diff
@@ -113,13 +113,9 @@ func populateTopo(f *framework.Fixture, ts *topo.Server) {
 		f.Fatalf("Can't create keyspace: %v", err)
 	}
 
-	// To test shard and tablet pruning, we have to first create the actual
-	// desired keyspace and desired shards (which won't be pruned) because the
-	// operator doesn't do that. Normally the vttablets do that, but Vitess is
-	// not actually running here.
-	if err := ts.CreateKeyspace(f.Context(), "keyspace1", &topodatapb.Keyspace{}); err != nil {
-		f.Fatalf("Can't create keyspace: %v", err)
-	}
```
frouioui, line 122: Why did this get removed?

frouioui (approved):

test/integration/vitesscluster/topology_test.go
mattlord: Because it had encoded the bug in the test. If you notice the comment:
```
	// To test shard and tablet pruning, we have to first create the actual
	// desired keyspace and desired shards (which won't be pruned) because the
	// operator doesn't do that. Normally the vttablets do that, but Vitess is
	// not actually running here.
```

It was creating the keyspace *because the operator did not create it*. That was the bug. So when we fixed that bug, this test failed with a topo `NodeExists` error when trying to create the keyspace because the keyspace already existed. Once we removed this block which was added to work around the bug, then the test passed.

deepthi (approved):

PR #668, opened 2025-03-02
[vitess-operator][helm-chart] update

luismacosta:
Purpose:
- [vitess-operator][helm-chart] update

Changes:
- create templates folder + .helmignore +  _helpers.tpl + values.yaml + Chart.yaml + README.md
- update operator.yaml + service_account.yaml

Tests:
```yaml
cd deploy
helm template . --name-template vitess-operator -n vitess-operator -f values.yaml > test.yaml --debug
```

@frouioui @mattlord can you please review this PR? Thanks


frouioui (commented):
I don't fully understand the purpose of this Pull Request and what its use would be.

deepthi:
As @frouioui pointed out in slack, we do not currently provide a helm chart for deploying vitess-operator, nor do we intend to do so.
Beyond that, my main objection to this PR is that it replaces `operator.yaml` that works directly with vitess-operator with a template that requires people to use helm.
You are welcome to publish a helm chart template for vitess-operator in some other public repo.

PR #695, opened 2025-04-29
Update post GA 22

frouioui:
Make sure we use the right version


frouioui:
Closing, was updated after the release of v22.0.1

Issue #549, opened 2024-04-11
Add VTAdmin setting for fetch credentials to CRD


leejones:
We deploy VTAdmin using the vitess-operator and make it accessible using an authenticated proxy. In order for it to work, we need the frontend to send the user's HTTP Cookie containing their credentials when the frontend Javascript makes requests to the VTAdmin API. By default, this is currently [disabled](https://github.com/planetscale/vitess-operator/blob/98bb02ab77c4ff526b91b9dd28ea8111812e457f/pkg/controller/vitesscluster/reconcile_vtadmin.go#L345) by the operator. We'd like the option to change this setting in the CRD (e.g. change to `same-origin`).

For example, something like this would be great:

```yaml
kind: VitessCluster:
spec:
  vtadmin:
    fetchCredentials: same-origin
```

The default could remain `omit` for backwards compatibility and/or security.

Currently we workaround this missing setting by mounting a custom nginx config that serves a custom `config.js` instead of the one rendered by the operator.

References:

- Docs on related [Javascript Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included)

PR #711, opened 2025-07-17
Add fetchCredentials setting for VTAdmin

leejones:
Closes https://github.com/planetscale/vitess-operator/issues/549

This PR adds a setting for VTAdmin that will allow users to change whether or not vtadmin-web will include credentials (e.g. cookies, auth headers) when sending requests to vtadmin-api. It defaults to the original value (`omit`) for backwards-compatibility and should be a no-op for existing users.

Our use case for this new setting is that we have an auth proxy in front of VTAdmin that uses a cookie for authentication. With the current vitess-operator default (`omit`), this auth cookie is not included when vtadmin-web makes requests to vtadmin-api and these requests fail because the auth proxy cannot authenticate them. With this new setting, we'll be able to change the value to `same-origin` so that VTAdmin works when behind the auth proxy.


frouioui (approved):
LGTM, thank you!

rohit-nayak-ps (approved):

PR #717, opened 2025-08-27
Change the default MySQL version to 8.4.6

GrahamCampbell:


GrahamCampbell:
I guess we probably need to change `mysql80Compatible` to `mysql84Compatible` too?

PR #723, opened 2025-09-24
allow specifying the hex width to use when generating shard ranges

frouioui:
This ports https://github.com/vitessio/vitess/pull/18633 to the vitess-operator.

This PR adds the field `hexWidth` in the CRD, in `VitessKeyspaceEqualPartitioning`. With it, users can specify the width of their shard names. The default value is set by the kubebuilder to 0, which makes it a no-op and backward compatible for previous versions of Vitess or for people not using hex width.

### DIFF
--- a/deploy/crds/planetscale.com_vitessclusters.yaml
+++ b/deploy/crds/planetscale.com_vitessclusters.yaml
@@ -2583,7 +2583,6 @@ spec:
                                 - databaseInitScriptSecret
                                 type: object
                             required:
-                            - hexWidth
                             - parts
                             type: object
                         type: object
--- a/pkg/apis/planetscale/v2/vitesskeyspace_types.go
+++ b/pkg/apis/planetscale/v2/vitesskeyspace_types.go
@@ -362,7 +362,7 @@ type VitessKeyspaceEqualPartitioning struct {
 	// +kubebuilder:default=0
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=65536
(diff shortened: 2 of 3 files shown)


deploy/crds/planetscale.com_vitessclusters.yaml
```diff
@@ -2140,6 +2140,12 @@ spec:
                             type: object
                           equal:
                             properties:
+                              hexWidth:
+                                default: 0
+                                format: int32
+                                maximum: 65536
```
nickvanw, line 2146: Is there a reasonable maximum for this that isn't int32 max?

nickvanw (approved):
LGTM pending one question about data validation

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -357,6 +357,13 @@ type VitessKeyspaceEqualPartitioning struct {
 	// +kubebuilder:validation:Maximum=65536
 	Parts int32 `json:"parts"`
 
+	// HexWidth is the number of hex characters to use for the shard range start and end.
+	// If not set or set to 0, it will be automatically computed based on the number of requested shards.
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65536
+	HexWidth int32 `json:"hexWidth"`
```
fatih, line 365: Is this required? If not we should add an omitempty tag

PR #629, opened 2024-10-25
Support extra flags for mysqld_exporter

stankevich:
Adds a new `mysqldExporter.extraFlags` property that can be used to pass custom [flags](https://github.com/prometheus/mysqld_exporter#collector-flags) to mysqld_exporter to enable/configure additional collectors. Also, makes it possible to override the `collect.info_schema.tables.databases` value if needed. The default behaviour and collector flags remain unchanged.

This PR largely follows existing code conventions in the operator. I added a new `FormatArgsConvertBoolean()` method to Flags to support mysqld_exporter's [flag format](https://github.com/prometheus/mysqld_exporter#flag-format). Added the tests for it and for the existing `FormatArgs()` and `Merge()` methods.

Closes https://github.com/planetscale/vitess-operator/issues/470.

PR #735, opened 2025-11-26
Add extraLabels to both subcontroller and vbsc pods

### LINKED ISSUE #736: Add extraLabels option for backup pods
Neither the vitessbackupstorage-subcontroller or vbsc pods generated by the operator have the option to be passed metadata.labels.

The vtbackup pod likewise doesn't allow labels to be passed.

jdoupe:
Fixes https://github.com/planetscale/vitess-operator/issues/736

By adding the following:
vitessbackupschedules.spec.extraLabels
vitessbackupstorages.spec.extraLabels
vitessclusters.spec.backup.extraLabels

For vtbackup(-init), the code has just been modified to pass along the existing extraLabels on the tabletPool.

@mattlord / @frouioui


pkg/controller/vitesscluster/reconcile_backup_storage.go
```diff
@@ -81,15 +81,25 @@ func newVitessBackupStorage(key client.ObjectKey, parentLabels map[string]string
...
 	}
 
+	spec := planetscalev2.VitessBackupStorageSpec{
+		Location:      *location,
+		Subcontroller: subcontroller,
+	}
+
+	// Populate ExtraLabels from ClusterBackupSpec if available.
+	if extraLabels != nil {
+		spec.ExtraLabels = make(map[string]string)
```
mattlord: Nit, but we can pre-allocate the space:
```
spec.ExtraLabels = make(map[string]string, len(extraLabels))
```

pkg/operator/vitessbackup/schedule.go
```diff
@@ -32,22 +32,32 @@ func NewVitessBackupSchedule(key client.ObjectKey, vt *planetscalev2.VitessClust
...
+
+		// To take backups we only care about having the vtctldclient installed in the container.
+		// For this reason, we re-use the vtctld Docker image and the same image pull policy.
+		Image:           vt.Spec.Images.Vtctld,
+		ImagePullPolicy: vt.Spec.ImagePullPolicies.Vtctld,
+	}
+
+	// Populate ExtraLabels from ClusterBackupSpec if available.
+	if vt.Spec.Backup.ExtraLabels != nil {
+		spec.ExtraLabels = make(map[string]string)
```
mattlord: Same nit here about pre-allocation.

mattlord (approved):
LGTM! Just a couple of tiny nits. Thanks, @jdoupe !

frouioui (commented):
Hi! The operator manifest and the vitess repository will need to be modified given this PR updates the CRDs. A guide on how to do this is available here:

https://github.com/planetscale/vitess-operator/pull/629#discussion_r1828138492

pkg/controller/vitesscluster/reconcile_backup_storage.go
```diff
@@ -81,15 +81,25 @@ func newVitessBackupStorage(key client.ObjectKey, parentLabels map[string]string
...
+		Location:      *location,
+		Subcontroller: subcontroller,
+	}
+
+	// Populate ExtraLabels from ClusterBackupSpec if available.
+	if extraLabels != nil {
+		spec.ExtraLabels = make(map[string]string, len(extraLabels))
+		for k, v := range extraLabels {
+			spec.ExtraLabels[k] = v
+		}
```
mhamza15, line 94: Nit: can use `maps.Clone`

pkg/operator/vitessbackup/schedule.go
```diff
@@ -32,22 +32,32 @@ func NewVitessBackupSchedule(key client.ObjectKey, vt *planetscalev2.VitessClust
...
+		Image:           vt.Spec.Images.Vtctld,
+		ImagePullPolicy: vt.Spec.ImagePullPolicies.Vtctld,
+	}
+
+	// Populate ExtraLabels from ClusterBackupSpec if available.
+	if vt.Spec.Backup.ExtraLabels != nil {
+		spec.ExtraLabels = make(map[string]string, len(vt.Spec.Backup.ExtraLabels))
+		for k, v := range vt.Spec.Backup.ExtraLabels {
+			spec.ExtraLabels[k] = v
+		}
```
mhamza15, line 52: Same as above, can use `maps.Clone`

pkg/controller/vitessbackupstorage/reconcile_subcontroller.go
```diff
@@ -69,6 +69,10 @@ func (r *ReconcileVitessBackupStorage) reconcileSubcontroller(ctx context.Contex
 		planetscalev2.ClusterLabel:   clusterName,
 		vitessbackup.LocationLabel:   vbs.Spec.Location.Name,
 	}
+	// Merge user-provided extra labels from ClusterBackupSpec.
+	if vbs.Spec.ExtraLabels != nil {
+		update.Labels(&labels, vbs.Spec.ExtraLabels)
```
mhamza15, line 74: This adds the custom labels to the label set that `ReconcileObject` uses to decide whether the existing pod belongs to this controller. If a subcontroller pod already exists and the user adds `extraLabels` later, the pod will not have those new labels yet, so reconciliation fails with `NameCollision` before it gets a chance to update the pod. I believe we should keep `labels` to only the stable operator labels here, and add `vbs.Spec.ExtraLabels` directly to the pod in `New` and `UpdateInPlace` instead.

pkg/operator/vttablet/vtbackup_pod.go
```diff
@@ -193,6 +193,9 @@ func NewBackupPod(key client.ObjectKey, backupSpec *BackupSpec, mysqldImage stri
 		},
 	}
 
+	// Apply ExtraLabels to the pod, similar to regular vttablet pods.
+	update.Labels(&pod.Labels, tabletSpec.ExtraLabels)
```
mhamza15, line 197: Scheduled backup pods still will not get the custom tablet pool labels because the job only copies the generated pod spec, not the generated pod labels: https://github.com/jdoupe/vitess-operator/blob/757fb6e82d4eb3bbfbd5db7eeeda54faebe57154/pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go#L684-L686.

Those labels are lost before Kubernetes creates the real backup pod. I think we need to also add the custom tablet pool labels to the job template.

mhamza15 (commented):
Left a few comments, LGTM otherwise. I also wrote up some regression tests that should reproduce the cases the two comments are referring to: https://github.com/mhamza15/vitess-operator/commit/f685780d3c0c6915a1036ad01db454ef650155af

PR #738, opened 2025-12-19
deploy: Remove duplicate operator service account definition

mdnfiras:
This solve an issue described in [vitessio/vitess/issues/19054](https://github.com/vitessio/vitess/issues/19054) and solved in [vitessio/vitess/pull/19053](https://github.com/vitessio/vitess/pull/19053).

I'm reimplementing this here to sync things up and get a better signal as to whether or not there are any unexpected/unintended side effects.

cc @mattlord

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -1,24 +1,4 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
-metadata:
-  name: vitess-operator-backup-storage
-rules:
-- apiGroups:
-  - planetscale.com
-  resources:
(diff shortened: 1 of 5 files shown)


mdnfiras:
I'm not sure what you mean by `making sure no new indentations etc are added`: indentations are actually missing, and it looks like I made a lot of changes in the git diff if I don't add the missing indentations.

Also the deployment's `spec.securityContext` is missing in `kustomize build ./deploy` output, and there are few other differences unrelated to my changes. Should I add the missing indentations so we can see clearly what **actually** changed in the git diff?

PR #776, opened 2026-04-09
Add a 14 day dependency cooldown to the Dependabot config

Sinjo:
As a mitigation against supply-chain attacks, we're adopting a 14 day cooldown on dependency updates. This covers a lot of the supply chain attacks via compromised dependencies that have been seen in the wild so far, where the compromised version is only available for less than a day.

This isn't perfect, or the whole story, but it's a good next step for us to take.

## Related Issue(s)

https://github.com/planetscale/renovate-config/pull/91

## Security Impact

Reduce risk of supply-chain attacks.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 14
     groups:
       all:
         patterns:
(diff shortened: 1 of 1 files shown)


mattlord (approved):
Thank you, @Sinjo ! ❤️

mhamza15 (approved):

PR #781, opened 2026-04-16
Migrate e2e tests from Buildkite to GitHub Actions

nickvanw:
## Summary

Moves the eight end-to-end tests (upgrade, backup-restore, three backup-schedule variants, vtorc-vtadmin, unmanaged-tablet, hpa) off the Buildkite `public` queue and onto GitHub Actions. Each test runs as a matrix job on `ubuntu-latest-8-cores` (8 vCPU / 32 GB), all in parallel, with no per-test concurrency gate.

## Why

Buildkite's agent model put every job inside a `docker:latest` container talking to a shared host Docker daemon, which forced a lot of incidental complexity on us: a 15-line Alpine bootstrap in every step to install Go from a tarball, a sibling-container networking hack to make `kubectl` reach `kind`, a `pre-exit` hook that ran `sudo fix-buildkite-agent-builds-permissions` to undo chown damage from the operator container mutating the shared checkout, and a `concurrency: 1` gate per test to keep two jobs on the same agent from fighting over the hardcoded `vitess-operator-pr:latest` image tag and the fixed `15306`/`14001`/`15999` port-forward ports.

On GitHub-hosted runners every job gets a fresh VM with its own Docker daemon and its own localhost, so all four problems evaporate for free. The Alpine bootstrap becomes `actions/setup-go`, the sibling-container hack is deleted, the permission fix-up is unnecessary on ephemeral runners, and the concurrency gate is gone because the resources the tests share on Buildkite aren't shared on GHA.

## Changes

- **New** `.github/workflows/e2e-test.yaml` — matrix over the 8 test targets, `fail-fast: false`, 40-minute timeout per job, installs `mysql-client` + `chromium-browser` (the latter is only needed by `vtorc-vtadmin` but harmless to preinstall everywhere).
- **`test/endtoend/utils.sh`**
  - Renamed `BUILDKITE_JOB_ID` → `CI_JOB_ID` (still defaults to `0` so `make upgrade-test` on a laptop is unchanged). The workflow sets it to `${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.test.target }}`.
  - Deleted `setupKubectlAccessForCI` entirely — the Buildkite-specific `docker network connect kind` + `kind get kubeconfig --internal` dance is not needed when kind runs directly on the runner host.
  - The `docker build --progress plain` branch now keys off `$CI` (GHA sets this automatically).
- **Deleted** `.buildkite/pipeline.yml` and `.buildkite/hooks/pre-exit`.
- **`docs/release-process.md`** — updated one stale "buildkite" reference.

## Follow-ups (outside this PR)

1. Verify `ubuntu-latest-8-cores` resolves — if the org hasn't provisioned larger runners, the first run will fail fast and we can swap the label to `ubuntu-latest` (4 vCPU / 16 GB) or whatever larger label is available.
2. After the first green run, update branch protection on `main` and `release-**` to require the new `e2e` matrix check names and drop the Buildkite required checks.
3. Turn off the Buildkite pipeline in the Buildkite UI so it stops running on PRs.


mattlord (approved):
❤️

maxenglander (approved):

mhamza15:
Built on top of this: https://github.com/planetscale/vitess-operator/pull/805

PR #787, opened 2026-05-12
`vtorc`: pass `--cell` flag on Vitess v24+

### LINKED ISSUE #786: Feature Request: pass `--cell` flag to VTOrc deployments (required in Vitess v25+)
### Feature Description

Vitess v25 makes VTOrc's `--cell` flag a hard requirement. The change is in https://github.com/vitessio/vitess/pull/20048 — `validateCell()` previously logged a warning and continued; in v25 it returns a `FAILED_PRECONDITION` error if `--cell` is empty. The flag itself was introduced in v24 _(https://github.com/vitessio/vitess/pull/19047)_ to give VTOrc cell/location context for analysis and recovery decisions

The operator's VTOrc deployment does **not** currently pass `--cell`. In `pkg/operator/vtorc/deployment.go`, the `flags()` function _(lines 212-223)_ sets only:

```go
"topo_implementation":        spec.GlobalLockserver.Implementation,
"topo_global_server_address": spec.GlobalLockserver.Address,
"topo_global_root":           spec.GlobalLockserver.RootPath,
"port":                       planetscalev2.DefaultWebPort,
"clusters_to_watch":          spec.Keyspace + "/" + spec.Shard,
"logtostderr":                true,
```

Once a cluster upgrades to Vitess v25, every VTOrc managed by this operator will fail to start with a `FAILED_PRECONDITION` error from `validateCell()` 😱

#### What needs to happen

`Spec.Cell` is already on the vtorc spec _(`deployment.go:57`)_, and 2 x other components in the operator already pass it to their binaries:

- `pkg/operator/vtgate/deployment.go:269` — `"cell": spec.Cell.Name`
- `pkg/operator/vtctld/deployment.go:236` — `"cell": spec.Cell.Name`

Adding `"cell": spec.Cell` to vtorc's `flags()` map matches the existing pattern

#### Version-skew consideration

`--cell` was only added to VTOrc in v24, so passing it unconditionally affects users on v23 or older. Worth confirming whether VTOrc on ≤v23 rejects unknown flags or tolerates them before deciding if the flag needs version gating

#### Impact summary

1. **v25+** — VTOrc will not start without this fix
2. **v24** — `--cell` is accepted, missing flag emits a warning
3. **≤v23** — `--cell` does not exist; behaviour with the flag passed needs verification

### Use Case(s)

Every cluster managed by `vitess-operator` that upgrades to Vitess v25 _(GA per the v25.0.0 milestone)_. Without this change, the operator's VTOrc deployments break on upgrade

timvaillancourt:
## What's this?

Resolves https://github.com/planetscale/vitess-operator/issues/786 — VTOrc's `--cell` flag is required in Vitess v25+ and the operator doesn't currently set it. After upgrading to v25 every VTOrc managed by this operator would `FAILED_PRECONDITION` at startup 😱

## Why version-gating instead of always passing it?

`--cell` was introduced in Vitess v24 _([vitessio/vitess#19047](https://github.com/vitessio/vitess/pull/19047))_ and became required in v25 _([vitessio/vitess#20048](https://github.com/vitessio/vitess/pull/20048))_. On `≤v23` the flag doesn't exist — and I verified against `upstream/release-22.0` that `vtorc` uses a plain cobra `Main` with no `FParseErrWhitelist` or `DisableFlagParsing` _(neither in `go/cmd/vtorc/cli/cli.go` nor `go/vt/servenv/`)_, so passing `--cell` to ≤v23 vtorc would CrashLoopBackOff with `unknown flag: --cell`. We need to only emit the flag when we know the binary supports it.

## How it works

1. New helper `vitess.MajorVersionFromImage(image)` parses the major version from the image tag _(handles `vitess/lite:v24.0.0-mysql80`, RC suffixes, digest-only refs, rolling tags)_
2. `vtorc.Spec.flags()` calls the helper against `spec.Image` and adds `"cell": spec.Cell` only when `ok && major >= 24`
3. `ExtraFlags` continues to override base flags _(the existing loop at `deployment.go:125-130` runs after the base map)_, so users on rolling tags or who need a custom value can still force it

`spec.Cell` is populated from `tabletPool.Cell` at `pkg/controller/vitessshard/reconcile_vtorc.go:145` — there's already one vtorc Deployment per `(cluster, keyspace, shard, cell)` tuple, so the value is always the cell that this particular vtorc is responsible for

## What's covered

| Image tag                              | Result            |
|----------------------------------------|-------------------|
| `vitess/lite:v25.0.0-mysql80`          | emits `--cell`    |
| `vitess/lite:v24.0.0-rc1-mysql80`      | emits `--cell`    |
| `vitess/lite:v23.0.5-mysql80`          | no `--cell`       |
| `vitess/lite:mysql80` _(rolling)_      | no `--cell`       |
| `vitess/lite:latest`                   | no `--cell`       |
| `vitess/lite@sha256:...` _(digest)_    | no `--cell`       |
| Empty / unparseable                    | no `--cell`       |

## Related Issue(s)

Resolves: https://github.com/planetscale/vitess-operator/issues/786

## Test plan

- [x] `go test ./pkg/operator/vitess/... ./pkg/operator/vtorc/... -count=1` _(12 cases for the version helper, 9 for the vtorc flag gate)_
- [x] `go build ./...` clean
- [x] `gofumpt -d` + `goimports -l` clean
- [ ] e2e suite isn't updated — it uses `vitess/lite:mysql80` _(rolling)_ which by design won't get `--cell`. When the e2e tag pins to a v25+ release it will start emitting automatically

## Release notes

Added a callout in `docs/release-notes/2_18_0_summary.md`

---

_Claude Code assisted with the implementation and prepared this PR summary — I provided direction and reviewed_

### DIFF
--- a/docs/release-notes/2_18_0_summary.md
+++ b/docs/release-notes/2_18_0_summary.md
@@ -1,0 +1,13 @@
+## Major Changes
+
+### Table of Contents
+
+- **[VTOrc `--cell` flag auto-applied on Vitess v24+](#vtorc-cell-flag)**
+
+### <a id="vtorc-cell-flag"/>VTOrc `--cell` flag auto-applied on Vitess v24+</a>
+
+VTOrc deployments now receive `--cell=<cell>` automatically when the configured image tag parses to Vitess v24 or newer. This is required for Vitess v25+ where `--cell` is a hard requirement ([vitessio/vitess#20048](https://github.com/vitessio/vitess/pull/20048)) and was introduced in v24 ([vitessio/vitess#19047](https://github.com/vitessio/vitess/pull/19047)).
+
+The flag is **only** emitted when the version is parseable from the image tag (e.g. `vitess/lite:v24.0.0-mysql80`). Rolling tags such as `vitess/lite:mysql80` or `vitess/lite:latest`, or digest-only references, do not get the flag — pin a versioned tag, or set `cell` explicitly via `vitessOrchestrator.extraFlags` if you need it.
+
+Pre-v24 users are unaffected.
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.26.2
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
+	github.com/blang/semver/v4 v4.0.0
 	github.com/google/uuid v1.6.0
 	github.com/planetscale/operator-sdk-libs v0.0.0-20220216002626-1af183733234
 	github.com/prometheus/client_golang v1.23.2
@@ -85,7 +86,6 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.41.10 // indirect
 	github.com/aws/smithy-go v1.24.3 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/cihub/seelog v0.0.0-20170130134532-f561c5e57575 // indirect
--- a/pkg/operator/vitess/version.go
+++ b/pkg/operator/vitess/version.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2026 PlanetScale Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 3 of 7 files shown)


mhamza15 (commented):
I feel slightly iffy on parsing the version from the image tag. I understand there isn't any existing pattern for gating on version, but maybe we should introduce a stronger pattern, maybe adding a new field in the CRD(s)? It's probably worth asking some of our fellow Kubernetes operator experts on how this is usually implemented, maybe they have some good ideas.

mattlord:
Closing this as stale. Please feel free to re-open it if you want to come back to it.

PR #797, opened 2026-06-29
Couple tablet refresh interval and availability gate in the operator

### LINKED ISSUE #790: [Feature request] Make tabletAvailableSeconds configurable to align with vtgate's tablet_refresh_interval
## Summary

`tabletAvailableSeconds` is hardcoded to `30s` in [`reconcile_tablets.go`](https://github.com/planetscale/vitess-operator/blob/v2.11.1/pkg/controller/vitessshard/reconcile_tablets.go#L47-L53), but vtgate's default `--tablet_refresh_interval` is `1m`. When a tablet restarts and gets a new IP, vtgates rediscover it by polling the topology — this can take up to the full `tablet_refresh_interval`. Because the operator considers a tablet **Available** after only 30s, it proceeds to drain the next replica before most vtgates have discovered the first one, causing **"no healthy tablet available for `tablet_type:REPLICA`"** errors during rolling restarts.

We would like `tabletAvailableSeconds` to be exposed as a configurable field on the `VitessCluster` / `VitessShard` CRD so operators can align it with their vtgate topology poll interval without forking the operator or tuning vtgate flags.

---

### The timing contract that `tabletAvailableSeconds` enforces

`tabletAvailableSeconds` is meant to create a safety buffer: the operator only calls a tablet **Available** — and therefore safe to trigger the next rolling step — after it has been continuously Ready for at least that many seconds. The assumption is that vtgates will have finished discovering the new tablet endpoint within that window.

### Why the default breaks that contract

vtgate polls the topology for tablet health on the interval set by `--tablet_refresh_interval` (default **1 minute**). The poll is not edge-triggered. A tablet that becomes Ready just after a poll will not be discovered by vtgate until the *next* poll — up to `tablet_refresh_interval` later.

With `tabletAvailableSeconds = 30s` and `tablet_refresh_interval = 60s` (default):

```
t=0s   Replica A restarts, pod becomes Ready, gets a new IP
t=1s   Most vtgates polled just before t=0 — they do not know about A yet
t=30s  Operator marks A as Available; begins draining Replica B
t=31s  Replica B goes NOT_SERVING
       → vtgates that have not yet discovered A still only know about B (now NOT_SERVING)
       → those vtgates have zero healthy REPLICA tablets
       → "no healthy tablet available for 'tablet_type:REPLICA'"
t=51s  First vtgates discover A is healthy — errors stop for them
t=60s  Last vtgates discover A — all vtgates recover
```

The exposure window = `actual discovery lag − tabletAvailableSeconds`. We observed discovery lags of **45–51 seconds** in production, leaving a window of up to **21 seconds** where errors occur.

---

### Observed vtgate discovery lag

After Replica A restarted and began serving at `t=0`, vtgates discovered it over a **51-second window** across 10 vtgate pods:

```
[t+0s]   vtgate-01:  replica-A false => true (REPLICA)   ← 1st: immediate
[t+44s]  vtgate-02:  replica-A false => true (REPLICA)   ← 2nd: 44s later
[t+46s]  vtgate-03:  replica-A false => true (REPLICA)
[t+46s]  vtgate-04:  replica-A false => true (REPLICA)
[t+46s]  vtgate-05:  replica-A false => true (REPLICA)
[t+46s]  vtgate-06:  replica-A false => true (REPLICA)
[t+46s]  vtgate-07:  replica-A false => true (REPLICA)
[t+46s]  vtgate-08:  replica-A false => true (REPLICA)
[t+48s]  vtgate-09:  replica-A false => true (REPLICA)
[t+51s]  vtgate-10:  replica-A false => true (REPLICA)   ← last: 51s later
```

Only **1 of 10 vtgates** discovered the restarted replica immediately. The remaining 9 took 44–51 seconds — consistent with `--tablet_refresh_interval=1m`.

### How the errors occurred (two bursts)

**Burst 1 (~24 errors over ~6s):**

The operator considered Replica A Available ~30s after it started serving. Shortly after, the operator scheduled a drain on Replica B. When Replica B went NOT_SERVING (MySQL socket gone during pod restart), 9 of 10 vtgates had still not discovered Replica A — they saw **zero REPLICA tablets**.

```
Replica B: true => false (REPLICA)
           error: dial unix /vt/socket/mysql.sock: no such file or directory
           → 9/10 vtgates have zero healthy REPLICA tablets
           → "no healthy tablet available for 'tablet_type:REPLICA'"
```

**Burst 2 (~14 errors over ~7s):**

After PRS, the former primary briefly became a REPLICA and then its pod was restarted. Two vtgates that had not yet discovered Replica B as serving again had zero REPLICA tablets.

```
Former primary: true => false (REPLICA)
                error: dial unix /vt/socket/mysql.sock: no such file or directory
vtgate-09:  no healthy tablet available for 'tablet_type:REPLICA'
vtgate-10:  no healthy tablet available for 'tablet_type:REPLICA'
```

### Why `tabletAvailableSeconds = 30s` was the trigger

The operator marked Replica A as Available after 30s — well before the 9 slowest vtgates had discovered it. If `tabletAvailableSeconds` had been ≥ 55s (the observed worst-case discovery lag), the operator would have waited until all vtgates knew about Replica A before draining Replica B, and no errors would have occurred.

---

## Workaround

We resolved the issue by reducing vtgate's `--tablet_refresh_interval` from the default `1m` to `25s` (below `tabletAvailableSeconds = 30s`):

```
--tablet_refresh_interval=25s
```

With this setting, all vtgates discover a restarted tablet within ~25s — safely before the 30s `tabletAvailableSeconds` window expires — and the "no healthy tablet available" errors have not recurred.

**Trade-off:** A shorter `tablet_refresh_interval` increases the polling load on the global topo server (etcd/ZooKeeper). In large clusters with many vtgate pods and keyspaces, this is measurable and may not be acceptable. The cleaner fix is to keep `tablet_refresh_interval` at its default and instead raise `tabletAvailableSeconds` so it exceeds the actual discovery time — which requires making it configurable.

---

## Proposed Solution

Expose `tabletAvailableSeconds` (or a `tabletAvailableTimeout` duration) as an optional field on `VitessClusterSpec`, defaulting to `30s` for backwards compatibility:

```yaml
spec:
  tabletAvailableTimeout: 90s   # default: 30s; set to > tablet_refresh_interval
```

**Sizing guidance:** set to at least `tablet_refresh_interval + a small buffer` (e.g. `× 1.5`). With the default `--tablet_refresh_interval=1m`, **90s** is a safe value. With a tuned `--tablet_refresh_interval=25s`, the current `30s` default is sufficient.

---

## References

- [`reconcile_tablets.go` L47–53](https://github.com/planetscale/vitess-operator/blob/v2.11.1/pkg/controller/vitessshard/reconcile_tablets.go#L47-L53) — hardcoded `tabletAvailableSeconds = 30`
- [vtgate `--tablet_refresh_interval`](https://vitess.io/docs/archive/16.0/reference/programs/vtgate/) — default `1m`

mcrauwel:
## Summary

Fixes #790.

The operator previously marked a restarted tablet **Available** after a hardcoded `30s` and could proceed to drain the next replica before vtgate's default `60s` topology refresh had rediscovered the first tablet. That could temporarily leave vtgate with no healthy replica during rolling restarts.

This PR adds a single optional `spec.tabletRefreshInterval` setting. The operator uses it for both:

- vtgate's `--tablet-refresh-interval` flag, and
- the VitessShard availability gate that paces tablet rollouts, calculated as `interval x 2` and rounded up to the next second.

The default remains `60s`, matching vtgate's existing default, which produces a `120s` tablet availability gate without requiring any configuration changes.

```yaml
spec:
  # optional; defaults to 60s
  # vtgate --tablet-refresh-interval = 60s
  # tablet availability gate         = 120s
  tabletRefreshInterval: 60s
```

The setting mainly exists so larger clusters can trade faster tablet rollouts for additional topology-server polling load.

## Rollout safety

Changing these two values independently could recreate the same availability gap, so the controllers stage updates using observed status:

- When increasing the interval, shard availability gates are raised before vtgate rolls out the longer refresh interval.
- When decreasing the interval, vtgate rolls out the shorter refresh interval before shard gates are lowered.
- A cell reports its observed interval only after every desired vtgate replica is running the current Deployment template.
- Existing legacy `tablet-refresh-interval` or `tablet_refresh_interval` extra flags are preserved until the coordinated rollout is safely staged. After that, the operator owns the flag and ignores either spelling in extra flags.

This also handles upgrades of existing clusters and cells added while an interval transition is in progress.

## Changes

- Adds and validates `spec.tabletRefreshInterval`, inherited through cells, keyspaces, and shards.
- Derives the tablet availability gate from the effective refresh interval.
- Emits the canonical `--tablet-refresh-interval` vtgate flag once rollout coordination is active.
- Publishes observed intervals through cell, keyspace, and shard status to coordinate safe transitions.
- Fixes the availability requeue duration, which previously interpreted seconds as nanoseconds, and avoids another full-window delay at the strict availability boundary.
- Updates generated CRDs and API documentation.
- Adds unit and integration coverage for validation, propagation, flag ownership, legacy migration, rollout ordering, availability calculation, and boundary requeue behavior.

---

_Most of this was written by Claude Code; I provided direction and review._


docs/api/index.html
```diff
@@ -449,6 +449,28 @@ <h3 id="planetscale.com/v2.VitessCluster">VitessCluster
...
+of a rolling update (such as draining the next tablet). This creates a
+safety buffer that accounts for the time it takes for vtgates to discover
+that the tablet is Ready and update their routing tables. If a tablet is
+Ready but vtgates don&rsquo;t know it yet, then it isn&rsquo;t actually available for
+serving queries.</p>
+<p>This should be set to at least vtgate&rsquo;s &ndash;tablet_refresh_interval (default
+1m) plus a small buffer, otherwise the operator may drain the next tablet
+before vtgates have discovered the previously restarted one, causing
+&ldquo;no healthy tablet available&rdquo; errors during rolling restarts.</p>
+<p>Default: 30</p>
```
mattlord: I wonder when and why a user would configure this? Should we always use the vitess cluster resource's vtgate tablet-refresh-interval value * 2? This feels like a potential footgun to me, when I would say that always using something greater than the defined tablet refresh interval is simply the correct thing to do...

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -116,6 +116,23 @@ type VitessClusterSpec struct {
...
 
+	// TabletAvailableSeconds is how long a tablet Pod must be consistently Ready
+	// before the operator considers it Available and proceeds to the next step
+	// of a rolling update (such as draining the next tablet). This creates a
+	// safety buffer that accounts for the time it takes for vtgates to discover
+	// that the tablet is Ready and update their routing tables. If a tablet is
+	// Ready but vtgates don't know it yet, then it isn't actually available for
+	// serving queries.
+	//
+	// This should be set to at least vtgate's --tablet_refresh_interval (default
```
mattlord: We should use the dashed version of the flag everywhere. The underscore variants will be going away at some point (you can use either today as part of the transition from underscores to dashes for all flags in Vitess).

mattlord (commented):
I love this new shape! ❤️ 

In `pkg/controller/vitessshard/reconcile_tablets.go` for the `tabletAvailableStatus`, it seems like we should requeue for the remaining time until the pod can become Available, not for the full derived window every time. With the new default this is 90s, so if a reconcile happens after the pod has already been Ready for 89s, we can still wait another 90s before marking it Available. That can almost double each rolling restart step. Kubernetes handles this as `Ready.LastTransitionTime + minReadySeconds - now`; I think that we should do the same here, with a small floor since this podutils version uses a strict `Before` check.

In `pkg/operator/vtgate/deployment.go` for the `baseFlags`, we should emit the canonical dashed Vitess flag name, `tablet-refresh-interval`, instead of `tablet_refresh_interval`, and update the tests/docs to match. Vitess still normalizes underscores today, but use of underscores is deprecated and the support for it will go away in a future release.

Otherwise it LGTM! Thank you, @mcrauwel !

pkg/apis/planetscale/v2/vitesscell_defaults.go
```diff
@@ -27,6 +28,9 @@ func DefaultVitessCell(vtc *VitessCell) {
 	DefaultLocalLockserver(&vtc.Spec.Lockserver)
 	DefaultVitessGateway(&vtc.Spec.Gateway)
 	DefaultTopoReconcileConfig(&vtc.Spec.TopologyReconciliation)
+	if vtc.Spec.TabletRefreshInterval == nil {
+		vtc.Spec.TabletRefreshInterval = &metav1.Duration{Duration: DefaultTabletRefreshInterval}
+	}
```
frouioui: We should create a `DefaultTabletRefreshInterval` function and follow the same pattern we're doing for the other fields.

pkg/apis/planetscale/v2/vitessshard_defaults.go
```diff
@@ -27,6 +28,9 @@ import (
 func DefaultVitessShard(dst *VitessShard) {
 	DefaultUpdateStrategy(&dst.Spec.UpdateStrategy)
 	DefaultTopoReconcileConfig(&dst.Spec.TopologyReconciliation)
+	if dst.Spec.TabletRefreshInterval == nil {
+		dst.Spec.TabletRefreshInterval = &metav1.Duration{Duration: DefaultTabletRefreshInterval}
+	}
```
frouioui: Same comment as https://github.com/planetscale/vitess-operator/pull/797/changes#r3515378256

pkg/operator/vtgate/deployment.go
```diff
@@ -288,6 +288,16 @@ func (spec *Spec) baseFlags() vitess.Flags {
...
 		"grpc_port":   planetscalev2.DefaultGrpcPort,
 	}
+
+	// The operator owns --tablet_refresh_interval so it stays consistent with
+	// the tablet-availability gate the VitessShard controller derives from the
+	// same value. It's normally defaulted by DefaultVitessCell; guard nil in
+	// case defaulting hasn't run.
+	if spec.Cell.TabletRefreshInterval != nil {
+		flags["tablet_refresh_interval"] = spec.Cell.TabletRefreshInterval.Duration.String()
+	}
```
frouioui: We should start using the dash-version of the flag.

frouioui (commented):
As the api specs are changing you will need to do this before merging: https://github.com/planetscale/vitess-operator/pull/629#discussion_r1828138492

docs/api/index.html
mcrauwel: The main use case is large clusters where the 60s default makes rolling restarts painfully slow — at 120s per step (now with the ×2 multiplier you suggested), a shard with 6 replicas takes 12 minutes to roll. Operators of big clusters may want to bring that down by tuning a shorter refresh interval, accepting the tradeoff of more topology-server polling load.

The footgun risk you're pointing at is real, but the operator owns both sides of the equation — the vtgate flag and the availability gate are always `interval × 2`, so even if someone sets `tabletRefreshInterval: 10s`, they can't accidentally drift the two apart the way they could with the old `tabletAvailableSeconds` knob. The only remaining risk is setting it so low that vtgates can't reliably refresh within one cycle — and at that point it's a documented knob with explicit "don't set this directly via ExtraFlags" guidance.

That said, if you feel the coupling guarantee is sufficient and the knob isn't worth the added API surface, I'm comfortable making it internal-only (operator always uses vtgate's built-in default × 2, no user-facing field). Happy to go either way — just wanted to make sure the intended use case was clear before removing it.

pkg/apis/planetscale/v2/vitesscell_defaults.go
mcrauwel: Done — renamed the constant to unexported `defaultTabletRefreshInterval` and introduced an exported `DefaultTabletRefreshInterval(**metav1.Duration)` following the same pattern as `DefaultUpdateStrategy`. Both `DefaultVitessShard` and `DefaultVitessCell` now delegate to it.

pkg/apis/planetscale/v2/vitessshard_defaults.go
mcrauwel: Done here too.

pkg/operator/vtgate/deployment.go
mcrauwel: Done — switched to `tablet-refresh-interval` (dashes) in both `deployment.go` and its test, and updated the flag name in the `VitessClusterSpec` comment.

mcrauwel:
Re frouioui's operator.yaml comment: done — regenerated CRDs via `make generate`, built via `kubectl kustomize ./deploy`, and spliced the updated schemas into both test operator YAMLs while preserving their test-specific Deployment sections.

mattlord (commented):
In `test/integration/vitesscluster/vitesscluster_test.go:219-230` It seems like this test still expects the old underscore flag spelling, but `pkg/operator/vtgate/deployment.go` now emits `--tablet-refresh-interval=40s`. I think we should update the assertion and nearby comment to the dashed spelling, otherwise this integration test should fail even though the production code was updated correctly.

In `pkg/operator/vtgate/deployment.go:234-240` I think we should enforce the new “operator owns tablet-refresh-interval” contract rather than only documenting it. `baseFlags()` sets `tablet-refresh-interval` from `spec.Cell.TabletRefreshInterval`, but then `ExtraVitessFlags` / gateway `ExtraFlags` are applied afterward and can still override `tablet-refresh-interval`, or add the underscore alias as a second normalized spelling. That seems like it can recreate the exact drift this PR is trying to remove. I think we should ignore or reject both `tablet-refresh-interval` and `tablet_refresh_interval` from vtgate extra flags, and add a small test for both spellings.

This is looking good, @mcrauwel ! ❤️

mcrauwel:
Thanks @mattlord — both addressed in 0fc55ba:

1. **Integration test spelling**: updated the assertion and comment in `vitesscluster_test.go` to the dashed `--tablet-refresh-interval=40s`.

2. **Enforcing the contract**: extracted the vtgate extra-flags overlay into `applyExtraFlags()`, which now ignores `tablet-refresh-interval` and `tablet_refresh_interval` (leading dashes tolerated, as before) from user-provided flags. Cluster-level `ExtraVitessFlags` and gateway `ExtraFlags` are merged into a single map by the VitessCell controller before reaching this point, so the one guard covers both paths. Added `TestApplyExtraFlagsIgnoresTabletRefreshInterval` covering both spellings (plus a dashed-prefix variant), asserting the operator-derived value survives, no underscore alias sneaks in as a second flag, and unrelated extra flags still apply. Also updated the `TabletRefreshInterval` API doc from "do NOT set this via extra flags" to "the operator ignores it there", and regenerated the docs.

Unit and integration suites pass locally.

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -376,26 +369,51 @@ func isTabletPrimary(ctx context.Context, vts *planetscalev2.VitessShard, tablet
...
 	// If the Pod is being deleted, we immediately mark it unavailable even
 	// though it might not have transitioned to Unready yet.
 	if pod.DeletionTimestamp != nil {
 		return corev1.ConditionFalse
 	}
 
+	// TabletRefreshInterval is a nillable field inherited from the parent
+	// VitessCluster. It's normally filled in by DefaultVitessShard, but guard
+	// against a nil value to be safe.
+	refreshInterval := metav1.Duration{Duration: time.Minute}
```
mhamza15: Nit: we're hardcoding `time.Minute` here, we could use [`defaultTabletRefreshInterval`](https://github.com/planetscale/vitess-operator/pull/797/changes#diff-fb546cb0aca88a6cfcc3327d36ad94358b1eb39eec38170f656fc27db21c1586R99) to stay consistent. Or could maybe use `EffectiveTabletRefreshInterval`?

mhamza15 (approved):

pkg/apis/planetscale/v2/defaults.go
```diff
@@ -185,3 +205,54 @@ var (
...
+		seconds++
+	}
+	if seconds < 1 {
+		seconds = 1
+	}
+	if seconds > 1<<31-1 {
+		seconds = 1<<31 - 1
+	}
+	return int32(seconds)
+}
```
frouioui, line 225: Can we simplify this function and just do `refreshInterval.Seconds()`?

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -638,6 +662,9 @@ type VitessClusterCellStatus struct {
 	PendingChanges string `json:"pendingChanges,omitempty"`
 	// GatewayAvailable indicates whether the vtgate service is fully available.
 	GatewayAvailable corev1.ConditionStatus `json:"gatewayAvailable,omitempty"`
+	// TabletRefreshInterval records the completed vtgate rollout so shard gates
+	// are not lowered while older Pods still use a longer interval.
+	TabletRefreshInterval *metav1.Duration `json:"tabletRefreshInterval,omitempty"`
```
frouioui: Why do we need this value in the status?

pkg/apis/planetscale/v2/vitesscluster_types.go
```diff
@@ -116,6 +116,30 @@ type VitessClusterSpec struct {
...
+	// This field is the only supported way to change the interval: the
+	// operator ignores tablet-refresh-interval (in either flag spelling) in
+	// ExtraVitessFlags and the gateway ExtraFlags, since overriding it there
+	// would bypass this coupling and reintroduce the bug.
+	// Existing clusters using either extra flag must move that value here when
+	// upgrading; the legacy flag is retained only until the safe rollout is staged.
+	//
+	// Default: 60s
+	// +kubebuilder:validation:XValidation:rule="self.matches('^([0-9]+([.][0-9]+)?(ns|us|ms|s|m|h))+$') && duration(self) >= duration('1s')",message="tabletRefreshInterval must be a valid duration of at least 1s"
+	TabletRefreshInterval *metav1.Duration `json:"tabletRefreshInterval,omitempty"`
```
frouioui: Since we only support >1s, can we simplify the validation and only accept units >= `s`

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -98,6 +98,11 @@ type VitessShardSpec struct {
 
 	// UpdateStrategy is inherited from the parent's VitessClusterSpec.
 	UpdateStrategy *VitessClusterUpdateStrategy `json:"updateStrategy,omitempty"`
+
+	// TabletRefreshInterval is inherited from the parent's VitessClusterSpec.
+	// The operator derives the tablet-availability gate from it.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^([0-9]+([.][0-9]+)?(ns|us|ms|s|m|h))+$') && duration(self) >= duration('1s')",message="tabletRefreshInterval must be a valid duration of at least 1s"
+	TabletRefreshInterval *metav1.Duration `json:"tabletRefreshInterval,omitempty"`
```
frouioui: same as https://github.com/planetscale/vitess-operator/pull/797/changes#r3626519400

pkg/controller/vitesscell/reconcile_vtgate.go
```diff
@@ -175,14 +178,22 @@ func (r *ReconcileVitessCell) reconcileVtgate(ctx context.Context, vtc *planetsc
...
 				status.Available = available.Status
 			}
+			if deploymentRolloutComplete(curObj) {
+				interval, err := deploymentTabletRefreshInterval(curObj)
+				if err != nil {
+					resultBuilder.Error(err)
+				} else {
+					status.TabletRefreshInterval = interval.DeepCopy()
+				}
+			}
```
frouioui, line 188: I am confused as to what is the intent of this status field. What are we using it for, and why do we need it?

pkg/apis/planetscale/v2/vitesskeyspace_types.go
```diff
@@ -83,6 +83,10 @@ type VitessKeyspaceSpec struct {
 
 	// UpdateStrategy is inherited from the parent's VitessClusterSpec.
 	UpdateStrategy *VitessClusterUpdateStrategy `json:"updateStrategy,omitempty"`
+
+	// TabletRefreshInterval is inherited from the parent's VitessClusterSpec.
+	// +kubebuilder:validation:XValidation:rule="self.matches('^([0-9]+([.][0-9]+)?(ns|us|ms|s|m|h))+$') && duration(self) >= duration('1s')",message="tabletRefreshInterval must be a valid duration of at least 1s"
+	TabletRefreshInterval *metav1.Duration `json:"tabletRefreshInterval,omitempty"`
```
frouioui, line 89: Same as https://github.com/planetscale/vitess-operator/pull/797/changes#r3626519400

frouioui (commented):
I am not sure why we need to modify the status so heavily. Left some questions about it.

Thanks for updating the tests! We also need to do the vitess part of https://github.com/planetscale/vitess-operator/pull/629#discussion_r1828138492, to updated the CRDs there too

pkg/apis/planetscale/v2/vitesscluster_types.go
mattlord: I think we need this status value because the spec only represents the desired interval. When increasing the interval, the shard gate must expand before the vtgate rollout; when decreasing it, every vtgate must finish its rollout before the gate shrinks. Recording the interval actually observed after rollout lets the independent controllers preserve that ordering across reconciles and operator restarts. I updated the field comment to make that purpose explicit.

pkg/controller/vitesscell/reconcile_vtgate.go
mattlord: This is the observed side of the rollout handshake. We only publish the interval after the vtgate Deployment rollout is complete, so the cluster controller can distinguish the desired spec from what every vtgate is actually running. That lets us expand the shard gate before increasing vtgate's interval, and wait for all vtgates before shrinking the gate when decreasing it. Without this status, independent reconciles could shrink the gate too early.

pkg/apis/planetscale/v2/defaults.go
mattlord: Updated this to use `refreshInterval.Seconds()`. I kept `math.Ceil` around the multiplied value because the availability check takes whole seconds, and rounding a fractional interval down would make the gate shorter than `interval × 2`.

pkg/apis/planetscale/v2/vitesscluster_types.go
mattlord: Updated the validation to accept only `s`, `m`, and `h` units while still requiring a duration of at least `1s`. The CRDs and API docs have been regenerated.

pkg/apis/planetscale/v2/vitessshard_types.go
mattlord: Updated this validation as well to accept only `s`, `m`, and `h`, and regenerated the shard CRD and API docs.

pkg/apis/planetscale/v2/vitesskeyspace_types.go
mattlord: Updated this validation as well to accept only `s`, `m`, and `h`, and regenerated the keyspace CRD and API docs.

mattlord:
The operator-side CRDs and `test/endtoend/operator/operator-latest.yaml` are regenerated in this PR. The matching `vitessio/vitess/examples/operator/operator.yaml` update is the remaining cross-repository follow-up and will need a companion Vitess PR before these changes merge.

frouioui (approved):

PR #798, opened 2026-06-30
Allow configuring vtbackup storage separately from tablet pools (#661)

### LINKED ISSUE #661: [Feature] Allow vtbackup pod spec to be override-able.
## DESCRIPTION

Expose the vtbackup pod spec via the VitessCluster object spec. This way, we can change the vtbackup pod spec to be different from vttablet pod spec, which it currently inherits the spec from.

## USE CASE

We rely on the vtbackup's feature of `--initial_backup` to create an empty backup during shard creation. Since the spec is inherited from vttablet, PVCs are set and disks are allocated to these pods. In large clusters, this often leads to additional capacity allocation that sometimes is unavailable.
Ideally we change the spec and remove the PVC allocation or change the `storageclass` to different disks.

mcrauwel:
## Summary

Fixes #661.

This change adds an optional `vtbackup` block alongside `tabletPools` in `VitessShardTemplate`. It allows vtbackup Pods to use scratch storage that differs from the tablet-pool storage they otherwise inherit.

Historically, vtbackup Pods inherited the first tablet pool's `dataVolumeClaimTemplate`. That behavior allocates a PVC and an underlying disk per shard, even for the initial backup of an empty database. Large clusters can therefore allocate substantial unused storage or fail to bootstrap when those PVCs cannot bind.

## Configuration

Set `useEmptyDirForInitialBackup` to let the initial backup use ephemeral scratch space while tablet Pods retain their normal PVCs:

```yaml
spec:
  keyspaces:
  - name: commerce
    partitionings:
    - equal:
        parts: 2
        shardTemplate:
          vtbackup:
            useEmptyDirForInitialBackup: true
          tabletPools:
          - cell: zone1
            type: replica
            replicas: 3
            dataVolumeClaimTemplate:
              accessModes: [ReadWriteOnce]
              resources:
                requests:
                  storage: 100Gi
            vttablet: {}
            mysqld: {}
```

A shard template can instead provide a dedicated PVC template for vtbackup Pods:

```yaml
vtbackup:
  dataVolumeClaimTemplate:
    storageClassName: cheap-disk
    accessModes: [ReadWriteOnce]
    resources:
      requests:
        storage: 10Gi
```

`useEmptyDirForInitialBackup: true` and `dataVolumeClaimTemplate` are mutually exclusive. The CRD rejects configurations that set both.

## Semantics

| `vtbackup` configuration | Initial backup | Scheduled backup |
|---|---|---|
| Block omitted or empty | Inherits the first tablet pool's PVC template | Inherits the first tablet pool's PVC template |
| `useEmptyDirForInitialBackup: true` | Uses ephemeral `emptyDir` scratch space | Inherits the first tablet pool's PVC template |
| `dataVolumeClaimTemplate` set | Uses the vtbackup-specific PVC template | Uses the vtbackup-specific PVC template |

Scheduled backups retain persistent scratch storage when `useEmptyDirForInitialBackup` is set because vtbackup restores the latest full database before taking a new backup. Using an unbounded `emptyDir` for that workload could cause node disk pressure or eviction.

Omitting the block, supplying an empty block, or explicitly setting `useEmptyDirForInitialBackup: false` preserves the historical behavior. Tablet Pod storage is unaffected.

## Storage reconciliation

The operator records a hash of the desired vtbackup PVC template on the initial-backup Pod and PVC. It separately records a hash of the Pod's effective `/vt/vtdataroot` mount and referenced volume source on the Pod. This accounts for `extraVolumes` and `extraVolumeMounts` that override the default data mount without changing the PVC template.

When the storage configuration changes, the operator:

1. Preserves a running initial-backup Pod.
2. Recreates a non-running initial-backup Pod when its effective data mount or volume source changes.
3. Replaces the PVC, after the old Pod is gone, only when the desired PVC template changes.
4. Stops and retries if it cannot verify Pod absence before replacing a PVC, preventing a replacement Pod from starting against stale storage.

For upgrades, the controller adopts hashless resources and resources created by the earlier single-hash implementation when their effective Pod and PVC configuration already matches the desired configuration. The comparison accounts for Kubernetes-assigned defaults and bindings without conflating an explicit empty storage class or a user-prebound volume with controller-assigned values. If the controller cannot read a legacy PVC, it preserves the Pod and retries instead of making a destructive decision from incomplete state.

This allows the controller to converge across PVC-to-`emptyDir`, `emptyDir`-to-PVC, custom-PVC-template, and effective Pod-volume changes. In particular, setting `useEmptyDirForInitialBackup: true` can recover an initial backup that is Pending because its inherited PVC cannot bind.

## Scope

This change only separates vtbackup data-volume configuration from tablet-pool storage. Broader vtbackup Pod overrides can build on the dedicated `vtbackup` block in follow-up work.

## Tests

- Unit coverage for inherited, explicit `emptyDir`, and custom templates on initial and scheduled backups.
- Scheduled-backup coverage confirming that `useEmptyDirForInitialBackup` applies only to the initial backup.
- Reconciliation coverage for PVC-to-`emptyDir`, `emptyDir`-to-PVC, custom-template, and effective Pod-volume transitions.
- Coverage confirming that effective Pod-volume changes preserve an unchanged PVC.
- Legacy-adoption coverage for hashless and earlier single-hash resources, Kubernetes-defaulted PVC fields, explicit empty storage classes, user-prebound volumes, and transient PVC lookup failures.
- Fault-injection coverage confirming that uncertain Pod or PVC reads preserve existing resources and retry safely.
- Coverage confirming that storage changes do not interrupt a running initial backup.
- Integration coverage confirming that `useEmptyDirForInitialBackup: true` removes the initial-backup PVC while tablet Pods retain theirs.
- Generated CRDs, deepcopy code, API reference documentation, and the bundled operator manifest are updated.

Local verification:

- `go test -count=1 ./pkg/...`
- `go test -race -count=1 ./pkg/controller/vitessshard ./pkg/controller/vitessbackupschedule ./pkg/operator/vttablet`
- Changed-line `golangci-lint`: 0 issues
- `git diff --check`


pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
```diff
@@ -683,7 +684,11 @@ func (r *ReconcileVitessBackupsSchedule) createVtbackupJob(
 		return nil, err
 	}
 
-	// Create the corresponding PVC for the new vtbackup pod
+	if vtbackupSpec.TabletSpec.DataVolumePVCSpec == nil {
+		return job, nil
+	}
```
frouioui, line 688: We should fail with an error here instead. Otherwise the job we create won't have any corresponding PVC and will fail.

pkg/controller/vitessshard/reconcile_backup_job.go
```diff
@@ -214,6 +214,17 @@ func vtbackupSpec(key client.ObjectKey, vts *planetscalev2.VitessShard, parentLa
...
+	// By default, vtbackup Pods inherit the first pool's data volume. A shard-level
+	// vtbackup override can replace that template or remove it entirely: when
+	// the override block is present but its DataVolumeClaimTemplate is nil,
+	// vtbackup Pods run with no PVC and fall back to ephemeral (emptyDir)
+	// scratch space. A nil DataVolumePVCSpec downstream means "no PVC", so no
+	// PVC is reconciled for the Pod (see reconcile loop above).
+	dataVolumeClaimTemplate := pool.DataVolumeClaimTemplate
+	if vts.Spec.Vtbackup != nil {
+		dataVolumeClaimTemplate = vts.Spec.Vtbackup.DataVolumeClaimTemplate
+	}
```
frouioui, line 308: `vts.Spec.Vtbackup.DataVolumeClaimTemplate` is an optional field, we should only override the `dataVolumeClaimTemplate` if we the user has defined something in the CR.

pkg/apis/planetscale/v2/vitessshard_types.go
```diff
@@ -116,6 +116,11 @@ type VitessShardTemplate struct {
 	// +listMapKey=name
 	TabletPools []VitessShardTabletPool `json:"tabletPools,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 
+	// Vtbackup can optionally override the data volume used by the initial and
+	// scheduled vtbackup Pods for this shard. When omitted, vtbackup Pods inherit
+	// the first tablet pool's DataVolumeClaimTemplate.
+	Vtbackup *VitessShardVtbackup `json:"vtbackup,omitempty"`
```
frouioui: Since we are modifying the CRD, we need to re-generate the operator yaml files, the test files, and port the related changes to the vitess repository, like so: https://github.com/planetscale/vitess-operator/pull/629#discussion_r1828138492

mattlord: We can do one vitessio/vitess PR once this and https://github.com/planetscale/vitess-operator/pull/797 are both merged.

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
mattlord: I looked at this more closely and I don't think the Job references a missing PVC. `NewBackupPod` derives both the volume and mounts from `DataVolumePVCSpec`; when it is nil, it omits the PVC volume and mounts `/vt/vtdataroot` from the existing `vt-root` `emptyDir`. That no-PVC Job is the behavior requested by #661. I've restored this path locally and strengthened the test to assert that the Job has no PVC-backed volume and that no PVC object is created.

pkg/controller/vitessshard/reconcile_backup_job.go
mattlord: I think the nil inner field is intentional here. The two pointers provide the three states this API needs: no `vtbackup` block inherits the pool PVC; an empty block selects `emptyDir`; a non-nil `dataVolumeClaimTemplate` selects a custom PVC. Checking both pointers collapses the first two states and leaves no way to request the no-PVC behavior from #661. I've restored the three-state handling and its tests in c2e444c.

pkg/controller/vitessbackupschedule/vitessbackupschedule_controller.go
mattlord: Addressed in c2e444c.

frouioui (approved):

mattlord (approved):

hallaroo:
Customer informed that this is merged: https://docs.google.com/document/d/1MSMamqiCXn9mhXEtB_VjCHETgmEpabM0E9UZly1setc/edit?tab=t.0#bookmark=id.1bjw329uqp8b

PR #800, opened 2026-07-06
Allow multiple tablet pools of the same type

absolutejam:
Removes the external data store requirement for multiple tabletpools.

Also adds `--default_pool_names` flag which can optionally be supplied to keep backwards compatible behaviour for the specified pools.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -8162,10 +8162,8 @@ string
 </td>
 <td>
 <p>Name is the pool&rsquo;s unique name within the (cell,type) pair.
-This field is optional, and defaults to an empty.
-Assigning different names to this field enables the existence of multiple pools with a specific tablet type in a given cell,
-which can be beneficial for unmanaged tablets.
-Hence, you must specify ExternalDatastore when assigning a name to this field.</p>
+This field is optional, and defaults to an empty string.
+Assigning different names to this field enables the existence of multiple pools with a specific tablet type in a given cell.</p>
 </td>
 </tr>
 <tr>
--- a/docs/api/index.html
+++ b/docs/api/index.html
@@ -8164,10 +8164,8 @@ string
 </td>
 <td>
 <p>Name is the pool&rsquo;s unique name within the (cell,type) pair.
-This field is optional, and defaults to an empty.
-Assigning different names to this field enables the existence of multiple pools with a specific tablet type in a given cell,
-which can be beneficial for unmanaged tablets.
-Hence, you must specify ExternalDatastore when assigning a name to this field.</p>
+This field is optional, and defaults to an empty string.
+Assigning different names to this field enables the existence of multiple pools with a specific tablet type in a given cell.</p>
 </td>
 </tr>
 <tr>
--- a/pkg/apis/planetscale/v2/labels.go
+++ b/pkg/apis/planetscale/v2/labels.go
@@ -36,7 +36,6 @@ const (
 	// TabletTypeLabel is the key for identifying the Vitess target tablet type for a Pod.
 	TabletTypeLabel = LabelPrefix + "/" + "tablet-type"
 	// TabletPoolNameLabel is the key for identifying the Vitess target pool name within the (cell,type) pair.
-	// This label is applicable to Vitess-unmanaged keyspaces.
 	TabletPoolNameLabel = LabelPrefix + "/" + "pool-name"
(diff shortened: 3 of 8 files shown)


absolutejam:
I didn't use table driven tests because I didn't see them used elsewhere and the coverage is only rudimentary, but the tests passed for me and I've deployed this fork into our internal Vitess clusters.

In our environments, I have also leveraged the `--default_pool_names` on our initial `replica` tablet pool (called `default`), so that it didn't churn UIDs (notably the PVCs).

mattlord (commented):
Thanks, @absolutejam ! ❤️ There are a few things that I think we should try and address before merging (sometimes it can be resolved in a discussion vs code changes).

I think that it's worth guarding against default-pool UID collisions in `pkg/controller/vitessshard/reconcile_tablets.go:286`. The new `--default_pool_names` behavior can make two pools in the same `(cell,type)` both use the legacy unnamed UID path. For example, an unnamed pool plus a default-named pool, or two named pools both listed in `--default_pool_names`, will generate the same tablet UIDs for the same indexes. That seems like it can make the desired tablet map overwrite one pool with the other, reusing the same pod/PVC/tablet alias identity. I think we should reject or otherwise prevent more than one “legacy UID” pool per `(cell,type)`, and add a regression test for that case.

I think that we should Include the pool name in “same pool” anti-affinity labels in `pkg/operator/vttablet/spec.go:67`. Named local pools can now represent distinct pools with the same cell/type, but `poolLabels()` still does not include `TabletPoolNameLabel`. That means the default anti-affinity that is described as applying to the “same exact pool” still groups separate named pools together. I think we should include the pool name label there when present so scheduling behavior matches the new pool identity model.

We will need to update the API/CRD wording for named local pools  
in `pkg/apis/planetscale/v2/vitessshard_types.go:147`  
`pkg/apis/planetscale/v2/labels.go:39`. The `Name` field comment still says a pool name requires `ExternalDatastore`, and the pool-name label comment still says it applies to Vitess-unmanaged keyspaces. Since this PR intentionally supports named local pools, I think we should update those comments and regenerate any CRD descriptions derived from them so users do not think the new configuration is unsupported.

Side note, the DCO is currently failing on the PR. If you click on that failed workflow it will give you precise instructions for how to fix it.

absolutejam:
Thanks for the thoughtful comments @mattlord - I think adding a guard for UID collisions is the right move.

I've updated the documentation now and restricted `--default_pool_name` to a single default pool as allowing multiple didn't really provide any benefit other than potentially allowing UID collisions. I also added some additional tests and a run-time guard too.

mattlord (commented):
1. **Blocking:** I think that we should return immediately when `vttabletSpecs()` fails. In `pkg/controller/vitessshard/reconcile_tablets.go:85`, the error is recorded but reconciliation continues with an empty desired set. `ReconcileObjectSet()` can then begin draining and deleting existing non-primary tablets, followed by their PVCs, because they all appear unwanted. A UID collision should fail safely without touching existing resources. I think we should add a reconciliation-level regression test that verifies no existing Pods or PVCs are modified or deleted.

2. **Blocking:** I think that `--default_pool_name` should only select the legacy UID path for a named, operator-managed pool. In `pkg/controller/vitessshard/reconcile_tablets.go:295`, setting this flag also makes unnamed pools use `UIDWithPoolName(..., "")` and makes a matching named `ExternalDatastore` pool stop using its existing name-aware UID. Enabling the compatibility flag for one pool can therefore churn unrelated tablet and PVC identities across the operator’s entire fleet. Empty names should always retain `UID()`, named external pools should always retain `UIDWithPoolName()`, and the override should apply only to named managed pools. I think a table-driven compatibility matrix would help pin this down.

3. **Blocking:** I think that disk reconciliation also needs to include the pool name. `pkg/controller/vitessshard/reconcile_disk.go:59` finds tablets using only `(cell,type)`, so each same-type managed pool checks every other pool’s PVC against its own requested size. When those pools use different disk sizes, one pool will wait forever for another pool’s PVC spec to match the wrong size, preventing the resize rollout from completing. I think we should filter using `TabletPoolNameLabel` and add a regression test with two same-type pools using different storage sizes.

4. **Non-blocking:** The PR description still refers to `--default_pool_names` and multiple “specified pools,” while the implementation provides the singular `--default_pool_name`. I think we should align that wording and clarify that the flag must remain configured while the named pool needs to retain its legacy UIDs.

Issue #801, opened 2026-07-08
Tablet UID Pod label strips leading zeros


absolutejam:
# Issue

Pod label `planetscale.com/tablet-uid` strips leading zeros.

# Root cause

The generated Pod label for UIDs with leading zeros is formatted with `strconv.FormatUint(uint64(tabletAlias.Uid), 10)`

https://github.com/planetscale/vitess-operator/blob/c0d5afeb1ec8dd4085dc6ed3e5cd947613a4deff/pkg/controller/vitessshard/reconcile_tablets.go#L298

Which results in leading zeros being removed, eg.  `0790125915` -> `790125915`.

While this doesn't affect the functionality of the operator or tablets, it makes it more difficult to correlate data because the UID is no longer correct. My use case is joining in Prometheus labels and I have some custom [Steampipe](https://steampipe.io/) connectors that allow me to query/join across Vitess, Kubernetes and other data-sources and I have to special-case the tablet UID.

I'll be happy to submit a PR shortly with a test case I have (a combination of labels that build the above uid).

# Version

Operator: v2.16.0

PR #802, opened 2026-07-08
Retain leading zeros for Pod uid label

absolutejam:
https://github.com/planetscale/vitess-operator/issues/801


absolutejam:
Just an FYI, I also have https://github.com/planetscale/vitess-operator/pull/800 open - so if either is accepted, the other will need to be rebased and fixed, but I didn't want to be presumptuous and based off the other branch.

pkg/controller/vitessshard/reconcile_tablets_test.go
```diff
@@ -0,0 +1,73 @@
+package vitessshard
+
```
frouioui, line 18: You will need a header comment here, similar to what `pkg/controller/vitessshard/reconcile_tablets.go` has.

frouioui (commented):
Looks good to me, and good catch. Just one comment

pkg/controller/vitessshard/reconcile_tablets.go
```diff
@@ -295,7 +296,7 @@ func vttabletSpecs(vts *planetscalev2.VitessShard, parentLabels map[string]strin
 				labels[k] = v
 			}
 			labels[planetscalev2.CellLabel] = tabletAlias.Cell
-			labels[planetscalev2.TabletUidLabel] = strconv.FormatUint(uint64(tabletAlias.Uid), 10)
+			labels[planetscalev2.TabletUidLabel] = fmt.Sprintf("%010d", tabletAlias.Uid)
```
mhamza15: Nit: it might be nice to extract this to a helper (maybe pkg/operator/vttablet/uid.go#UIDString` or similar) that other parts of the code can use. For example, here's another use-case: https://github.com/planetscale/vitess-operator/blob/f50e3b260ad31c191cd7f3e02c87d601ce92473b/pkg/operator/vttablet/spec.go#L95

mhamza15 (commented):
Codex caught that we'll need to update `test/endtoend/utils.sh` as well: https://github.com/planetscale/vitess-operator/blob/c0d5afeb1ec8dd4085dc6ed3e5cd947613a4deff/test/endtoend/utils.sh#L176 That drops the leading zeroes currently and won't match the selector anymore.

absolutejam:
- Removed the reformatting of `uid` in `utils.sh`
  - Created a helper and I've used that everywhere I can see the original, manual formatting was used
  - Added the license comment

mhamza15 (approved):

mattlord (approved):

mattlord:
@absolutejam GitHub actions seems to be stuck in some odd state. Typically the only way to move forward is to push an empty to commit. The CI is the only thing blocking the merge at this point. Thanks!

absolutejam:
Is it still the same? If so, I can push an empty commit (With sign-off, for DCO) - but will that get merged into the `main` branch or will it get squashed?

mattlord:
@absolutejam I'm sorry, but we might need to merge in origin/main in order to get this: https://github.com/planetscale/vitess-operator/pull/805

Let's see...

mattlord:
@absolutejam yeah, looks like that will be required. Sorry for the hassle.

mattlord:
@absolutejam I'm going to merge in origin/main. I have way too many things in my review queue and I think that is all that is holding this one up.

absolutejam:
Thanks for that - Apologies for the silence but I have been on leave

Issue #812, opened 2026-08-20
RFC: Volume snapshot based backups and replica provisioning


haung921209:
## Summary

Add an opt-in interface to vitess-operator that takes CSI `VolumeSnapshot`s of tablet
data volumes as a backup method, and uses those snapshots to provision new replica
tablets (PVC `dataSourceRef`). This was proposed for Vitess core in
vitessio/vitess#6952 and closed with the conclusion that the operator is the right
place for it:

> - This doesn't belong in core vitess because it is unclear how you would restore
> - Operator has the correct scope to implement a volume snapshot backup/restore
>
> — @deepthi, vitessio/vitess#6952 (2021-05-24)

As far as I can tell nobody has picked this up on the operator side since. I'd like
to implement it in phases, and want to agree on the design first. CloudNativePG's
`spec.backup.volumeSnapshot` interface is used as a reference point throughout, with
the places where MySQL/Vitess semantics differ called out explicitly.

## Motivation

- Speed: file-based backups scale with data size; snapshots are near-constant time.
  The original proposal reported ~3h for a 1.3TB shard with the builtin engine vs
  ~3min with a GCP disk snapshot.
- Cost: cloud snapshots are incremental at the block layer.
- Replica provisioning: restoring a new replica from object storage is the slow path
  for scale-up and node replacement today. Provisioning the PVC from a recent
  snapshot and letting GTID auto-positioning catch up is much faster for large shards.

I ran a proof-of-concept of the full flow against operator v2.17.0 / Vitess v24.0.2
on kind (laptop scale, so no performance claims — the numbers above stay attributed
to #6952). What the PoC verified end to end: snapshot → new PVC via `dataSourceRef`
→ mysqld comes up (crash recovery included for a SIGKILL-under-load image, with
`gtid_executed` reconstructed to the exact crash point) → GTID auto-positioning
catch-up (a 1,359s replication gap closed in 7.5s) → `CHECKSUM TABLE` matches the
primary. The PoC also surfaced two design constraints that are folded into the
sections below (mysqld supervision vs. cold fencing, and tablet-UID coupling of the
data volume layout).

## Goals (initial scope)

1. Scheduled, per-shard volume snapshot backups, orchestrated by the operator,
   with a cold (default) and an opt-in online mode.
2. Provisioning new replica tablets from the most recent usable snapshot
   (PVC `dataSourceRef`), with GTID auto-positioning for catch-up.
[...]

PR #813, opened 2026-08-20
Add volumeSnapshot backup method to VitessBackupSchedule

haung921209:
## Description

Implements the scheduled-backup slice of RFC #812: a third `BackupMethod` for `VitessBackupSchedule` that takes CSI VolumeSnapshots of tablet data volumes instead of running vtbackup/vtctldclient Jobs.

Opening as a **draft** — the API shape questions in #812 are still open, and this PR is meant to give that discussion something concrete to react to.

```yaml
spec:
  backupMethod: volumeSnapshot
  strategies:
  - name: commerce-snap
    keyspace: commerce
    shard: "-"
    volumeSnapshotClassName: my-snapclass   # optional, cluster default otherwise
```

### What it does

- At each scheduled time the controller creates a `VolumeSnapshot` of the data PVC of **one non-primary tablet** in the target shard (rdonly preferred, then replica; deterministic pick; the primary is never snapshotted). Target selection uses the observed tablet types from `VitessShard.status.tablets`.
- No Job is created. Snapshots carry the same labels and scheduled-time annotation as scheduled Jobs, and `successfulJobsHistoryLimit` / `failedJobsHistoryLimit` are applied to ready/errored snapshots (label-based cleanup, oldest first).
- If no eligible target exists at fire time (e.g. single-tablet shard), the run is skipped with a log line and retried at the next scheduled time.
- Snapshots are deliberately **not** owner-referenced to the schedule: unlike Jobs they are the backups themselves, and deleting a `VitessBackupSchedule` must not cascade-delete backups (same default CloudNativePG chose with `snapshotOwnerReference: none`). Making ownership configurable can be a follow-up (#812 open question 2).
- `VolumeSnapshot` objects are built as unstructured, so no new module dependency on external-snapshotter; the `snapshot.storage.k8s.io/v1` CRDs must be installed, and the operator ClusterRole gains `volumesnapshots` permissions (added to `deploy/role.yaml` and `operator-latest.yaml`).

### Consistency contract (documented on the API type)

The snapshot is crash-consistent, not application-consistent: restore relies on InnoDB crash recovery, which is safe only with durable settings (`sync_binlog=1`, `innodb_flush_log_at_trx_commit=1` — the Vitess defaults) and all MySQL data on a single volume. A fenced/cold mode (quiesce mysqld during the cut) is intentionally **not** in this PR — the right way to hold mysqld down needs discussion first (#812 open question 3: in-pod management restarts a cleanly shut down mysqld within seconds).

Restore needs no code at all: `dataVolumeClaimTemplate` already passes `dataSource`/`dataSourceRef` through to tablet PVCs, and vttablet treats a pre-populated data dir as a restart ("Attempting to restore, but mysqld already contains data"). Verified end to end on v2.17.0 — details in #812.

### Verification

Unit: `go test ./pkg/...` → 18 packages ok, exit 0 (3 new tests: target selection table test, snapshot creation via fake client, skip-when-no-target).

Live, on kind + csi-hostpath-driver + the example cluster (operator built from this branch):

- snapshots created each minute, targeting the non-primary replica's PVC, `readyToUse=true`, correct labels/annotations/snapshot class;
- history limit enforced ("deleted old VolumeSnapshot" after exceeding the limit, steady state at the configured count);
- no Jobs created; deleting the schedule left the snapshots in place;
- restoring one of these snapshots into a tablet PVC via `dataVolumeClaimTemplate.dataSource` produced a replica that rejoined via GTID auto-positioning and passed `CHECKSUM TABLE` against the primary.

One e2e caveat worth knowing: csi-hostpath cannot snapshot a volume with an actively-writing mysqld (`tar: file changed as we read it`), so e2e coverage for this method on kind will need a block-level test driver (e.g. topolvm) or an idle/quiesced target.

Ref: #812

Signed-off-by: Hyungmin Oh <hyungmin.oh@nhn.com>

### DIFF
--- a/deploy/crds/planetscale.com_vitessbackupschedules.yaml
+++ b/deploy/crds/planetscale.com_vitessbackupschedules.yaml
@@ -40,6 +40,7 @@ spec:
                 enum:
                 - vtbackup
                 - vtctldclient
+                - volumeSnapshot
                 type: string
               cluster:
                 type: string
@@ -131,6 +132,8 @@ spec:
                       type: string
                     shard:
                       example: '-'
+                      type: string
+                    volumeSnapshotClassName:
                       type: string
                   required:
                   - name
--- a/deploy/crds/planetscale.com_vitessclusters.yaml
+++ b/deploy/crds/planetscale.com_vitessclusters.yaml
@@ -172,6 +172,7 @@ spec:
                           enum:
                           - vtbackup
                           - vtctldclient
+                          - volumeSnapshot
                           type: string
                         concurrencyPolicy:
                           default: Forbid
@@ -257,6 +258,8 @@ spec:
                                 type: string
                               shard:
                                 example: '-'
+                                type: string
+                              volumeSnapshotClassName:
                                 type: string
                             required:
                             - name
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -101,6 +101,16 @@ rules:
   verbs:
   - '*'
 - apiGroups:
+  - snapshot.storage.k8s.io
+  resources:
+  - volumesnapshots
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
+- apiGroups:
   - autoscaling
   resources:
   - horizontalpodautoscalers
--- a/docs/api.md
+++ b/docs/api.md
@@ -2778,7 +2778,8 @@ This is populated for both cron-based and frequency-based schedules.</p>
 <p>
 <p>VitessBackupScheduleStrategy defines how we are going to take a backup.
 The VitessBackupSchedule controller uses this data to build either a vtbackup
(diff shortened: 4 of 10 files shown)


haung921209:
@mattlord could you take a look at this one when you have a moment? A few notes since it went ready for review:

- Rebased onto `main` after #818 (168d0fd9), so the `volumeSnapshot` path sits on top of the new `lastScheduledTimes` cursor.
- The workflows have been waiting for approval since that rebase — only DCO has run.
- Design context is in #812. This PR covers the schedule-side backup method only; cold/fenced snapshots are intentionally out of scope pending Q3 there.
- The two remaining Codex comments (selector without a schedule UID, tablet readiness before snapshotting) touch semantics I'd rather align on with you than guess at.

Happy to split or reshape it if a different cut is easier to review.

PR #831, opened 2026-09-16
vitessshard: compute Idle when the shard record is missing

### LINKED ISSUE #830: Old VitessShards remain idle=Unknown after Reshard complete removes source topology, blocking cleanup
## Summary

After completing a reshard from 32 to 64 shards and removing the old 32-shard partitioning from the keyspace configuration, obsolete Kubernetes `VitessShard` resources remain with `status.idle: "Unknown"`. The operator refuses to delete them even though an old source shard's topology record is already missing and the supplied serving topology references only the new layout.

Please address this cleanup failure without requiring users to patch controller-owned status or manually delete shard resources, while preserving serving-topology safety checks.

## Steps to reproduce

Reported sequence:
1. Start with a keyspace using a 32-shard partitioning.
2. Reshard to 64 shards and run `Reshard complete`.
3. Remove the old 32-shard partitioning from the applied keyspace configuration, expecting the operator to clean up its source shards.
4. Inspect remaining `VitessShard` resources and their `status.idle`.
5. Query an old source shard with `vtctldclient GetShard <keyspace>/<old-shard-range>` and inspect serving topology using `GetSrvKeyspaces`.

The exact workflow command flags and timing were not captured in this report; the sequence has not been independently reproduced in a test environment.

## Actual behavior and evidence

The operator reports:

```text
refusing to delete unwanted VitessShard ...: The shard can't be turned down because it's potentially in the serving set. You must migrate all served types in all cells to another shard before removing this shard.
```

- Remaining old shard resources show `status.idle: "Unknown"`.
- `GetShard` for an old source shard returns `node doesn't exist`.
- `vtctldclient GetSrvKeyspaces <keyspace>` returned a single cell with all three serving partitions using the new 64-shard layout:

| Cell | Served type | References |
| --- | --- | --- |
| cell_name | 1 — PRIMARY | `-04`, `04-08`, `08-0c`, …, `fc-` |
| cell_name | 2 — REPLICA | Same 64-shard layout |
| cell_name | 3 — RDONLY | Same 64-shard layout |

The old equal-32 ranges (`-08`, `08-10`, …, `f8-`) are absent from the supplied partitions. This evidence covers the returned a single name cell; it is not an independent verification of every possible cell.

## Expected behavior

Once the old partitioning is no longer desired and the operator can safely establish that no cell/tablet type serves the old shards, cleanup should converge automatically—even if reshard completion has already removed the global shard topology records.

A missing topology record should not leave obsolete resources permanently blocked solely because the existing idleness check requires that record. Conversely, missing records or topology read failures must not be treated as unconditional permission to delete potentially serving shards.

## Suspected cause / relevant code

Source inspection at commit `769469d985b125a27efdba484e6b5c2729ac95d7` suggests the following path. This commit is the inspected repository revision, not a confirmed deployed operator version.

1. [`ReconcileVitessShard.Reconcile` resets status before recomputing it](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitessshard/vitessshard_controller.go#L174-L184), so an earlier idle observation is not retained.
2. [`reconcileTopology` only assigns `Status.Idle` after both `GetShard` and `GetShardServingCells` succeed](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitessshard/reconcile_topo.go#L73-L96). A missing shard record takes the `TopoGetFailed` / `failed to get shard info: ...` branch and requeues instead. That event text is identified from code, not supplied runtime logs.
3. [The keyspace controller's `PrepareForTurndown` requires `Status.Idle == ConditionTrue`](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitesskeyspace/reconcile_shards.go#L143-L158). An unknown value blocks deletion with the serving-set warning.

This is consistent with source topology records disappearing before Kubernetes cleanup completes. The precise deletion timing and actor have not been independently confirmed from logs.

## Requested fix and regression coverage

- [ ] Handle confirmed missing shard records during obsolete-shard cleanup using a safe way to validate serving references across all relevant cells and tablet types.
- [ ] Distinguish confirmed absence from connection errors, timeouts, permission failures, and incomplete topology observations; continue blocking unsafe or unverifiable cleanup.
- [ ] Improve diagnostics to distinguish a missing topology record from an actual remaining serving reference.
- [ ] Add regression coverage for reshard completion removing source topology before the old partitioning is removed from the configuration.
- [ ] Cover repeated reconciliations/status resets so cleanup does not depend on catching a transient `idle: "True"` value.
- [ ] Verify that a remaining serving reference in any cell or tablet type still prevents deletion, including when the global shard record is missing.

## Workaround considered

Manually patching the status subresource to the string `"True"` was discussed as a way to pass the deletion guard. It is race-dependent because the shard controller recomputes status, can authorize destructive cleanup, and is not an appropriate normal resharding procedure. Successful cleanup using this workaround has not been confirmed in this report.

## Environment

- Resource API: `planetscale.com/v2`, `VitessShard`.
- Resharding: 32 source shards to 64 target shards.
- Serving topology supplied for the single cell
- Operator image/version, Vitess version, and Kubernetes version: not provided.

## Related report

https://github.com/planetscale/vitess-operator/issues/169 describes old shards remaining after `DropSources` and removal of the old partitioning. It is closed; this report does not establish whether the underlying cause is identical.
https://github.com/planetscale/vitess-operator/issues/741

mcrauwel:
Fixes #830.

## Problem

After a Reshard, `Reshard complete` deletes the source shard records from the global topo ([`dropSources` → `dropSourceShards`](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/vt/vtctl/workflow/server.go#L2547-L2553) → [`DeleteShard`](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/vt/vtctl/workflow/traffic_switcher.go#L608-L620)). When the old partitioning is then removed from the `VitessKeyspace`, the shard controller can no longer compute `status.idle` for the source shards: `Idle` was only assigned inside the `GetShard` success branch, a missing record took the `TopoGetFailed` requeue path, and the [status reset at the top of every reconcile](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitessshard/vitessshard_controller.go#L174-L184) meant the earlier `True` was never retained. The keyspace controller's [`PrepareForTurndown`](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitesskeyspace/reconcile_shards.go#L141-L152) then refuses forever:

```
refusing to delete unwanted VitessShard ...: The shard can't be turned down because it's potentially in the serving set.
```

There is no order of operations that avoids it:

- `Reshard complete` has no `--keep-data` flag ([only `MoveTables complete` and `Reshard cancel` do](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/cmd/vtctldclient/command/vreplication/reshard/reshard.go#L62-L66)), so source shard deletion can't be suppressed.
- Skipping `complete` isn't possible: [`Workflow delete` refuses once writes are switched](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/vt/vtctl/workflow/server.go#L1627-L1634), and the frozen workflow blocks the next Reshard.
- Removing the partitioning first and running `complete` afterwards fails too: [`buildTrafficSwitcher` needs every source shard record](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/vt/vtctl/workflow/server.go#L2403-L2416).

## Change

When `GetShard` returns `NoNode` (and only `NoNode`), compute `Idle` from the serving partitions. `GetShardServingCells` only needs the keyspace and shard name, so a `ShardInfo` is built from those. Zero serving cells and no error → `Idle=True`; any error, including a `PartialResult` from a cell that couldn't be read, leaves `Idle=Unknown` so the turndown stays blocked. `HasMaster`/`ServingWrites` are `False` without a record. Other `GetShard` errors behave as before.

The tablet-record lookup is skipped when the record is missing, since [it can only fail without one](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/controller/vitessshard/reconcile_topo.go#L98-L117) and would otherwise requeue every 5s until the shard is turned down.

The operator's own [`PruneShards`](https://github.com/planetscale/vitess-operator/blob/769469d985b125a27efdba484e6b5c2729ac95d7/pkg/operator/vitesstopo/prune_shards.go#L79-L84) already tolerates `NoNode`, so nothing else needs to change for cleanup to converge.

`reconcileTopology` is split into the connection part and `reconcileTopologyWithServer`, so the latter can be tested against `memorytopo`.

### Safety

The fallback deliberately keeps the same cell-observation tolerance as the existing success path: a cell without a `SrvKeyspace` is [already treated as "not serving this shard"](https://github.com/vitessio/vitess/blob/97cd5bd43dadba20b82b204b577e8f9fbdfae3b7/go/vt/topo/srv_keyspace.go#L161-L167) today, and any other read failure yields a `PartialResult` error, which keeps `Idle=Unknown`. A remaining reference in any cell or tablet type still blocks turndown.

## Tests

`TestReconcileTopologyIdleWithoutShardRecord` (memorytopo, two cells):

- record present, no serving references → `True`
- record deleted, no serving references → `True` on three consecutive reconciles with the controller's status reset, no `TopoGetFailed` events
- record deleted, one cell's RDONLY partition still references the shard → `False`
- record deleted, topo reads failing → `Unknown` with a `TopoGetFailed` event

### Reproduced and verified with the repo's example manifests

Against an operator built from `main` (769469d), on a local Kubernetes with the default storage class, using `test/endtoend/operator/`:

```sh
kubectl create namespace example
kubectl apply -f operator-latest.yaml                       # image built from main
sed -n '104,$p' 101_initial_cluster.yaml | kubectl apply -f -   # Secret + vitess-backups PVC
kubectl apply -f 201_customer_tablets.yaml                  # commerce + customer, one shard each
# create two tables in `customer` with a hash vindex (no MoveTables needed for this)
kubectl apply -f 302_new_shards.yaml                        # adds -8000 / 8000-

vtctldclient Reshard --workflow cust2cust --target-keyspace customer create --source-shards '-' --target-shards '-8000,8000-'
vtctldclient Reshard --workflow cust2cust --target-keyspace customer switchtraffic --tablet-types replica,rdonly
vtctldclient Reshard --workflow cust2cust --target-keyspace customer switchtraffic --tablet-types primary
# wait until the VitessShard for '-' reports idle=True (shard controller resync is 30s)
vtctldclient Reshard --workflow cust2cust --target-keyspace customer complete
# wait until the VitessShard for '-' reports idle=Unknown
kubectl apply -f 306_down_shard_0.yaml                      # removes the one-shard partitioning
```

Result on `main`: the `VitessShard` for `-` stays `idle=Unknown`, the `VitessKeyspace` reports `orphanedShards["-"] = {reason: Serving}`, and the old shard's three vttablet pods plus vtorc keep running (observed for over an hour).

```
$ kubectl -n example get vitessshard -o custom-columns='NAME:.metadata.name,SHARD:.spec.name,IDLE:.status.idle'
NAME                               SHARD   IDLE
example-customer-8000-x-1f632b0c   8000-   False
example-customer-x-8000-81888d38   -8000   False
example-customer-x-x-dc880356      -       Unknown
```

Rolling the operator Deployment to an image built from this branch, the orphaned shard is deleted on the first reconcile after start:

```
Warning  TurndownBlocked  refusing to delete unwanted VitessShard example-customer-x-x-dc880356: ...
Normal   Deleted          deleted VitessShard example-customer-x-x-dc880356
```

Afterwards `orphanedShards` is empty, the old pods are gone, and `GetTablets --keyspace customer` lists only the `-8000` and `8000-` tablets.

### Note on the existing upgrade e2e test

`test/endtoend/upgrade_test.sh` runs this exact sequence (`Reshard Complete`, then `306_down_shard_0.yaml`) and passes on `main`. That works only because it applies the YAML immediately after `Complete`: the `VitessShard` keeps a stale `idle=True` until the shard controller's next reconcile (observed 13s, bounded by the 30s resync period), and the keyspace controller consumes that stale value. Anyone who waits longer hits the bug every time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mattlord (commented):
1. **Blocking:** I think that we should leave `HasMaster` (and `ServingWrites`) as `Unknown` when the record is missing and only compute `Idle`. `False` isn't a status-only change: `vitessshardreplication` reads it, and `readyForPrimary` (`init_shard_primary.go:135`) treats `Unknown` as "wait" but `False` as "go initialize". With the record missing, this PR also skips the tablet lookup, so `Status.Tablets[*].Type` keeps the pool type that `reconcile_tablets.go:329` seeded (`"replica"`), the "registered in topology" check passes, and `wr.InitShardPrimary(force=true)` — or `electInitialShardPrimary` for shards with backups — gets called on any shard that is still *desired* without a record: the old source shards between `Reshard complete` and the partitioning removal, and every new shard between pod start and its first tablet creating the record. Both take `LockShard` first and etcd's `Lock` returns `NoNode` for a missing directory, so nothing can actually be reparented, but each attempt becomes an `InitShardFailed` warning and a 5s requeue for as long as the state lasts. I confirmed the chain by feeding `readyForPrimary` exactly the status this branch writes (`HasMaster: False`, one Running tablet with `Type: "replica"`) — it returns true. `Idle` is all that `PrepareForTurndown` needs, so keeping the other two `Unknown` preserves today's behavior there; the test's `HasMaster`/`ServingWrites == False` assertion would flip accordingly, and a small `readyForPrimary` test with `HasMaster: Unknown` and a pool-typed Running tablet would pin that the replication controller keeps waiting. No?

2. **Non-blocking:** I wonder if the upgrade e2e should wait past the shard controller's resync before applying `306_down_shard_0.yaml`. As the description notes, it passes on `main` only because it applies the YAML inside the window where `idle=True` is stale; sleeping longer than the 30s resync (or waiting for the `VitessShard` to be re-reconciled after `Complete`) would make it exercise this fix for real and catch a regression of it.

The `Idle` computation itself is right: `GetShardServingCells` only needs `si.ShardName()` and `si.keyspace`, treats a missing per-cell `SrvKeyspace` as "not serving" exactly as the success path already does, and returns `PartialResult` for anything else, so a partial view can't produce `True`; and with `dropSourceShards` deleting recursively there are no tablet records left for the skipped `pruneTablets` to miss. Thanks for the thorough reproduction in the description — it made the timing dependency in the existing e2e test obvious. ❤️

mcrauwel:
Good catch on `HasMaster`, thanks. I'd only looked at what the keyspace controller reads. Confirmed your chain: `readyForPrimary` returns early for `True` and `Unknown` only, so `False` plus a Running tablet still typed `replica` walks straight into `InitShardPrimary`, and `initRestoredShard` has the same switch.

Pushed a follow-up:

1. The NoNode branch now computes `Idle` only; `HasMaster` and `ServingWrites` stay `Unknown`, with a comment explaining why. The `reconcile_topo` test asserts `Unknown/Unknown` accordingly.
2. New `TestReadyForPrimaryHasMasterUnknown` in `vitessshardreplication`: `Unknown` with a Running `replica`-typed tablet → `false` plus a `replicationRequeueDelay` requeue; `False` with the same tablet → `true`. That pins the contract the shard controller now relies on.
3. Upgrade e2e: after `Reshard Complete`, `waitForVitessShardStatusField example "planetscale.com/keyspace=customer,planetscale.com/shard=x-x" hasMaster Unknown` before applying `306_down_shard_0.yaml`. The flip from `True` to `Unknown` is the observable sign that the shard controller has re-reconciled against the deleted record, so the teardown now goes through the no-record `Idle` path instead of the stale value. On `main` this step would time out at the 9-pod check, which is what we want.

Also merged `main` in to pick up the #823 rollout-wait changes in the same script.

mattlord (commented):
1. **Non-blocking:** I think that we should make the "record missing and topo unreadable" subtest (`pkg/controller/vitessshard/reconcile_topo_test.go:156`) reach the new branch. `factory.SetError` is global in memorytopo, so `GetShard` itself fails with a non-`NoNode` error and the subtest goes through the old generic `else`, which behaves the same on `main`. The new fail-closed path at `reconcile_topo.go:124` is the line that guards "an incomplete view must never be treated as permission", and nothing pins it today: I changed it locally to set `Idle=True` on error and all four subtests still pass. Corrupting one cell's `SrvKeyspace` does get there, e.g. `conn, _ := ts.ConnForCell(ctx, cellB)` followed by `conn.Update(ctx, "keyspaces/customer/SrvKeyspace", []byte("garbage"), nil)`. With that, `GetShardServingCells` returns `PartialResult`, `Idle` stays `Unknown` with the `TopoGetFailed` event, and the subtest fails under the mutation above.

3. **Non-blocking:** I wonder if the fallback should also require the keyspace record to exist before trusting `NoNode`. On the success path, `GetShard` succeeding is implicit proof that we're looking at a populated topo for this keyspace, and the fallback at `reconcile_topo.go:105` has no equivalent. With memorytopo I get `Idle=True` both when the keyspace is absent from the topo entirely (cells present, no `SrvKeyspace`) and when the global topo is empty (`GetCellInfoNames` returns no cells on `NoNode`, so nothing can contradict it). Before this PR both were `Unknown`. It only matters for a shard that's already orphaned because it is still serving, combined with the operator seeing an empty or wrong topo (lost or recreated etcd, a changed `rootPath`), but in that combination the turndown guard now fails open and the serving shard gets deleted. After `Reshard complete` the keyspace record is still there, and the keyspace controller creates it for new keyspaces (`pkg/controller/vitesskeyspace/reconcile_keyspace_information.go:54`), so a `ts.GetKeyspace` check that leaves `Idle=Unknown` on `NoNode` wouldn't get in the way of the case this PR fixes, and it costs one extra global read only while the shard record is missing. No?

Thanks for the quick turnaround! ❤️

mcrauwel:
Both taken, thanks.

**1. Fail-closed branch not covered.** You're right, `factory.SetError` made `GetShard` itself fail, so that subtest never left the generic path. Replaced it with two that do reach the fallback:
- `record missing and one cell unreadable`: corrupts zone2's `SrvKeyspace` via `ConnForCell` + `Update`, so `GetShard` is `NoNode` and `GetShardServingCells` returns `PartialResult`. Asserts `Idle=Unknown` + `TopoGetFailed`.
- `record missing and keyspace record missing`: deletes the keyspace record. Asserts `Idle=Unknown` + `TopoGetFailed`.

The global-error case stays as `record missing and topo unreachable`. I ran your mutation (`Idle=True` on the `PartialResult` path) and the first new subtest fails with `Idle = "True", want Unknown`; disabling the keyspace check fails the second the same way.

**3. Keyspace record as proof of a populated topo.** Agreed, added. The fallback now does `GetKeyspace` first and leaves `Idle=Unknown` on any error, with a comment explaining the empty/foreign-topo failure mode. `Reshard complete` keeps the keyspace record, so the cleanup case still converges.

**Upgrade Test flake.** Both red runs on this PR died on the `MoveTables SwitchTraffic` dial timeout, which I see is exactly what #841 addresses (with the `observedGeneration` wait on top, which my two-line version would have missed). I've left the e2e alone here beyond the `hasMaster` wait and will merge `main` once #841 lands, then re-run.

mattlord (approved):
LGTM! Both follow-ups landed the way I'd hoped. The keyspace-record check closes the empty or foreign topo case without getting in the way of the post-`Reshard complete` cleanup, and the two new subtests now pin both fail-closed branches. It's also nice to see the Upgrade Test go through the no-record `Idle` path for real now that #841 is in. Thanks for sticking with this one, @mcrauwel ! ❤️

pkg/controller/vitessshard/reconcile_topo.go
```diff
@@ -90,12 +102,56 @@ func (r *ReconcileVitessShard) reconcileTopology(ctx context.Context, vts *plane
...
+		// shard into a deletion candidate. `Reshard complete` leaves the
+		// keyspace record in place, so this doesn't get in the way of the
+		// cleanup case.
+		if _, err := ts.GetKeyspace(ctx, keyspaceName); err != nil {
+			// Leave Idle as Unknown.
+			r.recorder.Eventf(vts, corev1.EventTypeWarning, "TopoGetFailed", "shard record does not exist and failed to get keyspace record: %v", err)
+			resultBuilder.RequeueAfter(topoRequeueDelay)
+		} else {
+			shard := topo.NewShardInfo(keyspaceName, vts.Spec.Name, &topodatapb.Shard{}, nil)
+			if servingCells, err := ts.GetShardServingCells(ctx, shard); err == nil {
```
maxenglander: Non-blocking follow-up: `GetShardServingCells` suppresses every `NoNode` returned by `GetSrvKeyspace`, including a `NoNode` from `ConnForCell` when an enumerated cell's `CellInfo` is missing. In that case this can return an empty successful result and set `Idle=True` without observing that cell. Please consider distinguishing a missing `SrvKeyspace` from a missing `CellInfo`, or otherwise verifying every enumerated cell before authorizing turndown, so this path continues to fail closed.

maxenglander (approved):
The normal post-Reshard path is handled well, and the unit and end-to-end coverage is strong. I left one non-blocking safety follow-up inline.

